PR #14, opened 2020-06-01
Improve phonebook (devirtualize and use smart pointers) and improve pose prediction

charmoniumQ:
- I initially developed this with separating the implementation from the interface. This means each component will not have to rebuild phonebook every time. Since per-component build-times are not a problem, I have less motivation to do this. On the other hand, the negative impact of this separation makes it more difficult to modify, and it incurs the virtual function-call overhead. As such, I devirtualized phonebook on the way to making other changes. This also means that `main` can hold the _value_ of a phonebook, instead of a _pointer_ to a phonebook, which improves the memory-safety.

- I used smart pointers for each service (for inputs, outputs, and while in storage).
  - A consequence of this change, is that **all plugins** changed `switchboard*` to `const std::shared_ptr<switchboard>`.
  - While I was doing that, I also sprinkled in some `const` to the usage of `phonebook`.

- Phonebook itself cannot be passed as a smart-pointer, because dynamically loaded plugins need phonebook as an argument, but to be dynamically loaded, they must export C-linkage, which means they can't take smart-pointers. The good news is that phonebook is the _only_ argument to dynamically-loaded plugins; other services (switchboard, pose predict, logger, more to come) are accessed indirectly through phonebook as `std::shared_ptr`.

- While I was executing this change, I needed to change `pose_prediction` to use 2 classes. If the same class was both a `plugin` in spindle and a `service` in phonebook, who is responsible for deleting it? It would get deleted twice, now that phonebook properly reclaims memory at exit-time. As such, `pose_prediction` is a `plugin` in spindle, which instantiates `pose_prediction_impl` which is a `service` in phonebook.
  - While I was doing that, I also changed the return type of `pose_prediction` to not be a pointer. When passing-by-value is possible, and there is no reuse from using a pointer, it is better to pass-by-value and skip the heap-allocation.
  - This also reflects my conversation with Huzaifa that pose_predict always returns _something_, even if an accurate pose is not yet available from the sensors.
  - Jeffrey is aware of this change, and it will not negatively impact his work on OpenVINS pose prediction.

This loosely tracks [Trello #89](https://trello.com/c/nMISs4YZ/89-use-smart-pointers-in-phonebook).


debugview/plugin.cpp
```diff
@@ -145,34 +144,39 @@ class debugview : public plugin {
 			ImGui::Text("Resets to zero'd out tracking universe");
 		}
 		ImGui::Spacing();
-		const pose_type* fast_pose_ptr = pp->get_fast_pose();
+		const pose_type fast_pose_ptr = pp->get_fast_pose();
```
mhuzai, line 147: Minor: fast_pose_ptr is no longer a `ptr` but the actual fast pose itself. Similar issue on line 149.
```suggestion
		const pose_type fast_pose = pp->get_fast_pose();
```

debugview/plugin.cpp
```diff
@@ -187,9 +191,8 @@ class debugview : public plugin {
 		ImGui::Text("	Camera0: (%d, %d) \n		GL texture handle: %d", camera_texture_sizes[0].x(), camera_texture_sizes[0].y(), camera_textures[0]);
 		ImGui::Text("	Camera1: (%d, %d) \n		GL texture handle: %d", camera_texture_sizes[1].x(), camera_texture_sizes[1].y(), camera_textures[1]);
 		if(ImGui::Button("Calculate new orientation offset")){
-			const pose_type* pose_ptr = pp->get_fast_pose();
-			if(pose_ptr != NULL)
-				offsetQuat = Eigen::Quaternionf(pose_ptr->orientation);
+			const pose_type pose_ptr = pp->get_fast_pose();
```
mhuzai, line 194:
```suggestion
			const pose_type pose = pp->get_fast_pose();
```

debugview/plugin.cpp
```diff
@@ -307,30 +310,30 @@ class debugview : public plugin {
 
 			glUseProgram(demoShaderProgram);
 
-			const pose_type* pose_ptr = pp->get_fast_pose();
+			const pose_type pose_ptr = pp->get_fast_pose();
```
mhuzai, line 313:
```suggestion
			const pose_type pose = pp->get_fast_pose();
```

debugview/plugin.cpp
```diff
@@ -371,9 +374,9 @@ class debugview : public plugin {
 			headsetObject.color = {0.2,0.2,0.2,1};
 			headsetObject.drawMe();
 
-			const pose_type* groundtruth_pose_ptr = pp->get_fast_true_pose();
-			if(groundtruth_pose_ptr){
-				headsetPose = generateHeadsetTransform(groundtruth_pose_ptr->position, groundtruth_pose_ptr->orientation, tracking_position_offset);
+			const pose_type groundtruth_pose_ptr = pp->get_fast_true_pose();
```
mhuzai, line 377:
```suggestion
			const pose_type groundtruth_pose = pp->get_fast_true_pose();
```

charmoniumQ:
> Why is there
> `if (true /* pose is valid */) {`
> in several places? Are we expecting the `true` to change to something else in the future?

If in the future, we decide to not consider a pose valid until SLAM has run successfully, those conditionals are what should change I left them in so that we would not have to re-identify them in that case.

Currently, we return a "default" pose (position and rotation are zero vectors). Without changing the interface, we could also return a pose based solely on the IMU. This reflects conversations I have had with you and Finn.

mhuzai:
That's what I thought, but just wanted to make sure.

apache8080:
Can we add a comment over the `if (true)` statements that explain what you posted in this PR @charmoniumQ?

charmoniumQ:
I didn't do this initially because pose was used on line 334, but I think the solution is to get it again, and guard it (branch if reliable).

PR #16, opened 2020-06-12
Add "zero orientation" fetaure

charmoniumQ:


pose_prediction/plugin.cpp
```diff
@@ -32,11 +36,18 @@ class pose_prediction_impl : public pose_prediction {
         return correct_pose(true_pose);
     }
 
+	virtual void set_zero(const Eigen::Quaternionf& orientation) override {
+		std::lock_guard<std::mutex> lock {zero_mutex};
+		zero = (orientation * zero.inverse()).inverse();
```
mhuzai, line 41: This intuitively makes sense to me but could you explain this line a little bit more just to make things airtight? I guess a comment here and on line 71 would help clarify things for future readers.

charmoniumQ: Ok. I clarified the variable names (`orientation -> raw_o_times_offset`) and added an explanatory comment).

PR #18, opened 2020-06-14
Feat get cpu time

charmoniumQ:
Review the changes in this order:
- I added `common/cpu_timer.hpp`. It does what the name implies, using the automatic destructor to trigger timer end (RAII).
- I instrumented `common/threadloop.hpp, class threadloop` (plugin classes can choose to subclass this, so adding features to it adds features to every plugin that uses it) to spit out the elapsed time every cycle.
- But in order to do so, I need to know the name of the plugin, so I changed the constructor to have a `name` field (first in `common/plugin.hpp` but also in `common/threadloop.hpp` which inherits). This is quite painful, because I had to touch the constructor of _every plugin_ (see all the `plugin.cpp` changes are just to the constructor line).
- However, not all plugins use `threadloop`. I brought `timewarp_gl`, `gldemo`, and `debugview` up to using `threadloop`. They already had a thread with a loop; I just brought the body of the loop into its own function `_p_one_iteration()` and the prologue of the loop to `start()` (which must now call super's start (AKA `threadloop::start()`).
- However, this is this odd behavior with `glXMakeCurrent` has to be called from the child thread, not the parent, so I have this `first_time` variable.


charmoniumQ:
This still doesn't capture one thing: time spent in synchronous dependency handlers (AKA callbacks). I will work on that in a different PR.

common/threadloop.hpp
```diff
@@ -17,13 +19,19 @@ namespace ILLIXR {
...
 	void start() override {
 		_m_thread = std::thread([this]() {
-		while (!should_terminate()) {
-				_p_one_iteration();
+			while (!should_terminate()) {
+				std::chrono::nanoseconds duration;
+				{   TIME_THIS_BLOCK(duration);
+					_p_one_iteration();
+				}
+				std::cout << "cpu_timer," << get_name() << "," << duration.count() << std::endl;
```
mhuzai, line 34: I'm assuming the use of `cout` is intentional? That is, we dump to `cout` and then use `grep` to get what we need?

runtime/switchboard_impl.hpp
```diff
@@ -238,13 +239,14 @@ namespace ILLIXR {
 			while (!_m_terminate.load()) {
 				std::pair<std::string, const void*> t;
 				if (_m_queue.try_dequeue(t)) {
+					std::chrono::nanoseconds duration;
```
mhuzai, line 242: `duration` appears to be unused.

common/threadloop.hpp
apache8080: That would make sense to me. #UnixPhilosophy

debugview/plugin.cpp
```diff
@@ -274,12 +274,13 @@ class debugview : public plugin {
...
 
-	void main_loop() {
-		double lastTime = glfwGetTime();
-		glfwMakeContextCurrent(gui_window);
-		
-		while (!_m_terminate.load()) {
-
+	bool first_time = true;
+	void _p_one_iteration() override {
+		if (first_time) {
```
apache8080, line 279: Can you add a comment explaining this first_time logic?

common/threadloop.hpp
charmoniumQ: Exactly, `make run.dbg grep '^cpu_timer,' > all_timers.csv`.

charmoniumQ: If I logged to a file, that file object would need to be globally accessible at every site I need to write timing info. Currently, this is the only site, but that will change.

runtime/switchboard_impl.hpp
charmoniumQ: Currently, yes. This will change.

charmoniumQ:
> `decoder->loadSource();`
> `encoder->loadSource();`

I didn't notice those references when I changed encoder/decoder from pointers to values. This is fixed now.

runtime/switchboard_impl.hpp
charmoniumQ: This is removed now.

charmoniumQ:
Reviewers: I had to make a major new change
- This would time each component's threadloop, but synchronous callbacks get execute in switchboard's threadpool, so I added some stuff to make that work (`common/switchboard_impl.hpp`).

debugview/plugin.cpp
charmoniumQ: Done. I also `first_time -> first_iteration` to be consistent with the name of the method.

PR #21, opened 2020-06-30
Assorted fixes 2

charmoniumQ:
Changes from rebuttal branch:

- `Makefile`:
  - Added `runtime/plugin.{opt,dbg}.so` target, which allows ILLIXR to be built for Monado without changing.
- `common/cpu_timer.hpp`:
  - Now works with a greater variety of tyeps, due to `declval`/`decltype` in template args.
  - Added `PRINT_WALL_TIME_FOR_THIS_BLOCK` (previously we just had CPU time).
- `common/data_format.hpp`:
  - Added `rgb_depth`. This will be used by Zed and ElasticFusion.
  - Remvoed `global_config`. We use Switchboard for this now.
  - Added timestamp to `rendered_frame`, so we can compute end-to-end latency.
- `runtime/dynamic_lib.hpp -> common/dynamic_lib.hpp`
  - Moved to `common/` so that we can import in Monado in the future.
- `common/plugin.hpp`:
  - Renamed `plugin_main -> plugin_factory`. This name reflects the terminology of a [factory pattern](https://en.wikipedia.org/wiki/Factory_(object-oriented_programming)).
- `common/threadloop.hpp`:
  - Fixed `reliable_sleep` so that it can be used with higher precision.
  - I pulled the lambda out into its own function and call `std::bind` on it. This is more clear to readers who do not know about lambda-syntax and more efficient (I think constructing a bind is cheaper than constructing a lambda).
  - I added `_p_thread_setup()`, which runs in the new thread at initialization-time.
- `ground_truth_slam/data_loading.hpp`:
  - Check that file is good before continuing. Previously, this would silently fail, which causes problems when not run from the `ILLIXR/` directory (as in Monado).
- `ground_truth_slam/plugin.cpp`:
  - The default destructor is fine. No need to override.
- `offline_imu_cam/data_loading.hpp`:
  - Check that file is good before continuing. Previously, this would silently fail.
- `runtime/main.cpp`:
  - Changes necessitated by the above changes.
- `runtime/switchbaord_impl.hpp`:
  - Fix bug: `enqueue` was never being called in optimized build, because assertions were compiled away.
  - Time the synchronous dependencies.
- `timewarp_gl/plugin.cpp`:
  - Utilize `threadloop` instead of making a thread itself. This means CPU timers in threadloop measure timewarp now.
  - Compute M2P based on the 
- Every file I touch, I also change the header guard to `#pragma once` while I'm there.

Changes conspicuously absent:

- Zed code. Should be reviewed as a separate PR.
- ElasticFusion. Does not completely work yet.
- Optimized `_p_should_skip()` logic.` This change touched many components.

Required for:
- ElasticFusion (due to `_p_thread_setup()`).


common/data_format.hpp
```diff
@@ -46,18 +51,12 @@ namespace ILLIXR {
...
 
-	class global_config : public phonebook::service {
-	public:
-		global_config(GLFWwindow* _glfw_context) : glfw_context(_glfw_context) { }
-		GLFWwindow* glfw_context;
-	};
-
 	// Single-texture format; arrayed by left/right eye
 	// Single-texture format; arrayed by left/right eye
 	struct rendered_frame {
```
mhuzai, line 56: Is this format used given `rendered_frame_alt` below?

offline_imu_cam/data_loading.hpp
```diff
@@ -47,6 +47,7 @@ load_data(const std::string& data_path) {
 	std::map<ullong, sensor_types> data;
 
 	std::ifstream imu0_file {data_path + "imu0/data.csv"};
+	assert(imu0_file.good());
```
mhuzai, line 50: Please add `cam0_file.good()` and `cam1_file.good()` below.

charmoniumQ:
> I understand why ZED and EF are missing. Why is the `_p_should_skip()` logic not in this PR?

It would be a large amount of non-localized changes, so I wanted to do it in its own PR.

PR #22, opened 2020-07-03
Add Zed (IMU/camera/depth) plugin

Hyjale:
ZED plugin integration.


charmoniumQ:
Let's make the ZED camera -> ZED IMU data-race free. The way we do this depends on if we separate `imu_cam_type` into `imu_type` and `cam_type`. If we do that change, then this becomes much simpler, because the `zed_camera_thread` can publish independently from the `zed_imu_thread`. If they are fused, then the `zed_imu_thread` has to the publishing, which means it has to access the results in `zed_camera_thread` asynchronously, which could occur while a new result is being computed. This write/read conflict means `zed_camera_thread` has to be refactored to be double-buffered.

1. Turn `cv::Mat grayL` into `std::array<cv::Mat, 2> grayL`, likewise for `grayR`.
2. Remove the `std::atomic`s you have there and delete their accesses.
3. Add a `std::atomic<unsigned> which {0};` and `std::array<unsigned, 2> count` as an instance member of `zed_camera_thread`.
4. In `zed_camera_thread::_p_one_iteration()`, do `unsigned _which = !which.load();`, and at the end, `count[_which] = 1 + count[!_which]; which.store(_which);`, and in the middle, write to `grayL[_which] = ...` (and `grayR[_which]`).
6. Then, the reader (`zed_imu_thread::_p_one_iteration()`) does `unsigned _which = camera_thread_.which.load(); if (count != camera_thread_.count[_which]) { ... }`.
7. We'll figure out the rest together.

zed/zed-opencv/LICENSE
```diff
@@ -0,0 +1,22 @@
+MIT License
```
mhuzai, line 1: Why are we using Stereolabs' license and not our own NCSA one? Is this because of the reference code that you used?

Hyjale: I can change the directory structure. The Stereolab's license is from the reference code (their OpenCV sample).

apache8080: Did we use that code in our code? If we did then we have to keep the Stereolab License.

Makefile
```diff
@@ -1,4 +1,4 @@
-plugins    = ground_truth_slam/ offline_imu_cam/ open_vins/ pose_prediction/ gldemo/ timewarp_gl/ debugview/ audio_pipeline/
+plugins    = ground_truth_slam/ zed/ open_vins/ pose_prediction/ gldemo/ timewarp_gl/ debugview/ audio_pipeline/
```
apache8080, line 1: @charmoniumQ we may want to be able to take these plugins as arguments so people can pick and choose what they want a little easier. Editing the make file is a little cumbersome.

common/tests/test_cpu_timer.cpp
```diff
@@ -1,28 +0,0 @@
-#include <gtest/gtest.h>
```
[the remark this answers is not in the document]
Hyjale: My mistake, I got rid of it because it was interfering with ZED building.

zed/zed-opencv/LICENSE
Hyjale: I used their function of shared pointers between ZED images and their OpenCV counterparts.

common/tests/test_cpu_timer.cpp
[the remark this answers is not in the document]
Hyjale: Fixed.

Makefile
charmoniumQ: Marking as resolved, because there is a config-file PR.

charmoniumQ:
Merging in in smaller pieces.

PR #24, opened 2020-07-17
Add should_skip logic

charmoniumQ:
Approve after `assorted-fixes-2`


charmoniumQ:
Yes, both intentional. I have changed those components to utilize `_p_should_skip`.

The Hologram has a few additional unrelated (but beneficial) changes from the rebuttal branch.

mhuzai:
> 
> 
> Nit: are these version bumps of audio_pipeline and hologram on purpose?

Review when you get a chance.

PR #26, opened 2020-07-22
Added runner/config file

charmoniumQ:
This is just a documentation of interface PR. Functionality of the interface will come later. In the documentation, I have also written the rationale as for why this interface is necessary and why it is a useful and future-proof abstraction.

These issues will be addressed through a config file:

- Make script for the export stuff (it's difficult to remember to export and type manually).
- Fix the issue where we need a colon at the end.
- Fix the hard-coded path in offline_imu_cam and groundtruth_slam.
- Change the camera intrinsics without changing the source code.


charmoniumQ:
CMake, Nix, Buck, and Bazel also provide other ways of managing build of multiple projects. In retrospect, using a pre-made tool could be better than my custom Python script (faster, more well-known, less buggy, more features. etc.).

I think this PR should still be accepted because I believe the interface is "right". Even if we use CMake or something, we still need a list of plugins to build and other configuration options. I think those are best served by YAML and this particular schema. In the future, the same `runner.py` could parse the config file, set variables for CMake, and kick off **one** CMake build.

This PR is the smallest jump from the build scripts we had to a build script that uses the YAML config-file interface.

install_deps.sh
```diff
@@ -6,53 +6,54 @@ set -x -e
...
+        cd ../..
+    fi
 
     cd "${old_pwd}"
-
-	curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/get-poetry.py | python
-	cd runner
-	poetry install
-	cd ..
-
```
mhuzai: Why were these removed? Aren't they required for the runner?

charmoniumQ: I restored this installation step.

install_deps.sh
```diff
@@ -6,7 +6,7 @@ set -x -e
 
 if [ "${ID_LIKE}" = debian -o "${ID}" = debian ]
 then
-    sudo add-apt-repository ppa:graphics-drivers/ppa
+    sudo add-apt-repository-repository ppa:graphics-drivers/ppa
```
mhuzai, line 9: "repository" is duplicated?

charmoniumQ: Resolved the `apt-install-repository` problem (strange thing).

Maybe @Hyjale has insight on why this is necessary?

docs/building_ILLIXR.md
```diff
@@ -2,10 +2,34 @@
...
-- From the project root, `make run.dbg -j$(nproc)` will build the ILLIXR runtime standalone, the
-  ILLIXR plugins, and run it for you. `make run.opt -j$(nproc)`, is the same but with optimizations on and debug off.
+- Edit `config.yaml`. See `runner/config_schema.yaml` for the schema definition.
 
-- To build the runtime for Monado, `make all.dbg.so`.
+  * This config file defines the list of plugins to use and the XR application to run.
+
+  * Anywhere a path is required, a Git URL can be used in this format: `git+https://github.com/username/repo@rev?path=optional/path/within/repo` where rev is a tag, commit hash, or branch.
+
+  * Likweise, a zip "URL" can be used in this format: `zip+http://path/to/archive.zip?path=optional/path/within/zip`. Replace `http` with `file` for local archives.
```
mhuzai, line 11: Make it clear that these are future features.

install_deps.sh
Hyjale: This was added right before the ADA release and was necessary before installing `glslang-tools`.

PR #28, opened 2020-07-24
Fix installation problems

charmoniumQ:
- Remove unnecessary Vulkan stuff that was holding up the integration. It will be commented out so that we can add it in later.
- Add new CMake

### DIFF
--- a/install_deps.sh
+++ b/install_deps.sh
@@ -32,8 +32,8 @@ then
 			libeigen3-dev libboost-all-dev libatlas-base-dev libsuitesparse-dev libblas-dev \
 			glslang-tools libsdl2-dev libglu1-mesa-dev mesa-common-dev freeglut3-dev libglew-dev glew-utils libglfw3-dev \
 			libusb-dev libusb-1.0 libudev-dev libv4l-dev libhidapi-dev \
-			build-essential libx11-xcb-dev libxkbcommon-dev libwayland-dev libxrandr-dev \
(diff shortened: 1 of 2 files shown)


Hyjale:
We still need the `Vulkan-Headers` according to the [Monado dependencies](https://gitlab.freedesktop.org/monado/monado). `Vulkan-Loader` was the one that was causing issues and I ran Monado fine without it.

install_deps.sh
```diff
@@ -24,7 +26,7 @@ then
         git clone --branch 3.4.6 https://github.com/opencv/opencv/
         git clone --branch 3.4.6 https://github.com/opencv/opencv_contrib/
         mkdir -p opencv/build && cd opencv/build
-        cmake -DOPENCV_EXTRA_MODULES_PATH=../../opencv_contrib/modules ..
+        cmake -D CMAKE_BUILD_TYPE=Release -D CMAKE_INSTALL_PREFIX=/usr/local -DBUILD_TESTS=OFF -DBUILD_PERF_TESTS=OFF -DBUILD_EXAMPLES=OFF -DBUILD_JAVA=OFF -DWITH_OPENGL=ON -DOPENCV_EXTRA_MODULES_PATH=../../opencv_contrib/modules ..
```
apache8080, line 29: Nit: Do we let the user's know that we are installing opencv into their root directory?

charmoniumQ: Install deps now asks for permission before modifying the user's system.

PR #59, opened 2020-07-27
Fix OpenVINS Makefile

charmoniumQ:
Jae and Jeffrey reported errors with this. It seems to work or not work depending on the default C/C++ compiler. Some of us changed that, so we didn't notice this error.

I fixed this by setting the C/C++ compiler to Clang/Clang++ (respectively) by default. However, this default respects the CC/CXX environment variable, so it is overridable in special cases where one wants this.


apache8080:
Was the change made in the open_vins repo? I didn't see a PR there for this change. In the future when we make changes in the submodules and we are bumping the version here we should link to the PR in the submodule's source repo.

charmoniumQ:
I agree we have to work out a less frictionful process for changing submodules. So far, we've been updating the non-master branch of the submodule and then submitting a PR to change the commit pointer of the man repo, which has the advantage of fewer PRs to make. This is also for changes that don't need to go in master of OpenVINS, because they are specific to the ILLIXR integration.

PR #31, opened 2020-07-30
Implement SQLite metrics-logging backend

charmoniumQ:
- Should use transactions and prepared statements for bulk insertion.
- Should use a thread-per-table, exploiting parallelism between tables and parallelism from the main application.
  - This necessitates a file-per-table as well (since SQLite has no concurrency within one database, just make each table its own database).


charmoniumQ:
Abandoned. Better version coming soon.

PR #64, opened 2020-08-13
Kill subprocesses on failure

charmoniumQ:
- Will discuss after Aug 21.


charmoniumQ:
Reviewers: Previously, if one subprocess failed, the script would die while the other subprocesses were still running, because they ignore cancellation.

Now each subprocess respects cancellation.

charmoniumQ:
Coincidentally, I fix all three of these issues in [an unrelated PR](https://github.com/ILLIXR/ILLIXR/pull/78).

mhuzai (approved):

apache8080 (approved):
LGTM

PR #77, opened 2020-08-31
Add new logging interface

charmoniumQ:
Note to reviewers:
- A few weeks ago, I changed the logging infrastructure. The old one required code-generation which was difficult to use. I had a eureka that `std::any` could be used instead of my manual class-reflection.
  - Records are now a vector of `std::any`.
  - The logger-backend has to know what concrete type the `std::any` came from, each record also carries a pointer to a `record_header`, which holds the type of each field.
  - This is a pointer so that all of the records of that type read-only-share **one** `record_header`, saving us memory.
- I also took the opportunity to execute some renames `metric_logger` (applied inconsistently) -> `record_logger`, which has `record`s, `record_header`s, and `record_coalescer`s.


common/record_logger.hpp
```diff
@@ -0,0 +1,317 @@
...
+			// Check pointer first
+			if (this == &other) {
+				return true;
+			}
+
+			if (false
+				|| name != other.name
+				|| columns.size() != other.columns.size()
+				|| id != other.id
+				) {
```
apache8080, line 39: Is the `if (false` necessary?

common/record_logger.hpp
```diff
@@ -0,0 +1,317 @@
...
+		std::string name;
+		const std::vector<std::pair<std::string, const std::type_info&>> columns;
+	};
+
+	/**
+	 * @brief A helper class that lets one dynamically determine if some data gets used.
+	 *
+	 * When a use_taint gets copied, the original is considered used and the new one is considered unused.
+	 */
+	class use_taint {
```
apache8080, line 78: Nit: Can we use another word instead of taint? `use_taint` doesn't obviously describe what the class/function does.

common/record_logger.hpp
charmoniumQ: No, but it makes the other conditions all start with `|| `.

common/record_logger.hpp
charmoniumQ: I'm alluding to [this](https://en.wikipedia.org/wiki/Taint_checking); I will look for a better word.

mhuzai (approved):
Let's do this.

PR #79, opened 2020-09-01
Make exit cleanly

charmoniumQ:
Note to reviewers:
- Verify:
  - In `runtime/main.cpp:main`, either `cancellable_sleep` xor `signal_handler` will call `runtime_impl::stop`.
  - `runtime_impl::stop`, we call `switchboard::stop` and each plugin's `plugin::stop`.
  - `switchboard::stop` stops the switchboard threads.
  - Because `plugin::stop` is virtual, the call could go to `plugin` or to any child of `plugin`.
    - `threadloop` is a child of `plugin`, so it goes to `threadloop::stop`, which stops the plugin's individual thread, and then goes to `plugin::stop`.
    - There are no other children of `plugin`.
  - In cases like `zed`, where it has resources to clean up, it would override `plugin::stop` or `threadloop::stop`, clean up resources, and then call the thing it overrode.
- Test:
  - Drop down the delay in `main.cpp`.
  - Run `./runner.sh configs/native.yaml`.
  - It should exit without an error message.
  - `echo $?` should return 0.

### DIFF
--- a/common/extended_window.hpp
+++ b/common/extended_window.hpp
@@ -4,6 +4,7 @@
 #include <X11/Xlib.h>
 #include <GL/glx.h>
 #include <GL/glu.h>
+#include "phonebook.hpp"
 
 //GLX context magics
 #define GLX_CONTEXT_MAJOR_VERSION_ARB       0x2091
--- a/runtime/main.cpp
+++ b/runtime/main.cpp
@@ -40,7 +40,7 @@ int main(int argc, const char * argv[]) {
 	// And timer
 	cancellable_sleep cs;
 	std::thread th{[&]{
-		cs.sleep(std::chrono::seconds(10));
+		cs.sleep(std::chrono::seconds(60));
 		r->stop();
 	}};
 
--- a/runtime/main.cpp
+++ b/runtime/main.cpp
@@ -1,10 +1,55 @@
+#include <signal.h>
 #include "runtime_impl.hpp"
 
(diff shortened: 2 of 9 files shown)


common/extended_window.hpp
```diff
@@ -4,6 +4,7 @@
 #include <X11/Xlib.h>
 #include <GL/glx.h>
 #include <GL/glu.h>
+#include "phonebook.hpp"
```
mhuzai, line 7: Why is this include required?

runtime/main.cpp
```diff
@@ -1,10 +1,55 @@
...
-	r->wait();
+
+	// Two ways of shutting down:
+	// Ctrl+C
+	signal(SIGINT, signal_handler);
+
+	// And timer
+	cancellable_sleep cs;
+	std::thread th{[&]{
+		cs.sleep(std::chrono::seconds(10));
```
mhuzai, line 43: Can we keep the default 60 seconds?

common/extended_window.hpp
charmoniumQ: `phonebook::service` on line 15.

Suppose some file `A` includes this file, `extended_window.hpp`. `A` included `phonebook.hpp` for unrelated reasons, prior to including `extended_window.hpp`. That is why it was not causing an error.

runtime/main.cpp
charmoniumQ: Done

Issue #82, opened 2020-09-02
Add CI/CD workflow


apache8080:
A workflow needs to run on branches that builds and pushes docker images to ghcr.io and runs the tests in ILLIXR.

PR #83, opened 2020-09-03
#82 Add CI/CD Workflow

apache8080:
Handles #82


Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM ubuntu:18.04
+
+COPY . $HOME/ILLIXR/
```
mhuzai: How does this actually work? Does the GitHub server always puts ILLIXR in $HOME/ ?

install_deps.sh
```diff
@@ -73,37 +55,20 @@ then
...
-			-D CMAKE_INSTALL_PREFIX=install
-		sudo make -C "${temp_dir}/Vulkan-Headers/build" "-j$(nproc)" install
+		./scripts/install_vulkan_headers.sh
 	fi
 
 	if [ ! -d "${opt_dir}/googletest" ] && y_or_n "Next: Install gtest"; then
-		git clone https://github.com/google/googletest --branch release-1.10.0 "${opt_dir}/googletest"
-		cmake -S "${opt_dir}/googletest" -B "${opt_dir}/googletest/build"
-		make -C "${opt_dir}/googletest/build" "-j$(nproc)"
+		./scripts/install_gtest.sh
```
mhuzai, line 71: How are `opt_dir` and `temp_dir` being passed on to the child scripts?

install_deps.sh
```diff
@@ -119,26 +84,20 @@ then
 	# fi
 
 	if [ ! -d "${temp_dir}/OpenXR-SDK" ] && y_or_n "Next: Install OpenXR SDK from souce"; then
-		git clone https://github.com/KhronosGroup/OpenXR-SDK.git "${temp_dir}/OpenXR-SDK"
-		cmake -S "${temp_dir}/OpenXR-SDK" -B "${temp_dir}/OpenXR-SDK/build"
-		sudo make -C "${temp_dir}/OpenXR-SDK/build" "-j$(nproc)" install
+		./scripts/install_openxr.sh
```
mhuzai, line 87: Same question as above about passing `temp_dir` in.

install_deps.sh
apache8080: These files now get sourced so the variables get passed through.

Dockerfile
apache8080: The github server does a checkout of the PRs branch and then this line in the dockerfile copies that checked out code into the image.

mhuzai:
Why is the test still failing?

scripts/install_openxr.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+git clone https://github.com/KhronosGroup/OpenXR-SDK.git "${temp_dir}/OpenXR-SDK"
+echo "TEST"
```
mhuzai, line 4:
```suggestion
```

.github/workflows/main.yaml
```diff
@@ -1,25 +1,41 @@
...
+
+      - name: Get Branch Name
+        run:  echo "##[set-output name=branch;]$(echo ${GITHUB_REF#refs/heads/})"
+        id: extract_branch
+
+      - name: Login to ghcr.io
+        run: echo ${{ secrets.GHCR_TOKEN }} | docker login ghcr.io -u illixr-ci --password-stdin
+
+      - name: Pull Docker Cache
+        run: docker pull --all-tags ghcr.io/illixr/illixr-tests || true
```
mhuzai, line 25: What is the `true` for?

mhuzai (commented):
`main.yaml`'s future users would greatly benefit from comments. You can figure out what is going on by reading the "name:" fields, but actual comments describing what is going on would be great.

.github/workflows/main.yaml
charmoniumQ: If pulling fails, we can still build this package (without the benefit of the cache).

charmoniumQ:
> Why is the test still failing?

I am guessing because GitHub does not init/update submodules.

charmoniumQ (approved):

mhuzai (approved):

PR #85, opened 2020-09-03
Add zed

charmoniumQ:


zed/CMakeLists.txt
```diff
@@ -0,0 +1,52 @@
...
+SET(EXECUTABLE_OUTPUT_PATH ".")
+
+find_package(ZED 3 REQUIRED)
+find_package(OpenCV REQUIRED)
+find_package(CUDA ${ZED_CUDA_VERSION} EXACT REQUIRED)
+
+IF(NOT WIN32)
+    add_definitions(-Wno-format-extra-args)
+    SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -pthread")
+    SET(SPECIAL_OS_LIBS "pthread" "X11")
```
mhuzai, line 25: Why does ZED need X11?

zed/CMakeLists.txt
```diff
@@ -0,0 +1,52 @@
...
+
+link_directories(${ZED_LIBRARY_DIR})
+link_directories(${OpenCV_LIBRARY_DIRS})
+link_directories(${CUDA_LIBRARY_DIRS})
+
+FILE(GLOB_RECURSE SRC_FILES src/main.cpp)
+FILE(GLOB_RECURSE HDR_FILES include/*.hpp)
+
+ADD_EXECUTABLE(${execName} ${HDR_FILES} ${SRC_FILES})
+add_definitions(-std=c++17 -g -O3)
```
mhuzai, line 41: Shouldn't `DCMAKE_BUILD_TYPE` take care of setting -g and -O3?

zed/CMakeLists.txt
```diff
@@ -0,0 +1,52 @@
...
+if(COMMAND cmake_policy)
+	cmake_policy(SET CMP0003 OLD)
+	cmake_policy(SET CMP0015 OLD)
+endif(COMMAND cmake_policy)
+
+SET(EXECUTABLE_OUTPUT_PATH ".")
+
+find_package(ZED 3 REQUIRED)
+find_package(OpenCV REQUIRED)
+find_package(CUDA ${ZED_CUDA_VERSION} EXACT REQUIRED)
```
mhuzai, line 20: This reminds me: ZED SDK needs to be part of `install_deps.sh` now. But this would be tricky because ZED SDK has very specific requirements for OS and CUDA version. So we'll have to detect which OS and CUDA the user already has. And speaking of CUDA, should we add that as a dependency as well or leave that up to the user? We have so far been assuming that the user already has it.

zed/src/main.cpp
```diff
@@ -0,0 +1,238 @@
...
+  {
+    // Image setup
+    imageL_zed.alloc(image_size.width, image_size.height, MAT_TYPE::U8_C4, MEM::CPU);
+    imageR_zed.alloc(image_size.width, image_size.height, MAT_TYPE::U8_C4, MEM::CPU);
+
+    imageL_ocv = slMat2cvMat(imageL_zed);
+    imageR_ocv = slMat2cvMat(imageR_zed);
+
+		// experimental
+		runtime_parameters.enable_depth = false;
```
mhuzai, line 81: @Hyjale , is this still required?

zed/src/main.cpp
```diff
@@ -0,0 +1,238 @@
...
+      // Retrieve images
+      zedm->retrieveImage(imageL_zed, VIEW::LEFT, MEM::CPU, image_size);
+      zedm->retrieveImage(imageR_zed, VIEW::RIGHT, MEM::CPU, image_size);
+
+	  auto start_cpu_time  = thread_cpu_time();
+	  auto start_wall_time = std::chrono::high_resolution_clock::now();
+
+      // Conversion
+      // cv::cvtColor(imageL_ocv, grayL_ocv, CV_BGR2GRAY);
+      // cv::cvtColor(imageR_ocv, grayR_ocv, CV_BGR2GRAY);
```
mhuzai, line 118: Shouldn't there be a corresponding change in OV where it does this conversion itself?

mhuzai (commented):
Left a few comments but didn't actually test the code yet.

zed/src/main.cpp
```diff
@@ -0,0 +1,238 @@
...
+    time_type t;
+    ullong imu_time;
+
+	  std::size_t last_serial_no {0};
+};
+
+// This line makes the plugin importable by Spindle
+PLUGIN_MAIN(zed_imu_thread);
+
+int main(int argc, char **argv) { return 0; }
```
mhuzai, line 238: Why is this required?

zed/src/main.cpp
Hyjale: Not required.

zed/CMakeLists.txt
Hyjale: We should leave the ZED and CUDA installation to the user. I will add a README to the ZED plugin on how to do this. We should add a section for ZED in the ILLIXR docs as well.

zed/CMakeLists.txt
charmoniumQ: Try building it without. We will remove if it succeeds.

zed/CMakeLists.txt
charmoniumQ: I removed this.

zed/Makefile
```diff
@@ -0,0 +1,25 @@
...
+plugin.opt.so:
+	mkdir -p build/ && \
+	cd build/ && \
+	cmake .. -DDNDEBUG=1 -DCMAKE_BUILD_TYPE=Release .. && \
+	make "-j$(nproc)" && \
+	cd ../ && \
+	rm -f plugin.opt.so && \
+	ln -s build/libplugin.so plugin.opt.so && \
+	true
+
```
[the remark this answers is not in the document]
charmoniumQ: Added.

zed/src/main.cpp
charmoniumQ: Removed.

zed/src/main.cpp
charmoniumQ: Reverted.

zed/src/main.cpp
charmoniumQ: Removed.

zed/CMakeLists.txt
mhuzai: Why leave them up to the user? We install every other dependency for them, so why leave these two?

zed/CMakeLists.txt
```diff
@@ -37,6 +37,8 @@ link_directories(${CUDA_LIBRARY_DIRS})
 FILE(GLOB_RECURSE SRC_FILES src/main.cpp)
 FILE(GLOB_RECURSE HDR_FILES include/*.hpp)
 
+ADD_EXECUTABLE(${libName} ${HDR_FILES} ${SRC_FILES})
```
mhuzai, line 39: Why is this necessary?

zed/Makefile
```diff
@@ -20,13 +20,13 @@ plugin.opt.so: build/Release/Makefile
 build/Debug/Makefile:
 	mkdir -p build/Debug && \
 	cd build/Debug && \
-	cmake -DCMAKE_BUILD_TYPE=Debug  -DCMAKE_CXX_COMPILER=$(CXX) -DCMAKE_C_COMPILER=$(CC) ../.. && \
+	cmake -DCMAKE_BUILD_TYPE=Debug  -DCMAKE_CXX_COMPILER=$(CXX) -DCMAKE_C_COMPILER=$(CC) -DCMAKE_LIBRARY_PATH=/usr/local/cuda/lib64/stubs ../.. && \
```
mhuzai, line 23: The user's CUDA may not be here. Furthermore, why is this required in the first place?

zed/README.md
```diff
@@ -7,7 +7,8 @@
 
 ### Prerequisites
 
-- Windows 10, Ubuntu LTS
-- [ZED SDK](https://www.stereolabs.com/docs/installation/linux/) and its dependencies ([CUDA](https://developer.nvidia.com/cuda-downloads), [OpenCV](https://github.com/opencv/opencv/releases))
+- Ubuntu 18.04
+- [ZED SDK](https://www.stereolabs.com/docs/installation/linux/) 
+- [CUDA 10.2](https://developer.nvidia.com/cuda-toolkit-archive)
```
mhuzai, line 12: CUDA 11 runs fine as well -- I have tested it. We shouldn't limit people to 10.2.

mhuzai (approved):

Hyjale (approved):

PR #87, opened 2020-09-03
Newtonian pose prediction

charmoniumQ:


pose_prediction/plugin.cpp
```diff
@@ -10,31 +12,84 @@ class pose_prediction_impl : public pose_prediction {
...
+        //     future_timestamp,
+        //     Eigen::Vector3f{static_cast<float>(0), static_cast<float>(std::cos(time)), static_cast<float>(0)}, 
+        //     testQ
+        // };
+        // return fast_pose_type {
+        //     .pose = correct_pose(*test_pose),
+        //     .imu_time = std::chrono::high_resolution_clock::now(),
+        //     .predict_computed_time = std::chrono::high_resolution_clock::now(),
+        //     .predict_target_time = future_timestamp
+        // };
```
apache8080, line 54: If we want to keep this can we make it a function; if not lets delete this.

pose_prediction/plugin.cpp
```diff
@@ -99,9 +156,203 @@ class pose_prediction_impl : public pose_prediction {
...
+        Eigen::Matrix3d R_Gto1 = quat_2_Rot(quat_multiply(dq_1,q_0));
+        Eigen::Vector3d v1_dot = R_Gto1.transpose()*a_hat-Eigen::Vector3d{0.0, 0.0, 9.81};
+
+        Eigen::Vector4d k2_q = q1_dot*dt;
+        Eigen::Vector3d k2_p = p1_dot*dt;
+        Eigen::Vector3d k2_v = v1_dot*dt;
+
+        // k3 ================
+        Eigen::Vector4d dq_2 = quatnorm(dq_0+0.5*k2_q);
+        //Eigen::Vector3d p_2 = p_0+0.5*k2_p;
```
apache8080, line 212: Nit: remove this extraneous comment

pose_prediction/plugin.cpp
```diff
@@ -99,9 +156,203 @@ class pose_prediction_impl : public pose_prediction {
...
 
 		swapped_pose.orientation = apply_offset(raw_o);
+        swapped_pose.sensor_time = pose.sensor_time;
 
         return swapped_pose;
     }
+
+    // Slightly modified copy of OpenVINS method found in propagator.cpp
+    // Returns a pair of the predictor state_plus and the time associated with the
+    // most recent imu reading used to perform this prediction.
```
apache8080, line 166: @mhuzai If we are modifying a function from OpenVINS should we do it actually in OpenVINS?

apache8080 (commented):
I think I am mostly fine with this. Just a few small comments. One question I had is audio_pipeline is getting modified in this PR, is that because you merged master?

JeffreyZh4ng:
> I think I am mostly fine with this. Just a few small comments. One question I had is audio_pipeline is getting modified in this PR, is that because you merged master?

Yeah I made a change in earlier commit and master had the same change. Otherwise there were no changes to audio_pipeline

pose_prediction/plugin.cpp
JeffreyZh4ng: In this case I dont think so because this change make this method read switchboard plugs which we obviously dont want to be doing in OpenVINS

apache8080 (dismissed):

common/data_format.hpp
```diff
@@ -35,18 +35,34 @@ namespace ILLIXR {
 		ullong dataset_time;
 	} imu_cam_type;
 
+	typedef struct {
+		Eigen::Matrix<double,3,1> w_hat;
+		Eigen::Matrix<double,3,1> a_hat;
+		Eigen::Matrix<double,3,1> w_hat2;
+		Eigen::Matrix<double,3,1> a_hat2;
+		Eigen::Matrix<double,13,1> pose;
```
mhuzai, line 43: Isn't this is `state` (based on [this line](https://github.com/ILLIXR/open_vins/compare/50d1d5a2869b8d6bc9a276e984d1a2498c59c077...e27874535235c44d9999dfd383cc8bc7170b0397?diff=split#diff-18e35e680b5b2b4270f745b227d7aa49R189))? That is, why call it `pose`?

common/data_format.hpp
```diff
@@ -35,18 +35,34 @@ namespace ILLIXR {
...
+		time_type imu_time;
+	} imu_biases_type;
```
mhuzai, line 45: These are not biases. They are actual IMU readings. Rename both the datatype and the variable name (here as well as in OV).

pose_prediction/plugin.cpp
```diff
@@ -10,31 +12,62 @@ class pose_prediction_impl : public pose_prediction {
...
     { }
 
-    virtual pose_type get_fast_pose() const override {
-		const pose_type* pose_ptr = _m_slow_pose->get_latest_ro();
-		return correct_pose(
-			pose_ptr ? *pose_ptr : pose_type{}
-		);
+    // No paramter pose predict will just get the current slow pose based on the next vsync
+    virtual fast_pose_type get_fast_pose() override {
+		return get_fast_pose(std::chrono::high_resolution_clock::now());
```
mhuzai, line 21: Isn't the no parameter version supposed to return the pose for the next vsync? If we are not doing that because the timewarp vsync PR isn't here yet, then update the comment for now so it says that we return the pose for `now` (which may still be in the future with regards to the last IMU sample).

pose_prediction/plugin.cpp
```diff
@@ -10,31 +12,62 @@ class pose_prediction_impl : public pose_prediction {
...
     }
 
+    // future_time: An absolute timepoint in the future
+    virtual fast_pose_type get_fast_pose(time_type future_timestamp) override {
+
+        if (!_m_imu_biases->get_latest_ro()) {
+            const pose_type* pose_ptr = _m_slow_pose->get_latest_ro();
+            
+            return fast_pose_type{
+                .pose = correct_pose(pose_ptr ? *pose_ptr : pose_type{})
```
mhuzai, line 38: Why does this not set the computed time and target time? The caller has no way of knowing what pose prediction did, and will use the times regardless.

mhuzai (commented):
OV:
1. Why do you need to include `opencv2/core/utility.hpp` and `pose_prediction.hpp` in `slam2.cpp`?
2. In Propagator.h, fix [this comment](https://github.com/ILLIXR/open_vins/compare/50d1d5a2869b8d6bc9a276e984d1a2498c59c077...e27874535235c44d9999dfd383cc8bc7170b0397?diff=split#diff-878829c1dbfae37cf1dc0f070677af16R180). It doesn't match with what the function prototype says. Also say that it is a return value, not an input. Also, [this comment](https://github.com/ILLIXR/open_vins/compare/50d1d5a2869b8d6bc9a276e984d1a2498c59c077...e27874535235c44d9999dfd383cc8bc7170b0397?diff=split#diff-18e35e680b5b2b4270f745b227d7aa49R119) is not really required, but this is minor.

Why is hologram being updated? There is no diff.

pose_prediction/plugin.cpp
charmoniumQ: This was lifted from OpenVINS, so I'm thinking not modifying will reduce the diff-drift of our copy from the original.

apache8080 (commented):
Is[ this ILLIXR::imu_raw_types parameter in Propagator.cpp](https://github.com/ILLIXR/open_vins/compare/ea6ee1d48a52334f424ca8d355d49c090f563083...3cf3d3884d4281448a98d405314106461d30c451#diff-18e35e680b5b2b4270f745b227d7aa49R119) supposed to be called biases because it looks like this is just raw imu data propagated forward?

charmoniumQ:
> Is[ this ILLIXR::imu_raw_types parameter in Propagator.cpp](https://github.com/ILLIXR/open_vins/compare/ea6ee1d48a52334f424ca8d355d49c090f563083...3cf3d3884d4281448a98d405314106461d30c451#diff-18e35e680b5b2b4270f745b227d7aa49R119) supposed to be called biases because it looks like this is just raw imu data propagated forward?

Fixed (`imu_raw_type biases` -> `imu_raw_type imu_data`).

mhuzai (commented):
[This](https://github.com/ILLIXR/open_vins/compare/ea6ee1d48a52334f424ca8d355d49c090f563083...f11ff192fce57f3c5c3926c099d4de87f22a7bf2#diff-878829c1dbfae37cf1dc0f070677af16R180) and the variable name still says biases.

charmoniumQ:
> [This](https://github.com/ILLIXR/open_vins/compare/ea6ee1d48a52334f424ca8d355d49c090f563083...f11ff192fce57f3c5c3926c099d4de87f22a7bf2#diff-878829c1dbfae37cf1dc0f070677af16R180) and the variable name still says biases.

Fixed.

apache8080 (approved):

mhuzai (dismissed):

apache8080 (approved):

JeffreyZh4ng (approved):

PR #84, opened 2020-09-03
Minor changes

charmoniumQ:

PR #90, opened 2020-09-03
Add audio decoding

charmoniumQ:
- Approve after: #84


apache8080 (dismissed):

mhuzai:
I am unable to test this. Checking out the branch results in:
`fatal: not a git repository: audio_decoding/libspatialaudio/../../.git/modules/audio_decoding/libspatialaudio`

JeffreyZh4ng (commented):
LGTM. Illixr runs fine - verified that the component was being run

JeffreyZh4ng (dismissed):

mhuzai (approved):
As discussed, I tested the gitmodules fix on my home machine and it worked fine. This PR is good to go.

mhuzai:
@apache8080 brought up a good point that "audio_pipeline" is now misleading because it only contains encoding. One possible solution to this problem is to have just the `audio_pipeline` submodule, but have it internally launch two threads, one for encoding and one for playback (we should really name it playback and not decoding), just like ZED launches two threads. That would ensure that the term "audio_pipeline" still makes sense and we still get our two different threads. It would also remove the need for symlinks. Thoughts, @charmoniumQ?

charmoniumQ:
I think that is cleaner and it only builds `libspatialaudio` once, so I implemented to that design.

configs/monado.yaml
```diff
@@ -7,6 +7,7 @@ plugins:
   - path: timewarp_gl/
   - path: debugview/
   - path: audio_pipeline/
+  - path: audio_decoding/
```
mhuzai, line 10:
```suggestion

```

mhuzai (commented):
Wouldn't it be better to have an audio base class that both threads inherit from to minimize the duplication of code?

charmoniumQ:
> Wouldn't it be better to have an audio base class that both threads inherit from to minimize the duplication of code?

That's a good point; I updated the implementation (actually two instances, one class).

mhuzai (approved):

mhuzai (approved):

apache8080 (approved):

PR #57, opened 2020-09-04
Run ILLIXR headlessly

mhuzai:
We would like to be able to run without a physical display. This is useful during SSH connections and useful for profiling.


timewarp_gl/shaders/basic_shader.hpp
```diff
@@ -1,6 +1,6 @@
 #include <GL/gl.h>
 
-#define GLSL_VERSION            "430 core"
+#define GLSL_VERSION            "330 core"
```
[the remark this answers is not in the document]
mhuzai: Nice idea. I can't decide which option is better. Maybe 2 because 1 would require a new file that just has one line in it?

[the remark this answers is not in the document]
charmoniumQ: I've implemented 2.

charmoniumQ:
I'm approving this PR. However, I recommend two future changes:

1. After #112 is merged, use the external tool loader, instead of a dedicated loader.
2. Run headless in the CI pipeline. This will inform us about potentially-introduced runtime errors (i.e. segfaults).

charmoniumQ (commented):
I've only made changes to the CI, and those are minor, so I am approving this.

timewarp_gl/shaders/timewarp_shader.hpp
```diff
@@ -1,7 +1,5 @@
 #include <GL/gl.h>
 
-#define GLSL_VERSION            "430 core"
-
 const char* const timeWarpChromaticVertexProgramGLSL =
 	"#version " GLSL_VERSION "\n"
```
Zee2, line 4: `GLSL_VERSION` macro is still being used in the shader source despite the `#define` being removed above?

charmoniumQ: It's being defined at the command-line invocation of GCC. This way, it doesn't have to be duplicated in every shader.

mhuzai: After seeing Finn's comment I'm wondering if putting the GLSL version in the makefile is the most readable choice. Perhaps it should be in an include of its own.

docs/getting_started.md
```diff
@@ -48,20 +52,11 @@ of applications.  Monado only supports Ubuntu 18.10, because of a low-level driv
...
-    path: ../monado_integration
-  openxr_app:
-    path: ../Monado_OpenXR_Simple_Example
-```
+5. Inspect `configs/monado.yaml`. The schema definition (with documentation inline) is in `runner/config_schema.yaml`.
 
-5. Compile and run:
+6. Compile and run:
 
-        ./runner.sh config.yaml
```
charmoniumQ, line 64: :+1: for fixing out-of-date documentation

timewarp_gl/shaders/timewarp_shader.hpp
charmoniumQ: If you do, put it in `common/`.

common/extended_window.hpp
```diff
@@ -43,89 +48,105 @@ namespace ILLIXR{
...
-            win = XCreateWindow(dpy, root, 0, 0, width, height, 0, vi->depth, InputOutput, vi->visual, CWColormap | CWEventMask, &swa);
-            XMapWindow(dpy, win);
+
+#ifndef NDEBUG
+            printf("Creating window\n");
+#endif
+            XSetWindowAttributes attributes;
+            attributes.colormap = cmap;
+            attributes.background_pixel = 0;
+            attributes.border_pixel = 0;
```
charmoniumQ, line 110: Are these changes necessary for headless or something else?

common/extended_window.hpp
```diff
@@ -43,89 +48,105 @@ namespace ILLIXR{
...
-            int context_attribs[] =
-                {
-                    GLX_CONTEXT_MAJOR_VERSION_ARB, 4,
-                    GLX_CONTEXT_MINOR_VERSION_ARB, 0,
-                    //GLX_CONTEXT_FLAGS_ARB        , GLX_CONTEXT_FORWARD_COMPATIBLE_BIT_ARB,
-                    None
-                };
+                glXGetProcAddressARB((const GLubyte *) "glXCreateContextAttribsARB");
+            int context_attribs[] = {
+                GLX_CONTEXT_MAJOR_VERSION_ARB, 3,
```
charmoniumQ, line 131: Maybe this should also be defined for the project in an include file.

common/extended_window.hpp
mhuzai: This is _the_ fix for headless.

apache8080 (approved):
LGTM

PR #99, opened 2020-09-05
Download EuRoC in install_deps

mhuzai:
Previously, our makefile would download Vicon Room 1 Medium automatically if detected that it didn't exist. This is not done at the moment, and should be. Otherwise, the user would have to do it themselves, which is undesirable and causes an error the first time they try to run ILLIXR.
`install_deps.sh` is probably a better place to do this than `runner` since the dataset is a dependency of `offline_imu_cam`.

### DIFF
--- a/install_deps.sh
+++ b/install_deps.sh
@@ -88,11 +88,7 @@ then
 	fi
 
 	if [ ! -d "data1" ] && y_or_n "Next: Download Vicon Room 1 Medium SLAM dataset"; then
-		curl -o data.zip \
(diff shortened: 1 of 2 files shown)


charmoniumQ:
Redo after merging #83.

apache8080 (dismissed):

charmoniumQ (dismissed):

apache8080 (approved):

charmoniumQ (approved):

PR #98, opened 2020-09-05
Pull our OpenCV in install_deps

mhuzai:
We need to use our instrumented OpenCV instead of the vanilla one.

### DIFF
--- a/install_deps.sh
+++ b/install_deps.sh
@@ -74,7 +74,7 @@ then
 	sudo chown $USER: "${opt_dir}"
 
 	if [ ! -d "${temp_dir}/opencv" ] && y_or_n "Next: Install OpenCV from source"; then
-		git clone --branch 3.4.6 https://github.com/opencv/opencv/ "${temp_dir}/opencv"
+		git clone --branch 3.4.6-instrumented https://github.com/ILLIXR/opencv "${temp_dir}/opencv"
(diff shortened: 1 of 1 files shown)


charmoniumQ (dismissed):

apache8080 (dismissed):

charmoniumQ (approved):

scripts/install_opencv.sh
```diff
@@ -14,4 +14,9 @@ cmake \
 	-D WITH_OPENGL=ON \
 	-D OPENCV_EXTRA_MODULES_PATH="${temp_dir}/opencv_contrib/modules"
 sudo make -C "${temp_dir}/opencv/build" "-j$(nproc)" install
+
+# Load new library paths
 sudo ldconfig -v
+
+# Remove temp cmake files
+rm -f a.out cmake_hdf5_test.o
```
apache8080, line 22: Why is this necessary? Don't these files only appear if cmake failed.

mhuzai: These files are generated when `cmake` tests whether the compiler works or not. We didn't worry about them previously because we were invoking `cmake` from `/tmp/ILLIXR_deps` but now we invoke `cmake` from the main ILLIXR directory. `cmake` really should be cleaning up these files by itself but I couldn't find any documentation on this.

charmoniumQ: Seems reasonable.

charmoniumQ (approved):

apache8080 (approved):

PR #102, opened 2020-09-09
Sqlite Logging implementation

apache8080:


charmoniumQ (dismissed):

runtime/sqlite_record_logger.hpp
```diff
@@ -0,0 +1,205 @@
...
+	void pull_queue() {
+		const std::size_t max_record_batch_size = 1024 * 256;
+		std::vector<record> record_batch {max_record_batch_size};
+		std::size_t actual_batch_size;
+
+		std::cout << "thread," << std::this_thread::get_id() << ",sqlite thread," << table_name << std::endl;
+
+		std::size_t processed = 0;
+		while (!terminate.load()) {
+			std::this_thread::sleep_for(std::chrono::seconds{1});
```
mhuzai, line 88: Is it possible to do all of this _after_ terminate? That is, buffer everything until the very end. That will essentially ensure that there is no interference from the sqlite threads.

charmoniumQ: It does delay everything until after termination, because the body of this while loop is commented out.

mhuzai: I have no idea how I missed that.

mhuzai (commented):
Good to go once the memory leak is fixed.

mhuzai (approved):
I could not reproduce the issue on my home machine. After doing a clean build on the lab machine, I couldn't reproduce the issue there either. It seems that we ran into some weird partial build of audio that just hogged memory.

charmoniumQ (approved):

mhuzai (approved):

PR #36, opened 2020-09-10
Fix initial frame of reference problem

charmoniumQ:
- The initial viewpoint is not "straight ahead" and has to be zeroed out manually.


apache8080 (approved):

JeffreyZh4ng (dismissed):

pose_prediction/plugin.cpp
```diff
@@ -44,7 +63,7 @@ class pose_prediction_impl : public pose_prediction {
 	}
 
 	Eigen::Quaternionf apply_offset(const Eigen::Quaternionf& orientation) const {
-		std::lock_guard<std::mutex> lock {offset_mutex};
+		std::lock_guard<std::recursive_mutex> lock {offset_mutex};
```
mhuzai, line 66: This mutex is on the critical path of all pose queries. Since pose predict will run on whatever core is calling it, `offset_mutex` will bounce around. I understand this is required only because of a concurrent write to `offset` in `set_offset`, which can only come from debugview. Should we just get rid of that button from debugview and remove the mutex?

All of that being said, there might not be a perf hit at all because pose predict is only called a few hundred times a second, not a few thousand.

charmoniumQ: Resolved by using a `shared_mutex` which has a non-exclusive read state.

apache8080 (approved):

pose_prediction/plugin.cpp
```diff
@@ -18,12 +19,12 @@ class pose_prediction_impl : public pose_prediction {
...
-			std::lock_guard<std::recursive_mutex> lock {offset_mutex};
+			auto pose = correct_pose(*pose_ptr);
+			std::unique_lock lock {offset_mutex};
 			// check again, now that we have mutual exclusion
 			if (first_time) {
-				auto pose = correct_pose(*pose_ptr);
-				const_cast<pose_prediction_impl&>(*this).set_offset(pose.orientation);
-				const_cast<pose_prediction_impl&>(*this).first_time = false;
+				first_time = false;
+				offset = pose.orientation;
```
mhuzai: This diff is not functionally equivalent. Originally, `offset` was being updated in `set_offset` with some matrix math. Now, it just gets the value of the current pose and `set_offset` is no longer called. Why is that?

charmoniumQ: `offset` is the Identity matrix to begin with, so multiplying by the inverse is a no-op.

charmoniumQ: I can't call `set_offset` because the lock is no longer recursive (can't acquire it, and then call a method that acquires it).

I need to acquire it here, to lock out other "first time calls" to `pose_predict`; The lock picks exactly one call which "wins the race" to be the "first time call"

mhuzai: The locking is fine. That is not my concern. My concern is that `set_offset` returns `pose.orientation.inverse` the very first time because `offset` is the identity quaternion to begin with. The above line returns just `post.orientation`, without the inverse. What am I missing here?

mhuzai (dismissed):

pose_prediction/plugin.cpp
charmoniumQ: I think it should be inverted, so I changed this in the implementation.

Failing to invert the initial pose may give approximately correct behavior, if the initial pose was a quaternion, $q = q^{-1}$. There are many such quaternions, e.g. if the angular displacement is close to 0 or close to 180, on any plane.

mhuzai (approved):

apache8080 (approved):

mhuzai (approved):

PR #65, opened 2020-09-15
Implement Pose Lookup

charmoniumQ:
Review after #87.

- [X] Merge `zed-new` with Jeffrey's copy of `common/pose_prediction.hpp` (as of now, his branch is called `open_vins_pp`, but check with Jeffrey when you do this).
- [X] Create a plugin called `pose_lookup`, starting with a copy of `pose_prediction`.
- [X] Reuse `ground_truth_slam/data_loading.hpp`.

The class should do the following: 
- [X] constructor(): loads the data from ground_truth CSV file. See `offline_imu_cam/plugin.cpp` for how to find the data path (instead of hard-coding).
- [X] `get_true_pose()` -> `throw std::logic_error{"not implemented"}`
- [X] `true_pose_reliable()` -> `return false`. (Nobody should be asking for `true_pose`).
- [X] `fast_pose_reliable()` -> `return true`. (yihan: I assume this should be fast instead of true?)~
- [X] `get_fast_pose()` -> `return get_fast_pose(std::chrono::system_clock::now())`.
- [X] `get_fast_pose(time_type time)`: `return correct_pose(lookup(transposed_time))`.
  - [X] Transposition: `transposed_time = std::chrono::nanoseconds((pp->next_vsync(input_real_time) - plugin_start_real_time).count() + dataset_start_time`
    - The ground-truth map will contain poses at timestamps from 2014, but `now()` will be a time in 2020.
  - [X] Write `pose_type lookup(ullong timestamp)` looks up nearest pose (past or future) to the input `timestamp`.
    - (Jeffrey's `pp->next_vsync` handles this.) ~Assume vsync is an event that occurs once every 1/60 seconds.~
    - ~The time of the first vsync is arbitrary, as long as you are consistent.~
      - ~Any other thing based on real or simulated vsync will have the same frequency but different phase.~
      - ~This is OK, because it will still query pose-lookup once per vsync cycle, and pose-lookup will still return the "next" pose.~


pose_lookup/plugin.cpp
```diff
@@ -0,0 +1,172 @@
...
+		{
+			std::cout<<"found key"<<std::endl;
+		}
+		else
+		{
+			std::cout<<"key not found"<<std::endl;
+		}
+		auto temp = _m_sensor_data.find(nearest_timestamp)->second;
+		std::cout<<temp.position.x()<<", "<<temp.position.y()<<", "<<temp.position.z()<<", "<<temp.orientation.w()<<", "<< temp.orientation.x()<<", "<<temp.orientation.y()<<", "<<temp.orientation.z()<<std::endl;
+		*/
```
apache8080, line 114: Remove these comments if they aren't necessary.

pose_lookup/plugin.cpp
```diff
@@ -0,0 +1,172 @@
...
+    virtual fast_pose_type get_fast_pose(time_type time) override {
+		/*
+			ullong vsync_time = std::chrono::nanoseconds(get_vsync(time).time_since_epoch()).count();
+			ullong plug_in_time = std::chrono::nanoseconds(_m_start_of_time.time_since_epoch()).count();
+
+			std::cout<<"dataset_first: "<<dataset_first_time<<"vsync time: "<<vsync_time <<" plug_in_time: "<<plug_in_time<<std::endl;
+			std::cout<<"lookup time: "<<lookup_time <<std::endl;
+			std::cout<<"data end time: "<< _m_sensor_data.rbegin()->first <<std::endl;
+		*/
+	    	//set_offset(correct_pose(_m_sensor_data.begin()->second).orientation);
```
apache8080, line 62: Remove these comments if they aren't necessary.

pose_lookup/plugin.cpp
```diff
@@ -0,0 +1,172 @@
...
+
+		const time_type* estimated_vsync = _m_vsync_estimate->get_latest_ro();
+		time_type vsync;
+		if(estimated_vsync == nullptr) {
+			vsync = std::chrono::system_clock::now();
```
apache8080, line 67: Is making vsync now if it is null the right assumption? Should it be logging an error instead?

charmoniumQ: Let's log to stderr in addition. This condition would be true be if a plugin asks for pose _before the first vsync_. In such a case, getting a pose for "now" is an acceptable approximation.

apache8080 (dismissed):

configs/native.yaml
```diff
@@ -2,6 +2,7 @@ plugins:
   - path: ground_truth_slam/
   - path: offline_imu_cam/
   - path: open_vins/
+  - path: pose_lookup/
```
mhuzai, line 5: Both of these cannot be here. They both register with the same name and cause an assertion failure in phonebook. So we should comment out pose_lookup (uncommon case) and keep pose_prediction (common case).

pose_lookup/plugin.cpp
```diff
@@ -0,0 +1,158 @@
+#include "common/phonebook.hpp"
```
mhuzai, line 1: This entire file is about the same as pose_predict/plugin.cpp. Why the duplication? Why not inherit from pose_prediction_impl and just override the one or two methods that are necessary? I understand that _impl's are supposed to be internal, so we'll need some refactoring, but it might be worth it given that the code is pretty much identical.

pose_lookup/plugin.cpp
```diff
@@ -0,0 +1,158 @@
...
+		};
+
+	}
+
+
+private:
+	const std::shared_ptr<switchboard> sb;
+
+	mutable Eigen::Quaternionf offset {Eigen::Quaternionf::Identity()};
+	mutable std::mutex offset_mutex;
```
mhuzai, line 115: This needs to be a `shared_mutex`, like pose_predict, but also see the very first comment above.

mhuzai (commented):
Question about plugin naming: the class is called pose_lookup_plugin and thus so would the object factory. But the yaml file only says pose_lookup. So how is the component actually found?

As an aside, the performance of the ground truth pose is remarkable. Silky smooth images.

pose_lookup/plugin.cpp
```diff
@@ -0,0 +1,159 @@
...
+
+   virtual void set_offset(const Eigen::Quaternionf& raw_o_times_offset) override{
+	std::unique_lock lock {offset_mutex};
+	Eigen::Quaternionf raw_o = raw_o_times_offset * offset.inverse();
+	//std::cout << "pose_prediction: set_offset" << std::endl;
+	offset = raw_o.inverse();
+    }
+
+    Eigen::Quaternionf apply_offset(const Eigen::Quaternionf& orientation) const {
+	std::unique_lock lock {offset_mutex};
```
mhuzai, line 51:
```suggestion
	std::shared_lock lock {offset_mutex};
```

mhuzai (changes requested):

pose_lookup/plugin.cpp
charmoniumQ: Let's talk about this more after 2.2.

charmoniumQ:
> Question about plugin naming: the class is called pose_lookup_plugin and thus so would the object factory. But the yaml file only says pose_lookup. So how is the component actually found?

`pose_lookup/` is the path to the plugin, which is where the YAML file finds it. Spindle loads whatever class is in `PLUGIN_MAIN`.

mhuzai (dismissed):

apache8080 (dismissed):

PR #105, opened 2020-09-15
Gate prints for dbg only 

charmoniumQ:
Using Profile: opt in config/native.yaml will turn off logging.
Using Profile: dbg in config/native.yaml will turn it on.

However, errors should not be gated and stats that we collect with scripts should not be gated.

Review after #87 .


apache8080 (dismissed):

mhuzai (commented):
The libspatialaudio print still shows up, despite being guarded. I'm not sure why that is though. I tried a deepclean as well.

mhuzai (approved):

apache8080 (approved):

PR #123, opened 2020-09-17
Complete removal of ksAlgebra + timewarp improvements

### LINKED ISSUE #125: Long-standing TODO for strange matrix inverse behavior
There has been an outstanding TODO on timewarp since initial release regarding the behavior of the warp delta matrices and why the inverse was used; in addition, the timewarp shader performed strange, nonstandard matrix multiplication using 3x4 instead of 4x4 matrices. This TODO needs to be resolved and the strange shader behavior ought to be clarified.

Zee2:
We have been using/abusing `ksAlgebra` since the very beginning to serve our linear algebra needs. We had been gradually moving away from it over time, but quite a few places in the code still deeply depended on it. `ksAlgebra` was J.M.P. Van Waveren's personal linear algebra library that he used in his timewarp demos.

This PR entirely removes `ksAlgebra` and all associated files, references, and dependencies; it repairs existing components that depend on `ksAlgebra` to use custom implementations where necessary, and moves all `ksAlgebra` types to use the `Eigen` types and operations.

In addition, there was a long-outstanding TODO on Timewarp to clarify what was going on with the inverse/transpose operations, and why the timewarp shader was seemingly performing vector/matrix multiplication in the wrong direction. This PR cleans up those confusions as part of the migration away from `ksAlgebra` and towards `Eigen`.

This PR will resolve #125 and #124.


timewarp_gl/plugin.cpp
```diff
@@ -228,62 +226,40 @@ class timewarp_gl : public threadloop {
...
-		// Convert the projection matrix from [-1, 1] space to [0, 1] space.
-		const ksAlgebra::ksMatrix4x4f texCoordProjection =
-		{ {
-			{ 0.5f * renderProjectionMatrix->m[0][0],        0.0f,                                           0.0f,  0.0f },
-			{ 0.0f,                                          0.5f * renderProjectionMatrix->m[1][1],         0.0f,  0.0f },
-			{ 0.5f * renderProjectionMatrix->m[2][0] - 0.5f, 0.5f * renderProjectionMatrix->m[2][1] - 0.5f, -1.0f,  0.0f },
-			{ 0.0f,                                          0.0f,                                           0.0f,  1.0f }
-		} };
+		Eigen::Matrix4f texCoordProjection;
+		texCoordProjection <<  0.5f * (*renderProjectionMatrix)(0,0),         0.0f,                                          0.5f * (*renderProjectionMatrix)(0,2) - 0.5f,  0.0f ,
```
mhuzai, line 241: Add comment reminding readers that Eigen's default is column-major, not row-major.

common/math_util.hpp
```diff
@@ -0,0 +1,79 @@
+#pragma once
+
+namespace ILLIXR {
+	class math_util {
+		public:
+
+		static void projection_fov(Eigen::Matrix4f* result,
```
mhuzai, line 7: I know that this is the same logic as the ks algebra library, but it would be nice to have a reference for the math.

timewarp_gl/plugin.cpp
```diff
@@ -461,59 +426,31 @@ class timewarp_gl : public threadloop {
...
-		// We transform from 4x4 to 3x4 as we operate on vec3's in NDC space
-		ksAlgebra::ksMatrix3x4f timeWarpStartTransform3x4;
-		ksAlgebra::ksMatrix3x4f timeWarpEndTransform3x4;
-		ksAlgebra::ksMatrix3x4f_CreateFromMatrix4x4f( &timeWarpStartTransform3x4, &timeWarpStartTransform4x4 );
-		ksAlgebra::ksMatrix3x4f_CreateFromMatrix4x4f( &timeWarpEndTransform3x4, &timeWarpEndTransform4x4 );
-
-		// Push timewarp transform matrices to timewarp shader
-		glUniformMatrix3x4fv(tw_start_transform_unif, 1, GL_FALSE, (GLfloat*)&(timeWarpStartTransform3x4.m[0][0]));
-		glUniformMatrix3x4fv(tw_end_transform_unif, 1, GL_FALSE,  (GLfloat*)&(timeWarpEndTransform3x4.m[0][0]));
+		glUniformMatrix4fv(tw_start_transform_unif, 1, GL_FALSE, (GLfloat*)(timeWarpStartTransform4x4.data()));
```
mhuzai, line 452: I never really understood why 3x4 matrices were used in the first place, and now I don't understand why we can work with 4x4s. Can you clarify?

timewarp_gl/shaders/timewarp_shader.hpp
```diff
@@ -50,6 +53,7 @@ const char* const timeWarpChromaticFragmentProgramGLSL =
 	"	outColor.r = texture( Texture, vec3( fragmentUv0, ArrayLayer ) ).r;\n"
 	"	outColor.g = texture( Texture, vec3( fragmentUv1, ArrayLayer ) ).g;\n"
 	"	outColor.b = texture( Texture, vec3( fragmentUv2, ArrayLayer ) ).b;\n"
+	"	outColor.rgb = vec3(fragmentUv0,1);\n"
```
mhuzai, line 56: This overwrites the actual color output. Even though this shader is not used, this bug shouldn't be here.

timewarp_gl/shaders/timewarp_shader.hpp
```diff
@@ -35,6 +35,9 @@ const char* const timeWarpChromaticVertexProgramGLSL =
 	"	fragmentUv0 = curUv0.xy * ( 1.0 / max( curUv0.z, 0.00001 ) );\n"
 	"	fragmentUv1 = curUv1.xy * ( 1.0 / max( curUv1.z, 0.00001 ) );\n"
 	"	fragmentUv2 = curUv2.xy * ( 1.0 / max( curUv2.z, 0.00001 ) );\n"
+	// "	fragmentUv0 = curUv0.xy;\n"
+	// "	fragmentUv1 = curUv1.xy;\n"
+	// "	fragmentUv2 = curUv2.xy;\n"
```
mhuzai, line 40: Remove if not needed anymore.

timewarp_gl/plugin.cpp
```diff
@@ -228,62 +226,40 @@ class timewarp_gl : public threadloop {
...
 		// This was just temporary.
 		free(tw_mesh_base_ptr);
 
 		return;
 	}
 
 	/* Calculate timewarm transform from projection matrix, view matrix, etc */
-	void CalculateTimeWarpTransform( ksAlgebra::ksMatrix4x4f * transform, const ksAlgebra::ksMatrix4x4f * renderProjectionMatrix,
-                                        const ksAlgebra::ksMatrix4x4f * renderViewMatrix, const ksAlgebra::ksMatrix4x4f * newViewMatrix )
+	void CalculateTimeWarpTransform( Eigen::Matrix4f * transform, const Eigen::Matrix4f * renderProjectionMatrix,
```
mhuzai, line 237: Better to obtain references (e.g., `&transform`) to the parameters and not worry about asterisks later on.

timewarp_gl/plugin.cpp
Zee2: @mhuzai the initializer here (using the `<<` syntax) is actually storage-independent. The order in which the initializers are fed into the `<<` stream represents the actual structure of the matrix.

Are you talking about the `(*renderProjectionMatrix)(i,j)` accessors? If so, those are row-major. I.e., the first argument is which row, and the second argument is which column. Quoted from Eigen documentation:

> The primary coefficient accessors and mutators in Eigen are the overloaded parenthesis operators. **For matrices, the row index is always passed first.**

common/math_util.hpp
Zee2: Will do.

timewarp_gl/plugin.cpp
Zee2: This is a good one. I wasn't really sure either, but then I went down a long rabbit hole of linear algebra.

Previously, Van Waveren's code was using this weird GLSL hack where you could perform a transpose operation by multiplying the matrix on the right instead of the left. GLSL "flexibly" interprets vectors as either rows or columns depending on which side you multiply the matrix on; this allows you to do an implicit transpose operation on the matrix. 3x4 matrices were used because it was a right-multiplication on a vec3.

This change is essentially just removing this weird hack and doing all of the algebra "properly". It may be removing some tiny performance optimization (because 3x4 uses one fewer vector register, or something like that), but as far as I can read online, using 4x4 instead of 3x4 (i.e. removing the micro-optimization) has little effect, if any.

In my opinion, using standard 4x4 matrices with a standard multiplication order instead of the weird hack that was implemented previously has benefits for readability, maintainability, and extensibility. Especially if we are publishing this to be hacked upon and improved by others!

timewarp_gl/shaders/timewarp_shader.hpp
Zee2: Yep, this was for debugging. I may just remove this entire shader.

Zee2: Ah, that is for the alternative eye format. We'll remove this when that issue is closed (in another PR)

timewarp_gl/shaders/timewarp_shader.hpp
Zee2: Done

timewarp_gl/plugin.cpp
Zee2: Sounds good.

common/math_util.hpp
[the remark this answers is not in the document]
Zee2: Thanks for the help! Done

timewarp_gl/plugin.cpp
mhuzai: Thanks for the explanation!

timewarp_gl/plugin.cpp
mhuzai: I guess what is confusing me is that in the commit diff, the old code has a different initialization structure. It is exactly the transpose of the new structure. So I just assumed that it was because of Eigen's ordering. But if that is not the case, then why is the order different now? Was it column-major before?

Zee2: The old ksAlgebra matrices were initialized in **storage order,** column-major. Eigen matrices are initialized storage-order-independent, in "true order". In addition, the old ksAlgebra matrices were accessed `[i][j]`, where i = column and j = row. Eigen matrices are accessed oppositely.

mhuzai (approved):

charmoniumQ (approved):

PR #129, opened 2020-09-19
New OBJ renderer + polished assets + IPD rendering

Zee2:
Screenshot from 2020-09-19 02-57-36
Screenshot from 2020-09-19 02-59-02

While the island was cute, it's time for an overhaul of our rendering backend for `gldemo`.

Why?

- We need smaller, closer-up features in our demo scene so that we can more appropriately characterize occlusion artifacts of advanced reprojection techniques
- We would like flexibility to render whatever scene we would like, instead of hardcoded geometry data
- Having enormous amounts of hardcoded geometry in a header file is... extremely gross, and embarrassing
- Looks freaking awesome :rocket: 

How?

- Two single-file header-only libraries have been added to `/common/gl_util/lib`. `stb_image.h` and `tiny_obj_loader.h` together allow us to load images and OBJ files with ease and speed.
- `/common/obj.hpp` is a scene-loader that handles materials, textures, multiple objects, etc
- `ILLIXR_DEMO_DATA` defines the directory in which the code will search for a `scene.obj` and a `headset.obj`

In addition, `gldemo` did not previously render with proper IPD. Now it renders the two eye's viewpoints with an assumed 64mm of IPD.

NOTE: Please merge #123 first! This includes those changes, so please merge that first and I'll rebase this branch.


common/gl_util/obj.hpp
```diff
@@ -0,0 +1,246 @@
...
+		ObjScene(std::string obj_filename) {
+
+			// If any of the following procedures fail to correctly load,
+			// we'll set this flag false (for the relevant operation)
+			successfully_loaded_model = true;
+			successfully_loaded_texture = true;
+
+			std::string warn, err;
+
+			std::string obj_file = std::getenv("ILLIXR_OBJ_DATA") + std::string("/") + obj_filename;
```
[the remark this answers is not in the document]
Zee2: @charmoniumQ I was hoping there was a better way we could do this. Is there any way we can pass in this value from the YAML? The env var is a sort of kludgy way of doing this imo. If we can define OBJ locations in YAML, we can have "high", "medium", and "low" scene complexity profiles, etc. Or perhaps there is a better way?

common/gl_util/obj.hpp
```diff
@@ -0,0 +1,246 @@
...
+						newObject.num_triangles = buffer.size() / 3;
+					}
+
+					objects.push_back(newObject);
+				}
+			}
+			
+		}
+
+		~ObjScene() {
```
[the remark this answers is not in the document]
Zee2: OK, thanks for catching it.

common/gl_util/obj.hpp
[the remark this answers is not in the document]
Zee2: OK, from discussion with Sam we're still using env vars. Will apply the suggested fix.

debugview/plugin.cpp
```diff
@@ -507,50 +434,9 @@ class debugview : public threadloop {
...
-		headsetObject.init(vertexPosAttr,
-			vertexNormalAttr,
-			colorUniform,
-			headset_NUM_TRIANGLES,
-			&(headset_vertex_data[0]),
-			&(headset_normal_data[0]),
-			GL_DYNAMIC_DRAW
-		);
+		// Load/initialize the demo scene.
+		demoscene = new ObjScene("demo2.obj");
```
mhuzai, line 438: Minor: why not just `demo.obj`? Does that exist already? If you do decide to change this, remember to change it in `gldemo/plugin.cpp` as well.

gldemo/plugin.cpp
```diff
@@ -190,97 +130,80 @@ class gldemo : public threadloop {
...
-			glFramebufferTextureLayer(GL_DRAW_FRAMEBUFFER, GL_COLOR_ATTACHMENT0, eyeTextures[buffer_to_use], 0, 1);
-			glBindTexture(GL_TEXTURE_2D_ARRAY, 0);
-			glClearColor(0.6f, 0.8f, 0.9f, 1.0f);
-			glClear(GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT);
+				Eigen::Matrix4f modelViewMatrix = modelMatrix * view_matrix;
+				glUniformMatrix4fv(modelViewAttr, 1, GL_FALSE, (GLfloat*)(modelViewMatrix.data()));
+				glUniformMatrix4fv(projectionAttr, 1, GL_FALSE, (GLfloat*)(basicProjection.data()));
 
-			draw_scene();
+#ifdef USE_ALT_EYE_FORMAT
```
mhuzai, line 184: Remove because of #127.

debugview/shaders/demo_shader.hpp
```diff
@@ -0,0 +1,29 @@
...
+static const char* const demo_fragment_shader =
+    "#version " GLSL_VERSION "\n"
+    "precision mediump float;\n"
+    "uniform highp sampler2D main_tex;\n"
+    "in mediump vec2 uv;\n"
+    "out lowp vec4 outcolor;\n"
+    "void main() {\n"
+    // "    float dist = gl_FragCoord.z/gl_FragCoord.w;\n"
+    // "    outcolor = u_color + vec4(1.0, 1.0, 1.0,1.0) * clamp(dist* 0.03, 0., 1.) * 0.7f;"
+    // "    outcolor = outcolor * (0.9 + clamp(dot(norm, vec3(1., 0.7, 0.3)), 0., 1.) * 0.3);"
```
mhuzai, line 27: Remove if not needed.

gldemo/shaders/demo_shader.hpp
```diff
@@ -0,0 +1,26 @@
...
+
+static const char* const demo_fragment_shader =
+    "#version " GLSL_VERSION "\n"
+    "precision mediump float;\n"
+    "uniform highp sampler2D main_tex;\n"
+    "in mediump vec2 uv;\n"
+    "out lowp vec4 outcolor;\n"
+    "void main() {\n"
+    "       outcolor = texture(main_tex, uv);\n"
+    "}\n";
```
mhuzai, line 26: This is identical to debugview's shader. Perhaps better to have a common shader under `common/`. Of course, we should not do that if we expect the shaders to diverge in the future. Do we anticipate that that would happen?

mhuzai (commented):
I think we need to improve a couple of things before merging this in:
1. Move the assets (`demo_data`) from `gldemo/` to `common/` because they are used by `debugview` as well.
2. Update the runner so that it sets `ILLIXR_OBJ_DATA` instead of the user having to set it.

debugview/plugin.cpp
charmoniumQ, line 438: Let's make these values, not pointers, unless there is a compelling reason. Try removing the `new` and the `*` from 368. Lmk if this doesn't work for some reason.

gldemo/plugin.cpp
```diff
@@ -484,45 +402,10 @@ class gldemo : public threadloop {
...
-		rocksObject.init(vertexPosAttr,
-			vertexNormalAttr,
-			colorUniform,
-			Rocks_plane002_NUM_TRIANGLES,
-			&(Rocks_Plane002_vertex_data[0]),
-			&(Rocks_Plane002_normal_data[0]),
-			GL_STATIC_DRAW
-		);
+		// Load/initialize the demo scene.
+		demoscene = new ObjScene("demo2.obj");
```
charmoniumQ, line 408: Same here with the pointer.

gldemo/shaders/demo_shader.hpp
Zee2: Yes, shaders are intentionally not shared. This is because the debugview and gldemo are only visually identical by "coincidence" right now; that is not necessary and not semantically significant. In the future we may have drastically different things rendering in gldemo and the debugview.

gldemo/plugin.cpp
Zee2: We will merge #127 before this and I'll rebase/merge up this branch.

debugview/plugin.cpp
Zee2: Woops, yes, should be just demo.obj; I used to have multiple for testing/evaluation.

gldemo/plugin.cpp
Zee2: Done.

debugview/plugin.cpp
Zee2: Fixed, also renamed to `scene.obj`.

debugview/shaders/demo_shader.hpp
Zee2: Done.

debugview/plugin.cpp
Zee2: This doesn't work, because `ObjScene` does not have a default constructor, and C++ requires all members to have a default constructor... :/

`ObjScene` is constructed with the filename associated with the .obj file. This is obtained through reading envvars, which... can't be safely done with error checking in an initializer list.

gldemo/plugin.cpp
Zee2: This doesn't work, because `ObjScene` does not have a default constructor, and C++ requires all members to have a default constructor... :/

`ObjScene` is constructed with the filename associated with the .obj file. This is obtained through reading envvars, which... can't be safely done with error checking in an initializer list.

configs/native.yaml
```diff
@@ -11,6 +11,7 @@ plugins:
   # - path: zed/
 #data: zip+http://robotics.ethz.ch/~asl-datasets/ijrr_euroc_mav_dataset/vicon_room1/V1_02_medium/V1_02_medium.zip?path=mav0
 data: ../../ILLIXR/data1/
+demo_data: ../../ILLIXR/demo_data/
```
mhuzai, line 14: Please add to `native-ground-truth.yaml` as well.

runner/runner/main.py
```diff
@@ -268,6 +270,7 @@ async def load_monado(config: Dict[str, Any]) -> None:
             ILLIXR_PATH=str(runtime_path / "runtime" / f"plugin.{profile}.so"),
             ILLIXR_COMP=":".join(map(str, plugin_paths)),
             ILLIXR_DATA=config["data"],
+            ILLIXR_DEMO_DATA=config["demo_data"],
```
mhuzai, line 273: Remove. `monado.yaml` does not have demo_data because we don't run gldemo and debugview with Monado.

configs/native.yaml
Zee2: Done

runner/runner/main.py
Zee2: I mean... It would be really nice if we can run debugview with monado in the future.

common/gl_util/obj.hpp
```diff
@@ -0,0 +1,254 @@
+#pragma once
```
mhuzai, line 1: Sorry, forgot to add this in the first review.

If I pass an invalid path for the obj data in `opt` mode, nothing gets rendered (rightfully so), but there aren't any warnings or errors to indicate why. All the error checking in this file that results in a **catastrophic** failure should result in a print. That is, those prints should always happen and not be guarded by NDEBUG. This is the scheme that we use elsewhere in ILLIXR as well.

Zee2: Done.

gldemo/plugin.cpp
Zee2: OK, I made a default constructor. All done.

debugview/plugin.cpp
Zee2: OK, I made a default constructor. All done.

charmoniumQ (approved):

common/gl_util/obj.hpp
```diff
@@ -0,0 +1,259 @@
...
+				std::cout << "[OBJ WARN] " << warn << std::endl;
+#endif
+			}
+			if(!err.empty()){
+				std::cout << "[OBJ ERROR] " << err << std::endl;
+				successfully_loaded_model = false;
+			}
+			if(!success){
+				std::cout << "[OBJ FATAL] Loading of " << obj_filename << " failed." << std::endl;
+				successfully_loaded_model = false;
```
mhuzai, line 92: I didn't notice this before but these should be fatal. There isn't much point in continuing with the program if there is no app (gldemo) running. I think an `abort()` in both scenarios should do.

Rest looks good.

Zee2: Hm. Okay

mhuzai (approved):

charmoniumQ (approved):

PR #130, opened 2020-09-20
Timewarp doesn't sync with the application correctly

mhuzai:
Timewarp wakes up too late and misses vsync. It also publishes the new swap time fairly late.


charmoniumQ (approved):
Moving `lastSwapTime` to be closer to `glXSwapBuffers` seems right to me.

Zee2 (approved):
Looks good to me.

PR #133, opened 2020-09-21
Create a CONTRIBUTING.md with info on how to create PRs

mhuzai:
Reviewers: this PR currently only addresses the protocol for PRs. If there are any other things related to contributing that we wish to add, we can discuss them here.


Zee2:
I have a few thoughts about the "issue -> PR" conversion pattern, and @mhuzai suggested I post them here to spark a discussion. This pattern is quite different from what I've seen in previous (very large) projects at big companies, where external issues were raised by the community and an internal team dealt with them.

At Microsoft (MRTK), we have a similar situation to ILLIXR, where the project is mostly maintained/improved/fixed by a core team of developers that spend most of their time on the project. The project is also consumed/used by outside community members, as well as other teams inside Microsoft.

We use GitHub Issues to represent actual issues or bugs with the project, as well as future feature goals. However, when one of the team members creates a PR to fix one of the issues (or implement the feature that the issue represents), we keep the issue ticket separate.

We do this for a few reasons; one, there is a separation between the "problem with the code" and the "change that fixes the problem." This is not always a one-to-one mapping; one PR can fix several issues, especially if the issues are related. In addition, the discussions are not always the same between an issue and a PR; the issue discussion may mostly be consumed by talk of how to reproduce the issue, the nature of the issue, the user's environment/setup while experiencing the issue, etc. Whereas for the PR, the discussion will be mostly related to the code itself; code reviews, technical problems, design decisions, etc.

Another factor is that the issues are _typically_ (but not always) raised by community members who are using the software and are reporting the issues; however, the PRs are _typically_ (again, not always) created by the dedicated team members. By maintaining a semantic separation between the reported problem and the PR to fix it, there's also a sort of clean separation between the external audience of the issue and the internal team effort to fix it. (This doesn't mean the PR is private, in fact, all the PRs are public! It is just a semantic separation that allows community members to continue to contribute reproductions/bug info while the PR discussion is intact elsewhere.)

One last thing; as far as I can tell, the GitHub UI/backend is intended for this separate PR/issue pattern. Issues can be linked to PRs, and when the PR is merged, the issues are immediately and automatically closed. This means that a PR and the issue(s) that it resolves will never get out of sync; if an issue is resolved, the issue will be closed, with no human error factor.

Personally I've never used this new pattern of converting issues into PRs, so the above discussion is purely my own opinion. @charmoniumQ @mhuzai I'd love to hear anecdotes/stories/etc of how this conversion pattern is an improvement!

apache8080:
Currently we also use Issues to discuss new features, therefore tightly integrating issues and PRs works well for us. But we have discussed moving new feature discussions/proposals to a different RFC process. This may be something that we want to decide on.

CONTRIBUTING.md
```diff
@@ -0,0 +1,17 @@
+# Contributing Guidelines
+
+Please follow these steps when making pull requests (PRs):
+
+1. First, create an issue describing the problem that needs to be fixed. If an issue already exists, skip this step.
```
charmoniumQ, line 5:
```suggestion
If you are looking for an issue to fix, see the ["good first issue" label](https://github.com/ILLIXR/ILLIXR/issues?q=is%3Aopen+is%3Aissue+label%3A%22good+first+issue%22).
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,17 @@
...
+4. Fix the issue.
+5. Push commits up to GitHub.
+6. Get [GitHub Hub](https://github.com/github/hub).
+7. Open a PR by doing `hub pull-request -i <issue number>` from command line, where `<issue number>` is the number assigned to the issue by GitHub. You should be on the branch that you would like to open the PR with.
+
+Why are the above steps necessary?
+
+1. Assigning the issue to yourself ensures that multiple people don't work on the same thing in parallel.
+2. The branch naming scheme organizes things a bit for us, and also makes it easy to find branches.
+3. Opening a PR via Hub makes sure that the issue itself is converted into a PR. This is different than linking issues to PRs. While linking is nice, it still means that we have two separate threads of discussion, one on the issue and one on the PR. By converting the issue itself into a PR, everything resides in one place, making it easier to keep track of things.
```
charmoniumQ, line 17:
```suggestion
# Getting Help

You can get help from our development community in three ways:
3. Opening a PR via Hub makes sure that the issue itself is converted into a PR. This is different than linking issues to PRs. While linking is nice, it still means that we have two separate threads of discussion, one on the issue and one on the PR. By converting the issue itself into a PR, everything resides in one place, making it easier to keep track of things.
# Getting Help

You can get help from our development community in three ways:

1. [Main documentation site](https://illixr.github.io/ILLIXR/)
2. [API documentation site](https://illixr.github.io/ILLIXR/api/html/annotated.html)
3. [Gitter](https://gitter.im/ILLIXR/community)

```

charmoniumQ:
@Zee2 The only downside with having separate issue/PR entities is that discussion on the issue could change what the PR needs to be, so PR authors need to respond to two unsynchronized comment threads.

Zee2:
> Currently we also use Issues to discuss new features, therefore tightly integrating issues and PRs works well for us. But we have discussed moving new feature discussions/proposals to a different RFC process. This may be something that we want to decide on.

From what I can tell, even using Issues to discuss new features would still work well with having separate PRs from the issues, because there's still that notion of the separation of concern between the goal/requirement/request/etc and the implementation/code that arises from that goal.

Taking the RFC idea, the Issue could _be_ the RFC, whereas the PR is the implementation of that RFC, and the discussion on that PR page is exclusively dedicated to the technical discussion of the implementation (i.e. code review, etc)

mhuzai:
> 
> 
> > Currently we also use Issues to discuss new features, therefore tightly integrating issues and PRs works well for us. But we have discussed moving new feature discussions/proposals to a different RFC process. This may be something that we want to decide on.
> 
> From what I can tell, even using Issues to discuss new features would still work well with having separate PRs from the issues, because there's still that notion of the separation of concern between the goal/requirement/request/etc and the implementation/code that arises from that goal.
> 
> Taking the RFC idea, the Issue could _be_ the RFC, whereas the PR is the implementation of that RFC, and the discussion on that PR page is exclusively dedicated to the technical discussion of the implementation (i.e. code review, etc)

I think I'm with Finn on this. For issues that are opened as bugs, the PR is the fix and can be linked to the issue. For issues that are new features, we limit the discussion regarding the feature to the issue, limit the discussion regarding the code to the PR, and link the two.

I do see value in our current approach of converting issues into PRs, but that only really works for issues that are not bugs and issues that are opened by the internal team.

mhuzai:
Bump. We need to resolve this soon, especially before the workshop.

charmoniumQ:
I see Finn's point, separate comments on the problem vs the solution. Maybe for tickets whose impact is not externally visible (then _all_ comments are about the solution), we could use just a PR, but the common case would be issue/PR.

Do you accept the changes I suggested?

charmoniumQ (approved):

PR #127, opened 2020-09-21
Remove legacy frame format (rename "rendered_frame_alt")

Zee2:
Back in "ye olde tymes" ILLIXR used a unique frame layout that used layered textures to store left/right eyebuffers. Monado used a different style of eyebuffers, so we added a `USE_ALT_EYE_FORMAT` ifdef that allowed us to switch between these formats.

We haven't used these older-style ILLIXR-ish framebuffer layouts in almost a year now, and the code that renders to them is half-broken, too. IIn addition, the old-style framebuffer types do not even have the advanced timing/pose/profiling data that is necessary for our logging and timewarp to function properly! It is probably in our best interests to abandon the old-style eyebuffer formats and exclusively use Monado-style eyebuffers instead.

In addition, calling our main/primary framebuffer layout "alternative" is confusing.

This will need a wide-reaching fix, as it will modify `/common` and require some refactors in several components.


charmoniumQ (approved):

mhuzai (commented):
This PR breaks monado_integration because monado_integration still uses `rendered_frame_alt`. Please make the appropriate changes in monado_integration and open a PR there as well, so these changes can be tested together.

Zee2:
> This PR breaks monado_integration because monado_integration still uses `rendered_frame_alt`. Please make the appropriate changes in monado_integration and open a PR there as well, so these changes can be tested together.

Ah, yep, forgot about this; will do.

Zee2:
https://github.com/ILLIXR/monado_integration/pull/5 Please approve!

mhuzai (approved):

Issue #132, opened 2020-09-20
Make the default resolution 2K


mhuzai:
Right now the default is a small VGA-sized window which has to be manually updated to 2K for any profiling experiments. Furthermore, the runtime (extended_window), timewarp, and hologram each independently define their own window size. The window size needs to be unified (initially in a header file, eventually in YAML) and needs to be set to 2K by default.

DhairyaBahl:
If this issue is still there then kindly assign me to it.

charmoniumQ:
Thanks for showing interest @DhairyaBahl :smile: . Please read our [contributing guidelines](https://github.com/ILLIXR/ILLIXR/blob/cb1888a54c4cf31647a3ee4b531a38945ff21d39/CONTRIBUTING.md) (you're at step 3).

Zee2:
This is yet another thing that should be put in a config somewhere and distributed through Switchboard/Phonebook... Quite a few different moving parts all need to agree on various image/texture dimensions. This also includes things like the camera dimensions; `debugview` displays camera frames coming from the camera component; it would be great if the cameras, eyebuffers, etc etc can all be configured through YAML/JSON and shared/published between all the components. @charmoniumQ this is another instance of dense/complex shared config like I was talking about

DhairyaBahl:
@mhuzai I would like to work on this issue but i was not in touch with this issue from a long time, so i have to start from beginning if you don't mind. 

Thanks :)

mhuzai:
That's alright but we do need this soon.

DhairyaBahl:
@mhuzai I am on it. Will do my best to complete this asap.

Thanks :)

DhairyaBahl:
@mhuzai Kindly tell me if just changing the parameters of 

        xlib_gl_extended_window(int _width, int _height, GLXContext _shared_gl_context){
            width = _width;
            height = _height;

int the extended_window.hpp and also doing the same with parameters in gldemo timewarp and plugin.cpp files will solve this issue. 

Kindly guide Thanks :)

mhuzai:
No, it will not. Please look at our [previous conversation](https://github.com/ILLIXR/ILLIXR/pull/139#discussion_r494742101). Ask if something is unclear.

DhairyaBahl:
@mhuzai Okay . No issues :)

Issue #128, opened 2020-09-19
HMD + user biometric information needs to be centralized


Zee2:
Timewarp and `gldemo` currently just assume a default IPD. There is `HMD::GetDefaultHmdInfo` and `HMD::GetDefaultBodyInfo` but these are private to Timewarp. We should be exposing this data as a config file/options somewhere and publishing to Phonebook/Switchboard/something.

This does not need to talk to Monado/OpenXR because Monado will be handling the frontend/warp in the future.

charmoniumQ:
Let's read this through env-vars, which are set by the runner, which are determined by the YAML. See `ILLIXR_DATA` for an example of this workflow. I can help you with the runner/YAML side of it.

Zee2:
@charmoniumQ The body/HMD info is quite dense, and there are upwards of 20-30 different values that need to be set. Can this be stored in some structured file/config data or something? Passing these values directly through env-vars seems sketchy.

charmoniumQ:
We could pass an env-var that points to a data-file (in some format that is easily parseable).

mhuzai:
Rishi and I were discussing this yesterday. A lot of components require their own configurations and many of these configurations are as dense as HMD info; e.g., camera calibration for SLAM. So we need per-plugin configuration files that can be specified in the YAML, and then input parsers for each plugin that read the files and set the parameters during setup.

charmoniumQ:
Ok, I agree with @mhuzai. However, that sounds like a different feature/PR.

charmoniumQ:
Let's distinguish between configuration which needs to be shared between components (such as resolution) and component-private configuration (such as delta in Timewarp's "vsync - delta"). Obviously, shared stuff needs to go in the project-level config-file. As for project-specific stuff, we already use [`yaml-include`](https://github.com/tanbro/pyyaml-include), so one can have `config: !include plugin/config.yaml`. This means we don't have to change the parsing, and there is still One Config Document, distributed over multiple files.

Let's also distinguish between runtime and compile-time configuration. Compile-time configuration could result in a faster load-time, but runtime configuration means you don't have to recompile to change something. I think that constants which do not affect the semantics of the code should be runtime, passed with env-vars, and set by the loader. Constants that affect the semantics of the code (e.g. `GLSL_VERSION`) should be passed by `${CXX} -D` and set by the loader or by the `Makefile`.

charmoniumQ:
[...]

PR #139, opened 2020-09-23
Issue-132-make-the-default-resoultion-2k

DhairyaBahl:
I have changed the default resolution in the common/extended_window and changed the default value in the parameterized constructor to be set to 2048 x 1080 i.e. the default 2K size and i also changed the default value of eye-texture-width and eye-texture-height from 1024 each to the default 2K size. Furthermore i did the same for timewarp as well.Kindly guide me further if i have to resolve it further. 

Thanks.

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -6,6 +6,8 @@ on:
 
 jobs:
   illixr-tests-pr:
+    env:
+      ghcr_token: ${{ secrets.GHCR_TOKEN }}
     runs-on: ubuntu-18.04
     steps:
       # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
@@ -14,8 +16,9 @@ jobs:
         with:
           submodules: recursive
 
-      - name: Login to ghcr.io
-        run: echo ${{ secrets.GHCR_TOKEN }} | docker login ghcr.io -u illixr-ci --password-stdin
+      - if: ${{ env.ghcr_token }}
+        name: Login to ghcr.io
+        run: echo ${{ env.ghcr_token }} | docker login ghcr.io -u illixr-ci --password-stdin
 
       - name: Extract Branch Name
         shell: bash
@@ -31,7 +34,8 @@ jobs:
       - name: Build Docker Image
         run: docker build --cache-from ghcr.io/illixr/illixr-tests:${{ steps.extract_branch.outputs.branch }},ghcr.io/illixr/illixr-tests:latest . -t ghcr.io/illixr/illixr-tests:${{ steps.extract_branch.outputs.branch }}
 
-      - name: Push Docker Image
+      - if: ${{ env.ghcr_token }}
+        name: Push Docker Image
         run: docker push ghcr.io/illixr/illixr-tests:${{ steps.extract_branch.outputs.branch }}
 
       # For builds on master we also want to push the image as the latest tag.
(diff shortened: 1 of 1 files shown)


Zee2:
Hi @DhairyaBahl, I see that you emailed me about this; thanks so much for reaching and thanks for contributing. I'll respond to your email soon! In addition, @charmoniumQ can we have an update on what the best path forward is for sharing complex/dense config info between components? Any way we can start using YAML/JSON configs and exposing them as key/value pairs on Phonebook?

@DhairyaBahl as you can tell these things are still under discussion :)

Zee2:
@DhairyaBahl Note #128, we're discussing strategies for config information like this.

DhairyaBahl:
Okay @Zee2 I am Trying to understand how this whole project works. Although I am not that much experienced but trying my best to understand the code. I wanted to ask if I have some doubts about the code or some issue then where should I ask ?

Thanks for cooperation :)

Zee2:
@DhairyaBahl I'll respond to your email in a hot second, which will answer several of your questions, but one good place to ask questions is the Gitter. You can find the link in the "Chat on Gitter" badge on the main README.md.

DhairyaBahl:
@Zee2 Okay.. got it... Thanks :)

common/extended_window.hpp
```diff
@@ -20,8 +20,8 @@ namespace ILLIXR{
         Window                  win;
         GLXContext              glc;
         xlib_gl_extended_window(int _width, int _height, GLXContext _shared_gl_context){
-            width = _width;
-            height = _height;
+            width = 2048;
```
mhuzai, line 23: You should not be overriding these variables like this. Instead, the function parameters should be set where `xlib_gl_extended_window` is called from, in `runtime_impl`.

mhuzai (changes requested):
2K resolution is 2560x1440.

Please look at the original description of #132. The resolution should be defined only in one place in a header file under `common/`, and then in all these code locations the variable/define from the header file should be read.

common/extended_window.hpp
DhairyaBahl: @mhuzai Sorry for the mistakes. Now I've changed the wrong values in gldemo/plugin.cpp from 1024 each to the exact 2K values which you've provided. I also changed the values which were being called in runtime_impl.hpp from 896x640 to exact 2k values and also corrected the mistakes in timewarp_gl/plugin.cpp

Thanks

mhuzai: Thanks for fixing the constants but as mentioned in #132, we would like to define the resolution only in one place: in a header file in the  `common/` directory. Perhaps `data_format`, or maybe a new file for just constants. Then, all the places that you made changes in would simply include that header file and get the resolution values. This would ensure that changing the resolution changes it for all the components.

DhairyaBahl: I think the only file in the common/ is extended_window.hpp in which there are things like width and height where i can make changes . I have checked the whole  data_format.hpp but i don't think there is anything to modify that will change the resolution values.

Either I have to create a new file which i will have to include in each and every part of the code or the second option that i am understanding is to make changes in extended_window.hpp.

Kindly help and guide if i am going in correct direction.

Thanks :)

mhuzai: Yes, there are currently no defines for resolution in `common/`. But that is what we have to add. New defines for height and width, set to 2k, that can be used in the other components.

DhairyaBahl: @mhuzai Okay I got it, I have created a new file common/default_resolution.hpp in which I just initialised and declared width and height variable. Now i guess including it in the files like extended_window , timewarp, runtime will do the task.

Thanks

apache8080:
@DhairyaBahl Please merge with master to get latest CI/CD changes

DhairyaBahl:
@apache8080 Okay Got it thanks for the information :)

mhuzai:
What is the status of this? Are you still working on this PR, @DhairyaBahl?

DhairyaBahl:
@mhuzai Actually, I was going through my college exams from past 6 days just after completing them I will update and submit the PR. 

Sorry For inconvenience & Thanks for cooperation :)

mhuzai:
> 
> 
> @mhuzai Actually, I was going through my college exams from past 6 days just after completing them I will update and submit the PR.
> 
> Sorry For inconvenience & Thanks for cooperation :)

Please let us know if you are able to work on this, @DhairyaBahl.

PR #140, opened 2020-09-23
OV Integrator (Fix MTP)

JeffreyZh4ng:


imu_integrator/plugin.cpp
```diff
@@ -0,0 +1,477 @@
+#include <chrono>
+#include <iomanip>
+#include <thread>
+#include <eigen3/Eigen/Dense>
+
+#include "../open_vins/ov_msckf/src/state/Propagator.h"
+#include "../open_vins/ov_core/src/types/IMU.h"
```
charmoniumQ, line 7: I don't like components depending on other components instead of `common/`. What if we switch to OrbSLAM? I understand we will have to change a lot of code in OrbSLAM, but it would be a better starting if `imu_integrator` doesn't hard-depend on OpenVINS. However, I understand this might be necessary here.

Is it possible to make a struct in `common/` that gets included in OpenVINS and in `imu_integrator` that does this job?

imu_integrator/plugin.cpp
```diff
@@ -0,0 +1,477 @@
...
+		_m_imu_raw->put(new imu_raw_type{
+			w_hat,
+			a_hat,
+			w_hat2,
+			a_hat2,
+			state_plus,
+		});
+    }
+
+	std::vector<ov_msckf::Propagator::IMUDATA> select_imu_readings(const std::vector<ov_msckf::Propagator::IMUDATA>& imu_data, double time0, double time1) {
```
charmoniumQ, line 178: If this came from OpenVINS, drop a comment saying so (and a pointer to the original source).

imu_integrator/plugin.cpp
Zee2: I agree, having a generic integrator component depend on some other component/library's code seems hard to maintain and definitely tangles things

imu_integrator/plugin.cpp
```diff
@@ -0,0 +1,479 @@
...
+		, _m_imu_integrator_input{sb->subscribe_latest<imu_integrator_input>("imu_integrator_input")}
+		, _m_imu_raw{sb->publish<imu_raw_type>("imu_raw")}
+		, _seq_expect(1)
+	{}
+
+	virtual skip_option _p_should_skip() override {
+		auto in = _m_in->get_latest_ro();
+		if (!in || in->seq == _seq_expect-1) {
+			// No new data, sleep to keep CPU utilization low
+			std::this_thread::sleep_for(std::chrono::milliseconds{1});
```
Zee2, line 32: What exactly does this sleep? Which thread?  We probably shouldn't be sleeping the core runtime loop for an entire 1ms...? If this is just sleeping this components' loop, then that's fine.

imu_integrator/plugin.cpp
```diff
@@ -0,0 +1,479 @@
...
+	int counter = 0;
+	int cam_count = 0;
+	int total_imu = 0;
+	double last_cam_time = 0;
+
+	// Open_VINS cleans IMU values older than 20 seconds, we clean values older than 5 seconds
+	void clean_imu_vec(double timestamp) {
+		auto it0 = _imu_vec.begin();
+        while (it0 != _imu_vec.end()) {
+            if (timestamp-(*it0).timestamp > 5) {
```
Zee2, line 84: Probably best that this can be configurable at some point. Or, at least use a `IMU_SAMPLE_LIFETIME` macro or something similar.

zed/src/main.cpp
```diff
@@ -199,6 +204,7 @@ class zed_imu_thread : public threadloop {
     ullong imu_time;
 
     std::size_t last_serial_no {0};
+    long long _imu_integrator_seq{0};
```
Zee2, line 207: Are we okay with using `long long` instead of `int64_t` or something along those lines?

Zee2 (commented):
These are mostly nits.

JeffreyZh4ng:
Latest commit will not pass CI/CD, has dependencies on https://github.com/ILLIXR/ILLIXR/pull/148. In order for this to compile correctly we need the data_format, slam2.cpp, and pose_prediction from that branch.

imu_integrator/Makefile
```diff
@@ -0,0 +1,3 @@
+LDFLAGS = $(shell pkg-config eigen3 --libs)
+CPPFLAGS = $(shell pkg-config eigen3 --cflags) -I../open_vins/ov_core/src/ -I../open_vins/ov_msckf/src/
```
mhuzai, line 2: Remove OV deps.

imu_integrator/common
```diff
@@ -0,0 +1 @@
+../common/
```
mhuzai, line 1: Remove because runner adds it.

imu_integrator/plugin.cpp
```diff
@@ -0,0 +1,424 @@
...
+	double last_imu_offset;
+	bool has_last_offset = false;
+	int64_t _seq_expect, _stat_processed, _stat_missed;
+
+	[[maybe_unused]] int counter = 0;
+	[[maybe_unused]] int cam_count = 0;
+	[[maybe_unused]] int total_imu = 0;
+	[[maybe_unused]] double last_cam_time = 0;
+
+	// Open_VINS cleans IMU values older than 20 seconds, we clean values older than 5 seconds
```
mhuzai, line 88: Update comment. No need to mention OV.

imu_integrator/plugin.cpp
```diff
@@ -0,0 +1,424 @@
...
+		// Loop through all IMU messages, and use them to move the state forward in time
+		// This uses the zero'th order quat, and then constant acceleration discrete
+		if (prop_data.size() > 1) {
+			for(size_t i=0; i<prop_data.size()-1; i++) {
+
+				// Time elapsed over interval
+				double dt = prop_data.at(i+1).timestamp-prop_data.at(i).timestamp;
+
+				// Corrected imu measurements
+				w_hat = prop_data.at(i).wm - imu_value.block(10,0, 3, 1);
```
mhuzai, line 153: Read bias once and use that variable everywhere else.

imu_integrator/plugin.cpp
```diff
@@ -0,0 +1,424 @@
...
+		if (input_values == NULL) {
+			return;
+		}
+
+		if (!has_last_offset) {
+			last_imu_offset = input_values->t_offset;
+			has_last_offset = true;
+		}
+
+		Eigen::Matrix<double, 16, 1> imu_value;
```
mhuzai, line 112: Is there a better name for this variable? This one is too misleading in this context because IMU values could also be the actual readings from the IMU.

imu_integrator/plugin.cpp
```diff
@@ -0,0 +1,424 @@
...
+		Eigen::Vector4d q_0 = quat;
+		Eigen::Vector3d p_0 = pos;
+		Eigen::Vector3d v_0 = vel;
+
+		// k1 ================
+		Eigen::Vector4d dq_0 = {0,0,0,1};
+		Eigen::Vector4d q0_dot = 0.5*Omega(w_hat)*dq_0;
+		Eigen::Vector3d p0_dot = v_0;
+		Eigen::Matrix3d R_Gto0 = quat_2_Rot(quat_multiply(dq_0,q_0));
+		Eigen::Vector3d v0_dot = R_Gto0.transpose()*a_hat-Eigen::Matrix<double,3,1>(0.0, 0.0, 9.81);
```
mhuzai, line 256: Put gravity in a variable and use it everywhere else. Plus, doesn't gravity come from SLAM?

mhuzai (commented):
Rename folder to rk4_integrator.

imu_integrator/plugin.cpp
JeffreyZh4ng: In frustration of finding the flipped gravity bug I just hardcoded the gravity vector in RK4. I think this is a better solution than assuming that the gravity coming from SLAM is negative and then negating it here.

mhuzai (commented):
Why is OV being changed in this PR?

imu_integrator/plugin.cpp
mhuzai: Yeah, I agree. We shouldn't assume what kind of gravity vector SLAM will publish. However, sometime in the future we should standardize the interface even more and require all SLAMs to publish gravity in a specific way.

JeffreyZh4ng:
> Why is OV being changed in this PR?

I think its only because Jae and I were trying to fix common when testing

mhuzai:
> 
> 
> > Why is OV being changed in this PR?
> 
> ~I think its only because Jae and I were trying to fix common when testing and OV accident got added into the commit~
> Actually looking at the files changes, they dont show up when I click on them and only the "move common" shows up... Because of this I cant tell if there was actually anything changed in OV. I dont think anything was changed though.

Revert it then. Looks like the common issue has been fixed, so need to remove that change anyway.

JeffreyZh4ng:
> > > Why is OV being changed in this PR?
> > 
> > 
> > ~I think its only because Jae and I were trying to fix common when testing and OV accident got added into the commit~
> > Actually looking at the files changes, they dont show up when I click on them and only the "move common" shows up... Because of this I cant tell if there was actually anything changed in OV. I dont think anything was changed though.
> 
> Revert it then. Looks like the common issue has been fixed, so need to remove that change anyway.

I just made OV point to the correct commit in the OV repo. It shows up as "changing" code in OV but in reality I only updated the ref? But the latest is the correct location of where the submodule should be pointing to

mhuzai:
> 
> 
> > > > Why is OV being changed in this PR?
> > > 
> > > 
> > > ~I think its only because Jae and I were trying to fix common when testing and OV accident got added into the commit~
> > > Actually looking at the files changes, they dont show up when I click on them and only the "move common" shows up... Because of this I cant tell if there was actually anything changed in OV. I dont think anything was changed though.
> > 
> > 
> > Revert it then. Looks like the common issue has been fixed, so need to remove that change anyway.
> 
> I just made OV point to the correct commit in the OV repo. It shows up as "changing" code in OV but in reality I only updated the ref? But the latest is the correct location of where the submodule should be pointing to

OV's latest commit is c6a4f8835851d21f87add38260e73df63c6354a1, not 74f068d49d27569f9c932e9eee1e31a729ac9eb2.

Small favor: Can you make the types on this line `int64_t`? https://github.com/ILLIXR/ILLIXR/blob/master/gtsam_integrator/plugin.cpp#L93
This PR changes both ZED and offline_imu_cam so that they publish `int64_t`. So GTSAM should use the same type.

JeffreyZh4ng:
> > > > > Why is OV being changed in this PR?
> > > > 
> > > > 
> > > > ~I think its only because Jae and I were trying to fix common when testing and OV accident got added into the commit~
> > > > Actually looking at the files changes, they dont show up when I click on them and only the "move common" shows up... Because of this I cant tell if there was actually anything changed in OV. I dont think anything was changed though.
> > > 
> > > 
> > > Revert it then. Looks like the common issue has been fixed, so need to remove that change anyway.
> > 
> > 
> > I just made OV point to the correct commit in the OV repo. It shows up as "changing" code in OV but in reality I only updated the ref? But the latest is the correct location of where the submodule should be pointing to
> 
> OV's latest commit is c6a4f8835851d21f87add38260e73df63c6354a1, not 74f068d49d27569f9c932e9eee1e31a729ac9eb2.
> 
> Small favor: Can you make the types on this line `int64_t`? https://github.com/ILLIXR/ILLIXR/blob/master/gtsam_integrator/plugin.cpp#L93
> This PR changes both ZED and offline_imu_cam so that they publish `int64_t`. So GTSAM should use the same type.

Yeah, 74f068d49d27569f9c932e9eee1e31a729ac9eb2 is the top of issue-148.2-new-imu-integrator which is the correct version of OV that this current PR should be using (74f068d49d27569f9c932e9eee1e31a729ac9eb2 is the commit has that the runner is using). Should it point to the head or OV master instead?

And yeah I'll make that change

mhuzai:
> > > > > > Why is OV being changed in this PR?
> > > > > 
> > > > > 
> > > > > ~I think its only because Jae and I were trying to fix common when testing and OV accident got added into the commit~
> > > > > Actually looking at the files changes, they dont show up when I click on them and only the "move common" shows up... Because of this I cant tell if there was actually anything changed in OV. I dont think anything was changed though.
> > > > 
> > > > 
> > > > Revert it then. Looks like the common issue has been fixed, so need to remove that change anyway.
> > > 
> > > 
> > > I just made OV point to the correct commit in the OV repo. It shows up as "changing" code in OV but in reality I only updated the ref? But the latest is the correct location of where the submodule should be pointing to
> > 
> > 
> > OV's latest commit is c6a4f8835851d21f87add38260e73df63c6354a1, not 74f068d49d27569f9c932e9eee1e31a729ac9eb2.
> > Small favor: Can you make the types on this line `int64_t`? https://github.com/ILLIXR/ILLIXR/blob/master/gtsam_integrator/plugin.cpp#L93
> > This PR changes both ZED and offline_imu_cam so that they publish `int64_t`. So GTSAM should use the same type.
> 
> Yeah, 74f068d49d27569f9c932e9eee1e31a729ac9eb2 is the top of issue-148.2-new-imu-integrator which is the correct version of OV that this current PR should be using (74f068d49d27569f9c932e9eee1e31a729ac9eb2 is the commit has that the runner is using). Should it point to the head or OV master instead?
> 
> And yeah I'll make that change

Yeah, OV should point to head. Look at the history: https://github.com/ILLIXR/open_vins/commits/c6a4f8835851d21f87add38260e73df63c6354a1

mhuzai (approved):
:tada:

PR #134, opened 2020-09-23
Fast pose is not correlated with slow pose

JeffreyZh4ng:
The difference between fast pose and slow pose is immense, both in terms of position and orientation. Not only in magnitude but also in direction -- fast pose points in completely the opposite way from slow pose.

image


JeffreyZh4ng:
For this, the fix was in Debug view. In pose prediction, we swap the orientation of the fast_pose axes and quat so that the coordinate system matches what OpenGL expects. However, this never happens in slow_pose (changing the axes before we pushed slow pose to the topic would screw up computation downstream so this fix does the axes swap after we read the topic in debug view).

mhuzai:
The quats still don't match, presumably because of the offset that is applied to fast pose. Can you apply the same offset to slow pose as well? Otherwise, it becomes very hard to determine whether there is an actual bug or not.

As for position, it is still fairly far off. Look at the X coordinate. Fast pose is 17 cm away from slow pose along the x-axis. That's 17 cm in 16 ms, or 23 mph!

image

JeffreyZh4ng:
Closed because of ttps://github.com/ILLIXR/ILLIXR/pull/163

PR #144, opened 2020-09-24
Separate debug view fast pose into predicted pose and fast pose

JeffreyZh4ng:


debugview/plugin.cpp
```diff
@@ -147,37 +148,76 @@ class debugview : public threadloop {
 			ImGui::Text("Resets to zero'd out tracking universe");
 
 			if(ImGui::Button("Zero orientation")){
-				const pose_type fast_pose = pp->get_fast_pose().pose;
+				const pose_type predicted_pose = pp->get_fast_pose().pose;
```
charmoniumQ, line 151: :+1: for changing the misleading variable name

charmoniumQ:
Is this dependent on #140 ?

JeffreyZh4ng:
> Is this dependent on #140 ?

Technically its dependent but only because I made it dependent by branching off of #134 . I guess #134 is dependent on #140 and #144 is dependent on #134

Zee2:
As with the other PR that touches the frame data, this will need a corresponding PR in `monado_integration` as well.

JeffreyZh4ng:
Closed because of https://github.com/ILLIXR/ILLIXR/pull/163

PR #146, opened 2020-09-25
Fix runner

charmoniumQ:
I thought it was also time to have a reusable plugin list `core_plugins.yaml` and `rt_slam_plugins.yaml` which get included in each other config. This way, we only have to write the git revision of OpenVINS once.

I also removed the EuRoC scripts, because we don't need them anymore.


runner/environment.yml
```diff
@@ -3,10 +3,15 @@ channels:
   - defaults
 dependencies:
   - _libgcc_mutex=0.1=main
+  - aiofiles=0.5.0=py38_0
```
charmoniumQ, line 6: I added `aiofiles` and `aiohttp`.

runner/runner/main.py
```diff
@@ -1,34 +1,32 @@
...
 from yamlinclude import YamlIncludeConstructor  # type: ignore
 
 # isort main.py
-# black main.py
+# black -l 90 main.py
 # mypy --strict main.py
 
+root_dir = relative_to((Path(__file__).parent / "../..").resolve(), Path(".").resolve())
 
-def relative_to(path: Path, root: Path):
```
charmoniumQ, line 21: Moved to `util.py`

runner/runner/main.py
```diff
@@ -45,93 +43,47 @@ def fill_defaults(thing: Any, thing_schema: Dict[str, Any], path: Optional[List[
             fill_defaults(item, thing_schema["items"], path + [str(i)])
 
 
-def pathify(path: Union[Path, str], base: Union[Path, str], should_exist=True):
```
charmoniumQ, line 48: Moved to `util.py`

runner/runner/main.py
```diff
@@ -45,93 +43,47 @@ def fill_defaults(thing: Any, thing_schema: Dict[str, Any], path: Optional[List[
...
+    path: Path = await pathify(plugin_config["path"], root_dir, cache_path, True, True, session)
     var_dict = plugin_config["config"]
     so_name = f"plugin.{profile}.so"
     targets = [so_name] + (["tests/run"] if test else [])
     await make(path, targets, var_dict)
     return path / so_name
 
 
-async def build_runtime(config: Dict[str, Any], suffix: str, test: bool = False) -> Path:
+async def build_runtime(
```
charmoniumQ, line 74: Reran `black`, which changed the formatting.

runner/runner/util.py
```diff
@@ -0,0 +1,239 @@
...
+
+    See https://en.wikipedia.org/wiki/Filename#Reserved_characters_and_words
+    """
+
+    return escape_string(
+        string, "%", {"/": "slash", ".": "dot", "~": "tilde", "\\": "backslash"}
+    )
+
+
+async def pathify(
```
charmoniumQ, line 65: This one is the magic.

mhuzai:
I'll review this in a bit, but two quick things before then:
1. CI/CD is failing.
2. Not all config files have been updated and the ones that have been don't move all the submodules; e.g., only OpenVINS' path has been changed. I understand that OV is "top priority" but we should move away from submodules in one clean stroke.

configs/native.yaml
```diff
@@ -1,16 +1,15 @@
...
   - path: pose_prediction/
   - path: gldemo/
   - path: timewarp_gl/
   - path: debugview/
   - path: audio_pipeline/
   # - path: hologram/
   # - path: zed/
-#data: zip+http://robotics.ethz.ch/~asl-datasets/ijrr_euroc_mav_dataset/vicon_room1/V1_02_medium/V1_02_medium.zip?path=mav0
-data: ../../ILLIXR/data1/
+data: file+zip+http://robotics.ethz.ch/~asl-datasets/ijrr_euroc_mav_dataset/vicon_room1/V1_02_medium/V1_02_medium.zip#path=mav0
```
apache8080, line 12: We should either update our docs explaining this url setup or have the yaml setup in a way that the user doesn't need to supply the url int this setup and can instead tell us in other ways what type it is (zip, tar, git, etc.) or we infer it by parsing the actual url.

charmoniumQ: I added this to the documentation; the `zip` scheme can only operate with zip archives (not tar or anything else). This is a current deficiency that we can fix later.

apache8080 (commented):
Was able to successfully run native.yaml but when I run monado.yaml I get the two following errors:
with .cache removed: `ValueError: demo_data points to a dir but a file expected`
with .cache: `ValueError: mav0 points to a dir but a file expected`

apache8080 (dismissed):

charmoniumQ:
All your changes look good @apache8080

runner/config_schema.yaml
```diff
@@ -1,54 +1,126 @@
...
-    default: {}
-    type: object
-    additionalProperties:
-      type: string    
+  path:
+    oneOf:
+      - type: string
+      - type: object
+        additionalProperties: false
+        required:
```
mhuzai, line 9: We should add `name` here as well. It won't affect functionality but it will help readability. The git urls are not the most readable. The worst one is Hologram being HOTLab (although I agree that that's a one off).

runner/config_schema.yaml
```diff
@@ -1,54 +1,126 @@
 $schema: http://json-schema.org/schema#
 $id: http://illixr.github.io/schemas/config.json
-type: object
 definitions:
-  key_vals: &key_vals
-    default: {}
-    type: object
-    additionalProperties:
-      type: string    
+  path:
```
mhuzai, line 4: This schema is fairly complex now. My concern is whether people will be able to easily follow it and figure out what is going on in the config files.

configs/ci.yaml
```diff
@@ -1,19 +1,19 @@
...
 loader:
   name: tests
 
-data: /does_not_exist
-demo_data: /does_not_exist
+data:
+  subpath: mav0
+  relative_to:
+    archive_path:
+      download_url: 'http://robotics.ethz.ch/~asl-datasets/ijrr_euroc_mav_dataset/vicon_room1/V1_02_medium/V1_02_medium.zip'
```
mhuzai, line 16: It would help readability to add a comment that describes how this would be parsed. The config scheme is a dense read.

configs/native-ground-truth.yaml
```diff
@@ -1,18 +1,20 @@
...
-  # - path: zed/
-#data: zip+http://robotics.ethz.ch/~asl-datasets/ijrr_euroc_mav_dataset/vicon_room1/V1_02_medium/V1_02_medium.zip?path=mav0
-data: ../../ILLIXR/data1/
-demo_data: ../../ILLIXR/demo_data/
+plugin_groups:
+  - plugin_group:
+      - path: pose_lookup
+  - !include "core_plugins.yaml"
+  - plugin_group:
+    - path: pose_lookup/
```
mhuzai, line 7: Already listed on line 4.

mhuzai (commented):
In addition to the code comments, I have the following high level comments.

1. Given that this PR doesn't remove the submodules, what happens when we merge this in and a new user clones ILLIXR? They get all the submodules but the runner still does its cache thing? Might be confusing, but not a big deal -- just confirming.

2. Didn't we decide to remove plugin groups? They make it harder to read exactly what a config will run because you have to navigate multiple files to find out. They also make it easier to have duplicate plugins and not know, and plugin ordering constraints also get harder to manage.

3. I think the configs need more explicit documentation. The config schema alone is not sufficient. At the very least, we should document plugin ordering requirements.

charmoniumQ:
> In addition to the code comments, I have the following high level comments.
> 
>     1. Given that this PR doesn't remove the submodules, what happens when we merge this in and a new user clones ILLIXR? They get all the submodules but the runner still does its cache thing? Might be confusing, but not a big deal -- just confirming.
> 
>     2. Didn't we decide to remove plugin groups? They make it harder to read exactly what a config will run because you have to navigate multiple files to find out. They also make it easier to have duplicate plugins and not know, and plugin ordering constraints also get harder to manage.
> 
>     3. I think the configs need more explicit documentation. The config schema alone is not sufficient. At the very least, we should document plugin ordering requirements.

Correct on 1.

I won't remove the groups yet, so as to minimize the number of references I have to change.

I updated the documentation.

mhuzai (approved):
Approving, but we should discuss what to do with plugin groups in the next iteration.

Minor: runner/main.py could use some more comments.

mhuzai (approved):

PR #150, opened 2020-09-30
Adding depth sharing to ILLIXR frame format

### LINKED ISSUE #152: ILLIXR-standalone depth sharing
ILLIXR needs to support depth sharing with the standalone workflow (`gldemo`, etc).

Zee2:
This adds several new pieces of information to the `rendered_frame` struct

- Depth textures
- Projection matrix, assumed now to be identical per-eye

In addition, this adds several code-hygiene and QoL improvements to `timewarp_gl`.

This resolves #152, part of effort towards #151 (epic).


gldemo/plugin.cpp
```diff
@@ -321,8 +366,11 @@ class gldemo : public threadloop {
 		createSharedEyebuffer(&(eyeTextures[0]));
 		createSharedEyebuffer(&(eyeTextures[1]));
 
+		createSharedDepthBuffer(&(eyeDepthTextures[0]));
+		createSharedDepthBuffer(&(eyeDepthTextures[1]));
+
 		// Initialize FBO and depth targets, attaching to the frame handle
-		createFBO(&(eyeTextures[0]), &eyeTextureFBO, &eyeTextureDepthTarget);
+		createFBO(&(eyeTextures[0]), &eyeTextureFBO, &(eyeDepthTextures[0]), &eyeTextureDepthTarget);
```
mhuzai, line 373: There is a gap in my understanding of OpenGL here. Why create the FBO with just the 0th textures? How does the right eye (1st index) work?

gldemo/plugin.cpp
```diff
@@ -23,6 +23,8 @@ static constexpr int   EYE_TEXTURE_HEIGHT  = 1024;
 static constexpr std::chrono::nanoseconds vsync_period {std::size_t(NANO_SEC/60)};
 static constexpr std::chrono::milliseconds VSYNC_DELAY_TIME {std::size_t{2}};
 
+static constexpr GLenum DEPTH_FORMAT = GL_DEPTH_COMPONENT32;
+
 // Monado-style eyebuffers:
 // These are two eye textures; however, each eye texture
 // represnts a swapchain. eyeTextures[0] is a swapchain of
```
mhuzai, line 30: In what sense is "swapchain" being used here?

gldemo/plugin.cpp
```diff
@@ -321,8 +366,11 @@ class gldemo : public threadloop {
 		createSharedEyebuffer(&(eyeTextures[0]));
```
mhuzai, line 366: Please update the comment above this line.

timewarp_gl/plugin.cpp
```diff
@@ -342,49 +338,58 @@ class timewarp_gl : public threadloop {
     	tw_start_transform_unif = glGetUniformLocation(timewarpShaderProgram, "TimeWarpStartTransform");
     	tw_end_transform_unif = glGetUniformLocation(timewarpShaderProgram, "TimeWarpEndTransform");
     	tw_eye_index_unif = glGetUniformLocation(timewarpShaderProgram, "ArrayLayer");
```
mhuzai, line 340: Remove, since `ArrayLayer` is no longer used in the shader.

mhuzai (commented):
Left a few comments. Please also update to latest master.

Thanks for adding all the error checking!

gldemo/plugin.cpp
```diff
@@ -185,12 +196,16 @@ class gldemo : public threadloop {
 			glFlush();
```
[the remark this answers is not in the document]
Zee2: I did a little bit of reading into this. Apparently many modern drivers completely ignore `glFlush()` entirely. We really should be using a separate mechanism to synchronize accesses to these FBOs, like so:

https://www.khronos.org/opengl/wiki/GLAPI/glCheckFramebufferStatus

` GLenum glCheckFramebufferStatus(GLenum target​);`

However, if we don't use `glFinish` at all, we'll probably start messing up our timing, because otherwise the CPU thread doesn't really know when the rendering is "done". I'm going to switch it to `glFinish()` and test. However, I'm worried that using `glFinish()` will actually stall more than just the `gldemo` thread... will require some evaluation.

astro-friedel:
This PR has been inactive for several years

PR #153, opened 2020-10-02
Tweak threadloop timing to not include `_p_skip`

charmoniumQ:


mhuzai (commented):
If you can update with master, this is good to go.

mhuzai (approved):

PR #154, opened 2020-10-02
Start threads at the same time

charmoniumQ:


mhuzai:
Everything prior to the pause functionality has been tested thoroughly and looks good. Do you wish to keep the pause functionality or was it only in case the previous changes didn't work out?

charmoniumQ:
> Everything prior to the pause functionality has been tested thoroughly and looks good. Do you wish to keep the pause functionality or was it only in case the previous changes didn't work out?

If it is up to me, I recommend including the "separated start method" but not the full pause functionality, which is kind of a hack. I can easily make this change if you agree.

mhuzai:
> > Everything prior to the pause functionality has been tested thoroughly and looks good. Do you wish to keep the pause functionality or was it only in case the previous changes didn't work out?
> 
> If it is up to me, I recommend including the "separated start method" but not the full pause functionality, which is kind of a hack. I can easily make this change if you agree.

Yup, I agree with you.

And in that case, this is good to go once you make the change and update to latest master.

common/gl_util/lib/stb_image.h
```diff
@@ -4120,7 +4120,7 @@ static int stbi__zhuffman_decode_slowpath(stbi__zbuf *a, stbi__zhuffman *z)
    if (s >= 16) return -1; // invalid code!
    // code size is s, so:
    b = (k >> (16-s)) - z->firstcode[s] + z->firstsymbol[s];
-   if (b >= (int) sizeof (z->size)) return -1; // some data was corrupt somewhere!
+   if (size_t(b) >= sizeof (z->size)) return -1; // some data was corrupt somewhere!
```
mhuzai, line 4123: Please revert.

common/plugin.hpp
```diff
@@ -22,12 +22,17 @@ namespace ILLIXR {
...
 		/**
-		 * @brief A method which Spindle calls when it starts the component.
-		 * 
+		 * @brief A method which Spindle called after the constructor.
+		 *
 		 * This is necessary for actions which have to be started after constructions, such as
 		 * threads. These cannot be started in the constructor because virtual methods don't work in
 		 * consturctors.
 		 */
+		virtual void setup() { }
```
mhuzai, line 31: Is this still required? Given that there is no pause functionality?

common/gl_util/lib/stb_image.h
charmoniumQ: I think this is better. Extending an `int` to a `size_t` is safer than `size_t` to `int`.

apache8080 (dismissed):

apache8080 (approved):

PR #155, opened 2020-10-05
Issue 122 timewarp

charmoniumQ:


timewarp_gl/plugin.cpp
```diff
@@ -539,9 +541,24 @@ class timewarp_gl : public threadloop {
...
+			{lastSwapTime},
+			{imu_to_display},
+			{predict_to_display},
+			{render_to_display},
+		}});
+
 #ifndef NDEBUG
 		auto afterSwap = glfwGetTime();
 		printf("\033[1;36m[TIMEWARP]\033[0m Swap time: %5fms\n", (float)(afterSwap - beforeSwap) * 1000);
+		printf("\033[1;36m[TIMEWARP]\033[0m Motion-to-display latency: %3f ms\n", float(mtp.count()) / 1e6);
```
mhuzai, line 559: There is a compilation error here. `mpt.count()` should be `imu_to_display.count()`.

Also, we print both MTP and PTD. Might as well print RTD too.

mhuzai (commented):
If you can fix the compilation error and update with master, this is good to go.

Btw, moving the logging to before the GL querying saves a little bit of the CPU time you spend spinning on the query. Nice!

mhuzai (changes requested):
Please print this as well: `printf("\033[1;36m[TIMEWARP]\033[0m Render-to-display latency: %3f ms\n", float(render_to_display.count()) / 1e6);`

mhuzai (approved):

apache8080 (approved):

PR #138, opened 2020-10-10
QEMU image for ILLIXR

jprx:
To have a more portable dev environment we need a QEMU image that will allow a developer to emulate ILLIXR on any host machine.

#### Requirements
- [ ] A QEMU image that a user can download and successfully run all non-CUDA components including Monado.
- [x] Script to build QEMU image
- [ ] Build QEMU image on CI/CD for master and store image in Github Package Registry

#### Notes/Brainstorming
In order to enable to good graphics rendering we should use [Virglrenderer](https://www.collabora.com/news-and-blog/blog/2019/08/28/virglrenderer-state-of-virtualized-virtual-worlds/). We also may want to look into [this](https://medium.com/@calerogers/gpu-virtualization-with-kvm-qemu-63ca98a6a172) as a way of running CUDA code in the QEMU image on machines that do have NVIDIA gpus.


jprx:
Build qemu with `scripts/install_qemu.sh`, setup and launch a VM using `qemu/run.sh`. All documentation is provided in `qemu/INSTRUCTIONS.md`

apache8080 (commented):
This looks great. Can you also link to the QEMU Instructions on both the main README and the getting_started.md.

jprx:
Added instructions to getting_started page and also fixed a broken link on the main README

README.md
```diff
@@ -42,7 +42,7 @@ We request that you cite our following paper (new version coming soon) when you
 
 ## Getting Started and Documentation
 
-For more information, see our [getting started page](https://illixr.github.io/ILLIXR/docs/getting_started/).
+For more information, see our [getting started page](docs/getting_started.md).
```
apache8080, line 45: Change the link to this: https://illixr.github.io/ILLIXR/getting_started/

This is what is on our landing page and we want it to link to the getting started web page.

docs/getting_started.md
```diff
@@ -79,6 +79,10 @@ ILLIXR can also benchmark each component in isolation.
 
 3. Each component is a directory in `benchmark`. See those components for their documentation.
 
+## Virtual Machine
+
+ILLIXR can be run inside a Qemu-KVM image. Check out the instructions [here](../qemu/INSTRUCTIONS.md)
```
apache8080, line 84: Similar change here, this is going to be on our getting started docs webpage so this link needs to be the actual github URL to the INSTRUCTIONS.md

apache8080: Or move the qemu instructions to the docs folder and we can generate the webpage using mkdocs.

apache8080: You will also need to update the mkdocs.yml if you go this direction.

install_deps.sh
```diff
@@ -71,6 +71,10 @@ then
 		. ./scripts/install_gtest.sh
 	fi
 
+	if [ ! -d "${opt_dir}/qemu" ] && y_or_n "Next: Install qemu"; then
```
charmoniumQ, line 74:
```
-	if [ ! -d "${opt_dir}/qemu" ] && y_or_n "Next: Install qemu"; then
+	if [ ! -d "${opt_dir}/qemu" ] && y_or_n "Next: Install qemu (not necessary for core ILLIXR; necessary for virtualization)"; then
```

Let's add that, unlike the other deps, this one is really optional.

jprx:
Fixed links in README and getting_started.md, decided to go with a link to Github pages as opposed to using mkdocs. Can switch to readthedocs if wanted but personally I think it looks way better in Github.

charmoniumQ (approved):

qemu/run.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+QEMU_PATH="/opt/ILLIXR/qemu/build/x86_64-softmmu"
+
+# Download ubuntu-18.04.5-desktop-amd64.iso if it doesn't exist already
+if [ ! -f ubuntu-18.04.5-desktop-amd64.iso ]; then
+    wget https://releases.ubuntu.com/18.04/ubuntu-18.04.5-desktop-amd64.iso
+fi
+
+# Create a qcow2 image, if one doesn't exist, and launch with ubuntu installed
+if [ ! -f illixr.qcow2 ]; then
```
mhuzai, line 10: If you only check for building qcow2 in the if-else, the launch code (lines 14-20) won't have to be duplicated (lines 25-31) and can be written just once after the if-else.

qemu/INSTRUCTIONS.md
```diff
@@ -0,0 +1,53 @@
+# Setting up IlliXR in Qemu
```
mhuzai, line 1: IlliXR --> ILLIXR throughout this document.

mhuzai (commented):
This is great, @jprx. Just pointed out some minor stuff.

qemu/INSTRUCTIONS.md
jprx: Ah yes- my bad!! I thought I fixed this

qemu/run.sh
jprx: The second case is slightly different to the first in that it launches without the ubuntu ISO attached. That way we boot into the qcow and not back into the installer

jprx: This way if the user later deletes the ubuntu iso there's no issues at all, everything just works (tm)

mhuzai: Oh, interesting. I didn't know that `qemu-img create` carried over state to the launch command.

Sounds good!

jprx: It doesn't carry over state, its just that if the image hasn't been created then we know the user needs to create an image and install ubuntu. Once the image is created we assume they've successfully installed ubuntu (otherwise, they should remove the image and re-run the script per the instructions)

jprx: Its really "2 scripts in one"- an "install_ubuntu" and "run_ubuntu" just packaged into one that should just do the right thing

mhuzai: Sorry, I don't know what I'm missing here. Is the following not logically equivalent to what is in the script at the moment?

```
if [ ! -f illixr.qcow2 ]; then
    qemu-img create -f qcow2 illixr.qcow2 30G
fi

# Launch illixr.qcow2
$QEMU_PATH/qemu-system-x86_64 -enable-kvm -M q35 -smp 2 -m 4G \
        -hda illixr.qcow2 \
        -net nic,model=virtio \
        -net user,hostfwd=tcp::2222-:22 \
        -vga virtio \
        -cdrom ubuntu-18.04.5-desktop-amd64.iso \
        -display sdl,gl=on
```

jprx: Yes, the key difference is the `-cdrom` line- when we detect the existence of `illixr.qcow2` we run qemu without the cdrom inserted. That's the only difference but it's a significant one (this way we boot into the installed ubuntu and not the CD, and it also lets the user remove the installer media to save disk space).

mhuzai: But the `-cdrom` line is the same in both branches, no?

```
IF:   -cdrom ubuntu-18.04.5-desktop-amd64.iso \
ELSE: -cdrom ubuntu-18.04.5-desktop-amd64.iso \
```

jprx: Good catch- that was a typo on my part. I did some looking into it though and there was a bigger issue with the behavior after removing the installer media. I pushed a fix that should address both of these issues with the correct behaviour

mhuzai (commented):
After qemu finishes cloning, I get this error:
`ERROR: Cannot find Ninja`

jprx:
> After qemu finishes cloning, I get this error:
> `ERROR: Cannot find Ninja`

Nothing about qemu building should have changed, that's really weird. You're installing it from install_deps?

mhuzai (approved):

Issue #75, opened 2020-08-21
Replace `std::mutex` with `std::shared_mutex`


charmoniumQ:
A `std::shared_mutex` permits EITHER n readers OR 1 writer.

- When a thread reads a datastructure, it only needs to acquire `std::shared_lock`. This permits parallel reads, since they don't conflict, but locks out all writers.
- When a thread writes a datastructure, it needs to acquire `std::unique_lock`. This locks out all other writers and all readers.

immanuelsavio:
Can you assign this issue to my name

mhuzai:
@charmoniumQ, is this still an open issue? Are there any known places where the mutexes can be replaced?

charmoniumQ:
- `common/sqlite_record_logger.hpp` needs this (additionally the read is unsynchronized, so needs a `shared_lock` there, and the existing lock can be a `unique_lock`).
- `common/phonebook.hpp` needs this.
- However, `runtime/switchboard_impl.hpp` does _not_ need this, if we merge in #32 .

immanuelsavio:
I have made the required changes here #158

PR #158, opened 2020-10-21
all necessary occourances of std::mutex replaced with std::shared_mutex

immanuelsavio:
Fixes the issue #75

### DIFF
--- a/common/phonebook.hpp
+++ b/common/phonebook.hpp
@@ -4,7 +4,8 @@
 #include <stdexcept>
 #include <iostream>
 #include <cassert>
-#include <mutex>
+/*#include <mutex> - Replaced Mutex with Shared Mutex*/
+#include <shared_mutex>
 #include <memory>
 #include <unordered_map>
 
@@ -95,7 +96,7 @@ namespace ILLIXR {
 		 */
 		template <typename specific_service>
 		void register_impl(std::shared_ptr<specific_service> impl) {
-			const std::lock_guard<std::mutex> lock{_m_mutex};
+			const std::lock_guard<std::shared_mutex> lock{_m_mutex};
 
 			const std::type_index type_index = std::type_index(typeid(specific_service));
 #ifndef NDEBUG
@@ -116,7 +117,7 @@ namespace ILLIXR {
 		 */
 		template <typename specific_service>
 		std::shared_ptr<specific_service> lookup_impl() const {
-			const std::lock_guard<std::mutex> lock{_m_mutex};
+			const std::lock_guard<std::shared_mutex> lock{_m_mutex};
 
 			const std::type_index type_index = std::type_index(typeid(specific_service));
 
@@ -138,7 +139,7 @@ namespace ILLIXR {
 
 	private:
 		std::unordered_map<std::type_index, const std::shared_ptr<service>> _m_registry;
-		mutable std::mutex _m_mutex;
+		mutable std::shared_mutex _m_mutex;
 	};
 }
 
--- a/common/phonebook.hpp
+++ b/common/phonebook.hpp
@@ -4,7 +4,6 @@
 #include <stdexcept>
 #include <iostream>
 #include <cassert>
-/*#include <mutex> - Replaced Mutex with Shared Mutex*/
 #include <shared_mutex>
 #include <memory>
 #include <unordered_map>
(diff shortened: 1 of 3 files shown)


mhuzai:
Thank you for looking into this, @immanuelsavio. 

@charmoniumQ, does this have a dependency on #32?

runtime/switchboard_impl.hpp
```diff
@@ -4,7 +4,8 @@
 #include <vector>
 #include <iostream>
 #include <cassert>
-#include <mutex>
+/*#include <mutex> - Replaced mutex with shared_mutex*/
```
charmoniumQ, line 7:
```suggestion
```

common/phonebook.hpp
```diff
@@ -95,7 +96,7 @@ namespace ILLIXR {
 		 */
 		template <typename specific_service>
 		void register_impl(std::shared_ptr<specific_service> impl) {
-			const std::lock_guard<std::mutex> lock{_m_mutex};
+			const std::lock_guard<std::shared_mutex> lock{_m_mutex};
```
charmoniumQ, line 99: In order to actually derive benefit from the `std::shared_mutex`, we need to change `std::lock_guard` to `std::unique_lock` in places where we _write_ to protected variables (such as this) and `std::shared_lock` in places where we _read_ (such as `lookup_impl()`).

charmoniumQ (commented):
@mhuzai No dependency on #32. The intention of my comment regarding #32 is merely to minimize duplicated effort.

common/phonebook.hpp
immanuelsavio: Changed the required locks to unique and shared for writer and reader resp.

immanuelsavio (commented):
Changed and committed the shared and unique lock changes

runtime/sqlite_record_logger.hpp
```diff
@@ -180,7 +180,7 @@ class sqlite_record_logger : public record_logger {
 		if (result != registered_tables.cend()) {
 			return result->second;
 		} else {
-			const std::lock_guard lock{_m_registry_lock};
+			const std::shared_lock lock{_m_registry_lock};
```
charmoniumQ, line 183: This method needs to look like this:

```C++
	sqlite_thread& get_sqlite_thread(const record& r) {
		const record_header& rh = r.get_record_header();
		{
			const std::shared_lock{_m_registry_lock};
			auto result = registered_tables.find(rh.get_id());
			if (result != registered_tables.cend()) {
				return result->second;
			}
		}
		const std::unique_lock lock{_m_registry_lock};
		auto pair = registered_tables.try_emplace(rh.get_id(), rh);
		return pair.first->second;
	}
```

I am guarding `registered_tables.find` and its returned value with `shared_lock`, since it is a read.

I am guarding `registered_tables.try_emplace` with `unique_lock`, since that is a write.

runtime/switchboard_impl.hpp
```diff
@@ -334,7 +334,7 @@ namespace ILLIXR {
 			  - Calls topic.schedule, which acquires _m_callbacks_lock, (see its proof of thread-safety)
 			  Therefore this method is thread-safe.
 			 */
-			const std::lock_guard lock{_m_registry_lock};
+			const std::shared_lock lock{_m_registry_lock};
```
charmoniumQ, line 337: `try_emplace` is an operation that could trigger a write (if the key does not already exist. Therefore, we have to use `unique_lock` before calling `_m_registry.try_emplace`.

charmoniumQ (commented):
Almost there! Just a few more tweaks.

runtime/sqlite_record_logger.hpp
immanuelsavio: Is it similar for all the methods using shared_lock, i mean the localization of the process? 
Cause the previous code used to be this:

```cpp
sqlite_thread& get_sqlite_thread(const record& r) {
		const record_header& rh = r.get_record_header();
		auto result = registered_tables.find(rh.get_id());
		if (result != registered_tables.cend()) {
			return result->second;
		} else {
			const std::shared_lock lock{_m_registry_lock};
			auto pair = registered_tables.try_emplace(rh.get_id(), rh);
			return pair.first->second;
		}
	}
```
And the change you've asked is:
```cpp
sqlite_thread& get_sqlite_thread(const record& r) {
		const record_header& rh = r.get_record_header();
		{
			const std::shared_lock{_m_registry_lock};
			auto result = registered_tables.find(rh.get_id());
			if (result != registered_tables.cend()) {
				return result->second;
			}
		}
		const std::unique_lock lock{_m_registry_lock};
		auto pair = registered_tables.try_emplace(rh.get_id(), rh);
		return pair.first->second;
	}
```
I am assuming it is, and I'm changing the locks in that format if otherwise, I'll change it back.

charmoniumQ: The other instances of `shared_lock` I think already distinguish the reads from the writes. It's just in this place (as far as I know) that they were mixed together.

PR #159, opened 2020-10-21
Issue 88 add global shutter

Hyjale:
Realsense Sensor support for ILLIXR 

extras:
- removed cvtColor from debugview (and OpenVINS)
- ZED now retrieves monochrome images for SLAM 
- ZED retrieves RGB and depth images for reconstruction


debugview/plugin.cpp
```diff
@@ -179,8 +179,7 @@ class debugview : public threadloop {
 		}
 		if(last_datum_with_images->img0.has_value()){
 			glBindTexture(GL_TEXTURE_2D, camera_textures[0]);
-			cv::Mat img0;
-			cv::cvtColor(*last_datum_with_images->img0.value(), img0, cv::COLOR_BGR2GRAY);
```
charmoniumQ, line 183: Please make a corresponding change in `offline_imu_cam` (insert the `cvtColor` call or read the images in grayscale in the first place).

realsense/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
...
+
+# OpenCV
+find_package(OpenCV REQUIRED)
+include_directories(${OpenCV_INCLUDE_DIRS})
+
+# Intel RealSense
+find_package(realsense2 REQUIRED)
+include_directories("${realsense2_INCLUDE_DIR}")
+
+add_definitions(-Wno-format-extra-args -std=c++17 -g -O3)
```
charmoniumQ, line 11: Add `-Wall -Wextra -Werror`. The compiler will complain more, but it's usually worth it. Message me the new warnings this introduces, if any.

realsense/plugin.cpp
```diff
@@ -0,0 +1,179 @@
...
+		cfg.enable_stream(RS2_STREAM_INFRARED, 2, IMAGE_WIDTH, IMAGE_HEIGHT, RS2_FORMAT_Y8, FPS);
+		cfg.enable_stream(RS2_STREAM_COLOR, IMAGE_WIDTH, IMAGE_HEIGHT, RS2_FORMAT_BGR8, FPS);
+		cfg.enable_stream(RS2_STREAM_DEPTH, IMAGE_WIDTH, IMAGE_HEIGHT, RS2_FORMAT_Z16, FPS);
+		profiles = pipe.start(cfg, callback);
+
+		// Get depth scale which is used to convert the measurements into millimeters
+    depth_scale = pipe.get_active_profile().get_device().first<rs2::depth_sensor>().get_depth_scale();
+	}
+
+	virtual void start() override { plugin::start(); }
```
charmoniumQ, line 141: Delete `start()` and `stop()`, unless you need to do anything _in addition_ to the default action . The default action in this case is to call `plugin::start()` and `plugin::stop()`.

realsense/plugin.cpp
```diff
@@ -0,0 +1,179 @@
...
+	std::mutex mutex;
+	rs2::pipeline_profile profiles;
+	rs2::pipeline pipe;
+	rs2::config cfg;
+	float depth_scale;
+	rs2_vector gyro_data;
+	rs2_vector accel_data;
+	double ts;
+
+	Eigen::Vector3f la;
```
charmoniumQ, line 161: Declare `la` and `av` as locals on L83. Fewer instance variables => less state => fewer things can go wrong.

```C++
Eigen::Vector3f la = {accel.x, accel.y, accel.z};
```

realsense/plugin.cpp
```diff
@@ -0,0 +1,179 @@
...
+	time_type t;
+	ullong imu_time;
+
+	cam_type cam_type_;
+	accel_type accel_type_;
+	int iteration_cam = 0;
+	int iteration_accel = 0;
+	int last_iteration_cam;
+	int last_iteration_accel;
+	std::size_t last_serial_no {0};
```
charmoniumQ, line 174: If you don't need these serial numbers, delete them.

realsense/plugin.cpp
```diff
@@ -0,0 +1,179 @@
...
+					rs2::video_frame depth_frame = fs.get_depth_frame();
+					rs2::video_frame rgb_frame = fs.get_color_frame();
+					cv::Mat ir_left = cv::Mat(cv::Size(IMAGE_WIDTH, IMAGE_HEIGHT), CV_8UC1, (void*)ir_frame_left.get_data());
+					cv::Mat ir_right = cv::Mat(cv::Size(IMAGE_WIDTH, IMAGE_HEIGHT), CV_8UC1, (void*)ir_frame_right.get_data());
+					cv::Mat rgb = cv::Mat(cv::Size(IMAGE_WIDTH, IMAGE_HEIGHT), CV_8UC3, (void*)rgb_frame.get_data());
+					cv::Mat depth = cv::Mat(cv::Size(IMAGE_WIDTH, IMAGE_HEIGHT), CV_16UC1, (void*)depth_frame.get_data());
+					cv::Mat converted_depth;
+					depth.convertTo(converted_depth, CV_32FC1, depth_scale * 1000.f);
+
+					cam_type_.img0 = new cv::Mat{ir_left};
```
charmoniumQ, line 55: Declaring these all at once is slightly preferred:

```C++
cam_type_ = cam_type{
            // Make a copy, so that we don't have race
            new cv::Mat{imageL_ocv},
            new cv::Mat{imageR_ocv},
            new cv::Mat{rgb_ocv},
            new cv::Mat{depth_ocv},
            iteration_no,
        };
```

realsense/plugin.cpp
charmoniumQ: Same for the timestamps.

realsense/plugin.cpp
```diff
@@ -0,0 +1,179 @@
...
+public:
+	realsense(std::string name_, phonebook *pb_)
+	: plugin{name_, pb_}
+	, sb{pb->lookup_impl<switchboard>()}
+	, _m_imu_cam{sb->publish<imu_cam_type>("imu_cam")}
+	, _m_rgb_depth{sb->publish<rgb_depth_type>("rgb_depth")}
+	{
+		auto callback = [&](const rs2::frame& frame)
+    {
+				std::lock_guard<std::mutex> lock(mutex);
```
charmoniumQ, line 41:
```suggestion
				std::lock_guard<std::mutex> lock(mutex);
				// This lock guarantees that concurrent invocations of `callback` are serialized.
				// Even if the API does not invoke `callback` in parallel, this is still important for the memory-model.
				// Without this lock, prior invocations of `callback` are not necessarily "happens-before" ordered, so accessing persistent variables constitutes a data-race, which is undefined behavior in the C++ memory model.
```

charmoniumQ: @mhuzai Make sure I explained this properly.

realsense/plugin.cpp
```diff
@@ -0,0 +1,179 @@
...
+		cfg.enable_stream(RS2_STREAM_GYRO, RS2_FORMAT_MOTION_XYZ32F, GYRO_RATE); // adjustable set to 0, 200 (default), 400 hz
+		cfg.enable_stream(RS2_STREAM_INFRARED, 1, IMAGE_WIDTH, IMAGE_HEIGHT, RS2_FORMAT_Y8, FPS);
+		cfg.enable_stream(RS2_STREAM_INFRARED, 2, IMAGE_WIDTH, IMAGE_HEIGHT, RS2_FORMAT_Y8, FPS);
+		cfg.enable_stream(RS2_STREAM_COLOR, IMAGE_WIDTH, IMAGE_HEIGHT, RS2_FORMAT_BGR8, FPS);
+		cfg.enable_stream(RS2_STREAM_DEPTH, IMAGE_WIDTH, IMAGE_HEIGHT, RS2_FORMAT_Z16, FPS);
+		profiles = pipe.start(cfg, callback);
+
+		// Get depth scale which is used to convert the measurements into millimeters
+    depth_scale = pipe.get_active_profile().get_device().first<rs2::depth_sensor>().get_depth_scale();
+	}
```
charmoniumQ, line 139:
```suggestion
		profiles = pipe.start(cfg, [&](const rs2::frame& frame) { this->callback(frame); });

		// Get depth scale which is used to convert the measurements into millimeters
    depth_scale = pipe.get_active_profile().get_device().first<rs2::depth_sensor>().get_depth_scale();
	}
	
	
	void callback(const rs2::frame& frame) {
		// callback code from L40 to L124 goes here now.
	}
```

For the sake of readability, let's make void callback(const rs2::frame& frame) an instance-method on realsense. This is what we do in offline_imu_cam, debugview, and others because putting a lot of code in lambdas is confusing (due to capture-semantics).

realsense/plugin.cpp
```diff
@@ -0,0 +1,179 @@
...
+					if (s == "Accel") {
+						rs2::motion_frame accel = mf;
+						accel_data = accel.get_motion_data();
+						accel_type_.accel_data = &accel_data;
+						accel_type_.iteration = iteration_accel;
+						iteration_accel++;
+					}
+
+					if (s == "Gyro") {
+						if (last_iteration_accel == accel_type_.iteration) { return; }
```
charmoniumQ, line 75: Note that this code expects frames to arrive like Accel[0], Gyro[0], Accel[1], Gyro[1], ...

If the sensor sends Gyro[0], Accel[0], Gyro[1], Accel[1], ..., then Gyro[1] will pair itself with the most recent accel, i.e. Accel[0]. This frame mismatch is probably not too bad since the framerate of Accel and Gyro is _so_ high. @mhuzai @apache8080 Comment if you think this is a problem; we may be able to switch to using the global clock to see if data is new.

charmoniumQ (approved):

realsense/plugin.cpp
```diff
@@ -0,0 +1,172 @@
...
+            cv::Mat ir_left = cv::Mat(cv::Size(IMAGE_WIDTH, IMAGE_HEIGHT), CV_8UC1, (void*)ir_frame_left.get_data());
+			cv::Mat ir_right = cv::Mat(cv::Size(IMAGE_WIDTH, IMAGE_HEIGHT), CV_8UC1, (void *)ir_frame_right.get_data());
+			cv::Mat rgb = cv::Mat(cv::Size(IMAGE_WIDTH, IMAGE_HEIGHT), CV_8UC3, (void *)rgb_frame.get_data());
+			cv::Mat depth = cv::Mat(cv::Size(IMAGE_WIDTH, IMAGE_HEIGHT), CV_16UC1, (void *)depth_frame.get_data());
+			cv::Mat converted_depth;
+			float depth_scale = pipe.get_active_profile().get_device().first<rs2::depth_sensor>().get_depth_scale(); // for converting measurements into millimeters
+			depth.convertTo(converted_depth, CV_32FC1, depth_scale * 1000.f);
+			cam_type_ = cam_type{
+				new cv::Mat{ir_left},
+				new cv::Mat{ir_right},
```
mhuzai, line 70: This is incorrect. The IR images have IR dots in them, and so cannot be used for SLAM. It is just dawning on me that none of the RealSense depth cameras would have stereo RGB. I guess we'll need the T265 after all.

Don't grab IR images for the moment. This PR, for now, should just be about RGB+Depth for scene reconstruction. At the same time, we'll need the stereo and IMU code in the future when the new camera arrives. So I think the way to do this is to check the RealSense camera model. If it is depth, only publish RGB+Depth. If it is tracking, only publish stereo-RGB+IMU.

mhuzai (commented):
Nice work! I left a few comments. Also please update to latest master. It should also fix CI/CD on this branch.

realsense/plugin.cpp
mhuzai: Good catch. SLAM is super sensitive to this sort of thing, so we should maintain the correct ordering.

realsense/plugin.cpp
mhuzai: I don't completely follow. The idea is that the RealSense thread will call `callback` every time it has a frame but there is no guarantee that the previous invocation of `callback` would have finished by then?

common/data_format.hpp
```diff
@@ -11,7 +11,7 @@
 #include <GL/gl.h>
 #include <GLFW/glfw3.h>
 //#undef Complex // For 'Complex' conflict
-#include "phonebook.hpp"
+#include "phonebook.hpp"
```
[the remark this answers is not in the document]
Hyjale: Misclick.

common/data_format.hpp
```diff
@@ -35,6 +35,11 @@ namespace ILLIXR {
 		ullong dataset_time;
 	} imu_cam_type;
 
+		typedef struct {
+				std::optional<cv::Mat*> rgb;
+				std::optional<cv::Mat*> depth;
+		} rgb_depth_type;
+
```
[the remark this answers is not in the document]
Hyjale: ZED and ElasticFusion was using the previous `rgb_depth_type`. EF required rgb/depth data as well the timestamp.

realsense/plugin.cpp
```diff
@@ -0,0 +1,172 @@
...
+#include "common/switchboard.hpp"
+#include "common/data_format.hpp"
+
+using namespace ILLIXR;
+
+static constexpr int IMAGE_WIDTH = 640;
+static constexpr int IMAGE_HEIGHT = 480;
+static constexpr int FPS = 30;
+static constexpr int GYRO_RATE = 400; // 200 or 400
+static constexpr int ACCEL_RATE = 250; // 63 or 250
```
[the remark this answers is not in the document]
Hyjale: Tested with OV and responds to my movements pretty well. Haven't had a chance to try with proper calibration (sorry), but will let you know.

scripts/install_apt_deps.sh
```diff
@@ -12,5 +12,5 @@ sudo apt-get install -y \
 	 glslang-tools libsdl2-dev libglu1-mesa-dev mesa-common-dev freeglut3-dev libglew-dev glew-utils libglfw3-dev \
 	 libusb-dev libusb-1.0 libudev-dev libv4l-dev libhidapi-dev \
 	 build-essential libx11-xcb-dev libxcb-glx0-dev libxcb-randr0-dev libxrandr-dev libxkbcommon-dev libwayland-dev \
-	 libgtest-dev pkg-config libgtk2.0-dev wget
+	 libgtest-dev pkg-config libgtk2.0-dev wget
```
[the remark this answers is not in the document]
Hyjale: I removed them because I thought that was the culprit behind the CI/CD fails. I will add them back in again.

zed/src/main.cpp
```diff
@@ -53,12 +57,17 @@ class zed_camera_thread : public threadloop {
...
     , image_size{zedm->getCameraInformation().camera_configuration.resolution}
     {
+				runtime_parameters.sensing_mode = SENSING_MODE::STANDARD;
         // Image setup
-        imageL_zed.alloc(image_size.width, image_size.height, MAT_TYPE::U8_C4, MEM::CPU);
-        imageR_zed.alloc(image_size.width, image_size.height, MAT_TYPE::U8_C4, MEM::CPU);
+        imageL_zed.alloc(image_size.width, image_size.height, MAT_TYPE::U8_C1, MEM::CPU);
+        imageR_zed.alloc(image_size.width, image_size.height, MAT_TYPE::U8_C1, MEM::CPU);
+				rgb_zed.alloc(image_size.width, image_size.height, MAT_TYPE::U8_C4, MEM::CPU);
+				depth_zed.alloc(image_size.width, image_size.height, MAT_TYPE::F32_C1);
```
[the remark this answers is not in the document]
Hyjale: Perhaps I forgot to add it in? This ZED code works with KinectFusion, however.

realsense/plugin.cpp
Hyjale: It seems like there is a way to turn off the emitter according to this: https://github.com/IntelRealSense/librealsense/pull/3066.
The `realsense-viewer` also has an option to turn off the emitter as well.

If we can turn off the emitter, the images look like regular grayscale images.

realsense/plugin.cpp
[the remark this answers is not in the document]
Hyjale: If we can call for pose data instead in RealSense, that would be more efficient. However, D455 does not have pose data built-in, while the T265 does.

common/data_format.hpp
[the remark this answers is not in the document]
mhuzai: Gotcha. Please add a timestamp here as well. It will most certainly be required.

zed/src/main.cpp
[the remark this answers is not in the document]
mhuzai: Better to add it for clarity and uniformity.

realsense/plugin.cpp
mhuzai: Oh, very interesting. Give it a shot and see how it looks!

realsense/plugin.cpp
```diff
@@ -0,0 +1,183 @@
...
+                    Eigen::Vector3f av = {gyro_data.x, gyro_data.y, gyro_data.z};
+
+                    // Time as ullong (nanoseconds)
+                    ullong imu_time = static_cast<ullong>(ts * 1000000);
+
+                    // Time as time_point
+                    using time_point = std::chrono::system_clock::time_point;
+                    time_point uptime_timepoint{std::chrono::duration_cast<time_point::duration>(std::chrono::nanoseconds(imu_time))};
+                    std::time_t time2 = std::chrono::system_clock::to_time_t(uptime_timepoint);
+                    time_type t = std::chrono::system_clock::from_time_t(time2);
```
mhuzai, line 113: Too convoluted. Look at how ZED does it.

zed/src/main.cpp
```diff
@@ -30,10 +32,12 @@ std::shared_ptr<Camera> start_camera() {
...
     init_params.camera_resolution = RESOLUTION::VGA;
-    init_params.coordinate_units = UNIT::METER;
+    init_params.coordinate_units = UNIT::MILLIMETER; // for kf
     init_params.coordinate_system = COORDINATE_SYSTEM::RIGHT_HANDED_Z_UP_X_FWD; // Coordinate system used in ROS
-    init_params.camera_fps = 60;
-    init_params.depth_mode = DEPTH_MODE::NONE;
+    init_params.camera_fps = 15;
+    init_params.depth_mode = DEPTH_MODE::PERFORMANCE;
+    init_params.depth_stabilization = true;
+		// init_params.depth_minimum_distance = 0.1;
```
mhuzai, line 40: Set exposure to 4% here.

realsense/plugin.cpp
```diff
@@ -0,0 +1,183 @@
...
+        {
+            cfg.disable_all_streams();
+            cfg.enable_stream(RS2_STREAM_ACCEL, RS2_FORMAT_MOTION_XYZ32F, ACCEL_RATE); // adjustable to 0, 63 (default), 250 hz
+            cfg.enable_stream(RS2_STREAM_GYRO, RS2_FORMAT_MOTION_XYZ32F, GYRO_RATE); // adjustable set to 0, 200 (default), 400 hz
+            cfg.enable_stream(RS2_STREAM_INFRARED, 1, IMAGE_WIDTH, IMAGE_HEIGHT, RS2_FORMAT_Y8, FPS);
+            cfg.enable_stream(RS2_STREAM_INFRARED, 2, IMAGE_WIDTH, IMAGE_HEIGHT, RS2_FORMAT_Y8, FPS);
+            cfg.enable_stream(RS2_STREAM_COLOR, IMAGE_WIDTH, IMAGE_HEIGHT, RS2_FORMAT_BGR8, FPS);
+            cfg.enable_stream(RS2_STREAM_DEPTH, IMAGE_WIDTH, IMAGE_HEIGHT, RS2_FORMAT_Z16, FPS);
+            profiles = pipe.start(cfg, [&](const rs2::frame& frame) { this->callback(frame); });
+            profiles.get_device().first<rs2::depth_sensor>().set_option(RS2_OPTION_EMITTER_ENABLED, 0.f);
```
mhuzai, line 48: If this is the line that disables the IR emitter, add a comment saying so.

zed/src/main.cpp
```diff
@@ -176,6 +198,12 @@ class zed_imu_thread : public threadloop {
             imu_time,
         });
 
+        if (rgb && depth) {
+            _m_rgb_depth->put(new rgb_depth_type{
+                    rgb,
+                    depth
```
mhuzai, line 204: Add timestamp.

scripts/install_apt_deps.sh
```diff
@@ -1,9 +1,11 @@
-#!/bin/bash
+ #!/bin/bash
 
 . /etc/os-release
-sudo apt install -y software-properties-common curl
+sudo apt install -y software-properties-common curl gnupg
```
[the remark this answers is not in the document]
Hyjale: I think the dependency 3 lines down needs gnupg. I remember the ci complaining about this.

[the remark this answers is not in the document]
mhuzai: Oh, interesting. We fixed it in CI by adding it to the Dockerfile. But now that we need it here, we should remove it from the Dockerfile. So please remove it from there (line 12 of Dockerfile). Also, make it `gnupg2` here.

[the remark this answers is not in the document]
mhuzai: Address the above. Rest looks good.

mhuzai (approved):

PR #164, opened 2020-11-02
MIXER Integration for ADA center demo

jheo4:
Todo Lists

- [x] add mixer into illixr build system
- [x] add integration points on both sides
- [x] test simple dataflow
- [x] add frame roundtrip
- [ ] render the frame as a background

### DIFF
--- a/configs/mixer.yaml
+++ b/configs/mixer.yaml
@@ -1,0 +1,19 @@
+plugins:
+  - path: ground_truth_slam/
+  - path: offline_imu_cam/
+  - path: open_vins/
+  - path: pose_prediction/
+  - path: mxre_reader/
+  - path: mxre_writer/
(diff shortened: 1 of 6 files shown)


mhuzai:
Thanks for the PR @jheo4. Looking at the todo list, I'm assuming this is a work in progress, so I marked it as so. Two things before we merge into mainline ILLIXR (whenever that is): 1) we'll need some documentation for the MXRE plugin and 2) the MXRE app has to be a separate application from gldemo.

mhuzai:
@jheo4, @JeffreyZh4ng, do we need to keep this PR around given that it is quite stale at this point? Also, does `feat-offloaded-slam` supersede this?

jheo4:
@mhuzai I close this PR. As you said, we have feat-offloaded-slam. :)

Issue #173, opened 2020-11-05
Clean up OpenGL code


mhuzai:
The OpenGL code inside ILLIXR can be considerably cleaned up. TODOs:
- Remove unnecessary GLFW dependencies from gldemo and timewarp (it seems GLFW is only required for time)
- Remove extraneous glXMakeCurrent calls
- Remove gldemo's 1x1 window
- Move gldemo's setup code from `start()` to `setup()`
- Remove creation of generic context in runtime_impl. Instead, gldemo should publish its context on SB and timewarp should create a new shared context using the published context. This would make the flow identical to OpenXR where the app creates and issue first and then passes it to the runtime.
- In general, each thread should have its own context (which can be shared if so desired) instead of passing around the same context across threads (which requires synchronization and causes data races otherwise, as in #171).

Zee2:
We cannot have multiple GL contexts for each thread: the context must be shared, unfortunately.

mhuzai:
Having shared contexts and different contexts are two different things. We already have multiple contexts in our system: look at the constructors of runtime_impl and gldemo. They are shared, but they are different.

Zee2:
I didn't even know that runtime_impl was creating its own GL context. Where/who/what/when did that happen? Why?

charmoniumQ:
GLFW removed in #227 and #250.

astro-friedel:
We are moving away from OpenGL and adopting Vulkan, in part due to the context issues on Android devices.

PR #174, opened 2020-11-05
Don't make GL context current in runtime thread

mhuzai:
Reviewers, please see comment in commit and #173 for more details.


apache8080 (approved):
LGTM

Issue #45, opened 2020-07-26
Run clang-format


charmoniumQ:
- Deliberate on the format
- Also standardize naming conventions (always use `_m_` for private or never)
- Figure out how to enforce (make part of CI?)

mhuzai:
We really need to get this conversation started. Indentation is pretty much non-existent in our code at this point.

PR #175, opened 2020-11-06
Fix ground truth

### LINKED ISSUE #117: Fix Ground Truth
The current ground truth plugin is unable to find a ground truth pose at any moment in time.

mhuzai:
Fixes #117.

Note that OV is appearing to drift significantly. This PR only fixes the calculation of ground truth. Fixing of OV's drift should happen in another PR.


apache8080 (dismissed):

pose_prediction/plugin.cpp
```diff
@@ -14,32 +14,38 @@ class pose_prediction_impl : public pose_prediction {
...
-				.orientation = Eigen::Quaternionf{1, 0, 0, 0},
-			}
-		);
-    }
+		pose_type offset_pose;
+
+		// Subtract offset if valid pose pointer, otherwise use zero pose
+		if (pose_ptr) {
+			offset_pose = *pose_ptr;
+			offset_pose.position -= *(_m_ground_truth_offset->get_latest_ro());
```
charmoniumQ, line 40:
```suggestion
			auto offset = _m_ground_truth_offset->get_latest_ro();
			assert(offset);
			offset_pose.position = offset;
```

We should check the pointer returned by switchboard. It will be null if no data has been pushed.

ground_truth_slam/plugin.cpp
```diff
@@ -26,29 +27,52 @@ class ground_truth_slam : public plugin {
...
 		true_pose->sensor_time = datum->time;
-		// std::cout << "The pose was found at " << true_pose->position[0] << ", " << true_pose->position[1] << ", " << true_pose->position[2] << std::endl; 
-
 		_m_true_pose->put(true_pose);
+
+		// Ground truth position offset is the first ground truth position
+		if (first_time) {
+			first_time = false;
+			_m_ground_truth_offset->put(new Eigen::Vector3f{true_pose->position});
+		}
```
charmoniumQ, line 48:
```suggestion
		// Ground truth position offset is the first ground truth position
		if (first_time) {
			first_time = false;
			_m_ground_truth_offset->put(new Eigen::Vector3f{true_pose->position});
		}
		_m_true_pose->put(true_pose);
```

Reorder, because consumer assumes that `_m_true_pose` nonnull =>  `_m_ground_truth_offset` nonnull.

charmoniumQ (approved):

pose_prediction/plugin.cpp
mhuzai: Is this necessary given that `_m_ground_truth_offset` would be non-null if `pose_ptr` is non-null? (no relaxed atomics)

charmoniumQ: I agree that it is not necessary, because _the current implementation_ translates each put into a CAS. I still think it's a good idea anyway.

charmoniumQ (approved):

ground_truth_slam/plugin.cpp
```diff
@@ -26,29 +27,51 @@ class ground_truth_slam : public plugin {
...
 
 private:
 	const std::shared_ptr<switchboard> sb;
 	std::unique_ptr<writer<pose_type>> _m_true_pose;
+	std::unique_ptr<writer<Eigen::Vector3f>> _m_ground_truth_offset;
 
 	const std::map<ullong, sensor_types> _m_sensor_data;
 	std::map<ullong, sensor_types>::const_iterator _m_sensor_data_it;
+
+	bool first_time{true};
```
e3m3, line 74: Should this be set in the constructor?

ground_truth_slam/plugin.cpp
```diff
@@ -26,29 +27,51 @@ class ground_truth_slam : public plugin {
 	}
 
 	void feed_ground_truth(const imu_cam_type *datum) {
-		ullong rounded_time = datum->dataset_time;
+		ullong rounded_time = floor(datum->dataset_time / 10000);
```
e3m3, line 30: Is there an implicit cast here?

ground_truth_slam/plugin.cpp
```diff
@@ -26,29 +27,51 @@ class ground_truth_slam : public plugin {
...
+				  << true_pose->position[0] << ", "
+				  << true_pose->position[1] << ", "
+				  << true_pose->position[2] << ")"
+				  << " | "
+				  << "Quat: ("
+				  << true_pose->orientation.w() << ", "
+				  << true_pose->orientation.x() << ", "
+				  << true_pose->orientation.y() << ","
+				  << true_pose->orientation.z() << ")"
+				  << std::endl;
```
e3m3, line 62:
```suggestion
		std::cout << "Ground truth pose was found at T: " << rounded_time
                        << " | "
                        << "Pos: ("
                        << true_pose->position[0] << ", "
                        << true_pose->position[1] << ", "
                        << true_pose->position[2] << ")"
                        << " | "
                        << "Quat: ("
                        << true_pose->orientation.w() << ", "
                        << true_pose->orientation.x() << ", "
                        << true_pose->orientation.y() << ","
                        << true_pose->orientation.z() << ")"
                        << std::endl;
```

pose_prediction/plugin.cpp
```diff
@@ -14,33 +14,40 @@ class pose_prediction_impl : public pose_prediction {
...
         , _m_vsync_estimate{sb->subscribe_latest<time_type>("vsync_estimate")}
     { }
 
-    // No paramter get_fast_pose() should just predict to the next vsync
-	// However, we don't have vsync estimation yet.
-	// So we will predict to `now()`, as a temporary approximation
+    // No parameter get_fast_pose() predicts to the next vsync if a vsync
+    // estimate is available, and predicts to `now` otherwise
     virtual fast_pose_type get_fast_pose() const override {
 		const time_type *vsync_estimate = _m_vsync_estimate->get_latest_ro();
```
e3m3, line 24:
```suggestion
		const time_type * const vsync_estimate = _m_vsync_estimate->get_latest_ro();
```

pose_prediction/plugin.cpp
```diff
@@ -14,33 +14,40 @@ class pose_prediction_impl : public pose_prediction {
...
-				.sensor_time = std::chrono::system_clock::now(),
-				.position = Eigen::Vector3f{0, 0, 0},
-				.orientation = Eigen::Quaternionf{1, 0, 0, 0},
-			}
-		);
     }
 
+	virtual pose_type get_true_pose() const override {
+		const auto *pose = _m_true_pose->get_latest_ro();
+		const auto *offset = _m_ground_truth_offset->get_latest_ro();
```
e3m3, line 35:
```suggestion
		const auto * const pose = _m_true_pose->get_latest_ro();
		const auto * const offset = _m_ground_truth_offset->get_latest_ro();
```

pose_prediction/plugin.cpp
```diff
@@ -14,33 +14,40 @@ class pose_prediction_impl : public pose_prediction {
...
+		pose_type offset_pose;
+
+		// Subtract offset if valid pose and offset, otherwise use zero pose
+		if (pose && offset) {
```
e3m3, line 39: Can `(pose && !offset) || (!pose && offset)` be true? If so, when?

ground_truth_slam/plugin.cpp
mhuzai: That's what I would have done but the other plugins do it here, so kept it here for "consistency".

ground_truth_slam/plugin.cpp
mhuzai: I know where you're coming from, but I wouldn't worry about spacing right now. Depending upon your tab settings, the original code looks aligned. We should just circle back to this when we have solution to #45.

ground_truth_slam/plugin.cpp
mhuzai: Possibly, but the compiler didn't return any warnings.

pose_prediction/plugin.cpp
mhuzai: Hypothetically, if switchboard moved to a weaker memory model, writes to two different streams could finish out of order, resulting in one stream being valid and the other being invalid.

Checking that both are valid is safer because it assumes nothing about the ordering of writes on the producer's end or about the producer actually writing to both.

ground_truth_slam/plugin.cpp
charmoniumQ: There is a safe implicit cast here.

While this will work, I prefer to change `ground_truth_slam/data_loading.hpp:46`. This way, you never have to ask what unit the time is in, because it is _always_  in dataset units.

ground_truth_slam/plugin.cpp
e3m3: Totally, that's why I also made it a suggestion, so that we can't just commit in place as we go. I'll ignore them from now on until we handle #45.

pose_prediction/plugin.cpp
e3m3: Awesome! Can we add that in the comments? :smiley:

ground_truth_slam/plugin.cpp
e3m3: Gotcha. If you don't mind, lets clean as we go, since handling #45 likely won't catch issues like this.

ground_truth_slam/plugin.cpp
e3m3: I am not familiar with the term - what do you mean @charmoniumQ?

charmoniumQ: By "dataset units" I mean the unit of time our dataset uses (in this case nanoseconds). I think it's bad practice to divide by 1000 just to make up for another place in code we divide by 1000. Instead, I'm suggesting we remove both divisions and use the same units as our input.

mhuzai: Good catch. Will fix.

pose_prediction/plugin.cpp
mhuzai: Hah, will do.

therishidesai (dismissed):
Everything looks good to me. Just respond to @e3m3's comments and merge.

charmoniumQ (approved):

e3m3 (approved):

PR #177, opened 2020-11-06
Issue 147 kimera vio

Hyjale:
[] Kimera-VIO integration
[] config / runner changes


configs/native.yaml
```diff
@@ -13,6 +13,7 @@ data:
       download_url: 'http://robotics.ethz.ch/~asl-datasets/ijrr_euroc_mav_dataset/vicon_room1/V1_02_medium/V1_02_medium.zip'
 demo_data: demo_data/
 loader:
+  kimera_path: .cache/paths/https%c%s%sgithub.com%sILLIXR%sKimera-VIO/
```
mhuzai, line 16: Don't harcode. What if someone wants to run a local copy of Kimera and not the git repo one? This shouldn't be a part of the config. This should be determined and set by the runner.

configs/rt_openvins_plugins.yaml
```diff
@@ -0,0 +1,9 @@
+plugin_group:
```
mhuzai, line 1: Why a separate file for OV? Why not just comment out OV in rt_slam_plugins?

mhuzai (commented):
It's fine for now but please use better names for PRs in the future (not just the branch name).

configs/native.yaml
apache8080: It is fine that this is hardcoded and there is a future issue to make this less ugly. This is the best we can do right now.

configs/rt_openvins_plugins.yaml
apache8080: I think it is better to have a separate file for OV just so people can differentiate between the two. Also it points people toward having multiple configs. Commenting things out of configs seems a bit uglier to me.

configs/native.yaml
mhuzai: People will most certainly make changes to Kimera-VIO during the hackathon (instrumentation, etc.) and this would become problematic then. We can ask them to update this path manually but I don't understand what the blocker is in doing this from inside the runner. The runner always knows the full path because it builds the plugins.

apache8080: They wouldn't have to change this during the hackathon.

configs/rt_openvins_plugins.yaml
mhuzai: There are 2 SLAMs x 2 integrators, so that's 4 configs. That's too much.

configs/native.yaml
apache8080: If they do we can show them how. The changes to the runner are bigger than just a couple of days of work.

charmoniumQ: @mhuzai Let's get this one in and push another PR correcting it--if we have time.

mhuzai (dismissed):

apache8080 (dismissed):

Dockerfile
```diff
@@ -1,6 +1,8 @@
 FROM ubuntu:18.04
 
-ENV DEBIAN_FRONTEND noninteractive
+ENV DEBIAN_FRONTEND=noninteractive
+ENV TZ=Europe/Moscow
+RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone
```
mhuzai, line 5: As discussed yesterday, move to line 14 or something.

scripts/install_opengv.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+git clone https://github.com/laurentkneip/opengv.git "${temp_dir}/opengv"
+
+cmake \
+	-S "${temp_dir}/opengv" \
+	-B "${temp_dir}/opengv/build" \
+    -DEIGEN_INCLUDE_DIR="${temp_dir}/gtsam/gtsam/3rdparty/Eigen" \
+    -DEIGEN_INCLUDE_DIRS="${temp_dir}/gtsam/gtsam/3rdparty/Eigen"
```
mhuzai, line 9: Minor, but use same format and indentation as other files for consistency. So `-D <FLAG>` and not `-D<FLAG>`.

mhuzai (commented):
Left some minor comments, but looks good in general.

scripts/install_opencv.sh
```diff
@@ -2,17 +2,40 @@
...
+	    -D BUILD_EXAMPLES=OFF \
+	    -D BUILD_JAVA=OFF \
+	    -D WITH_OPENGL=ON \
+	    -D WITH_VTK=ON \
+	    -D WITH_CUDA=ON \
+	    -D WITH_CUBLAS=1 \
+	    -D CUDA_NVCC_FLAGS=-allow-unsupported-compiler \
+	    -D BUILD_opencv_cudacodec=OFF \
+	    -D OPENCV_EXTRA_MODULES_PATH="${temp_dir}/opencv_contrib/modules"
+fi
```
charmoniumQ, line 38:
```suggestion
if ! nvcc --version &> /dev/null
then
    echo "nvcc not found"
else
    echo "nvcc found"
    extra_cmake_args="
	    -D WITH_CUDA=ON
	    -D WITH_CUBLAS=1
	    -D CUDA_NVCC_FLAGS=-allow-unsupported-compiler
	    -D BUILD_opencv_cudacodec=OFF"
fi
cmake \
	    -S "${temp_dir}/opencv" \
	    -B "${temp_dir}/opencv/build" \
	    -D CMAKE_BUILD_TYPE=RelWithDebInfo \
	    -D CMAKE_INSTALL_PREFIX=/usr/local \
	    -D BUILD_TESTS=OFF \
	    -D BUILD_PERF_TESTS=OFF \
	    -D BUILD_EXAMPLES=OFF \
	    -D BUILD_JAVA=OFF \
	    -D WITH_OPENGL=ON \
	    -D WITH_VTK=ON \
	    -D OPENCV_EXTRA_MODULES_PATH="${temp_dir}/opencv_contrib/modules" \
	    ${extra_cmake_args}
```

charmoniumQ: Less repetition, easier to add flags to both, or just one case.

charmoniumQ (dismissed):

mhuzai (approved):

charmoniumQ (approved):

PR #180, opened 2020-11-06
Remove submodules

charmoniumQ:


.gitmodules
```diff
@@ -1,12 +1,3 @@
 [submodule "slam1/okvis"]
 	path = slam1/okvis
 	url = https://github.com/ILLIXR/okvis.git
```
mhuzai, line 3: Why not remove okvis? That is, why not remove `.gitmodules` altogether?

charmoniumQ: It's weird because the OKVIS code is not in the working tree, but the `.gitmodules` entry is still there. Normally, one should not manually edit the file, but I will try this time.

mhuzai (approved):
Good to go once updated with latest master.

apache8080 (approved):

Issue #176, opened 2020-11-06
Cannot build librealsense2-dkms for Ubuntu 20 LTS


e3m3:
The IntelRealSense project does not yet support the LTS kernels for Ubuntu 20.04 (See installation process [here](https://github.com/IntelRealSense/librealsense/blob/master/doc/distribution_linux.md)).

From a fresh, up-to-date copy of Ubuntu 20.04, following the [ILLIXR getting started page](https://illixr.github.io/ILLIXR/getting_started) while using the 'master' branch produces the following error:
`Building initial module for 5.4.0-52-generic
Error!  The /var/lib/dkms/librealsense2-dkms/1.3.14/5.4.0-52-generic/x86_64/dkms.conf for module librealsense2-dkms includes a BUILD_EXCLUSIVE directive which
does not match this kernel/arch.  This indicates that it should not be built.
Skipped.`

It appears that this issue has been encountered before, and that some success can be found by building the SDK manually using the project's "backend" installation process. See [Issue #6436](https://github.com/IntelRealSense/librealsense/issues/6436) at the IntelRealSense repository.

mhuzai:
Hi stranger. Thanks for the detailed info. It seems the easiest way forward is to install librealsense automatically only if the OS is 18.04. For non-18.04 OSes, we can have the users install it themselves.

If you can change scripts/install_apt_deps.sh accordingly and open a PR, that would be great.

Issue #179, opened 2020-11-06
Fix GTSAM RelWithDebInfo Build


mhuzai:
If you build GTSAM with RelWithDebInfo, the final `.so` is not called `libgtsam.so`. It is called something along the lines of `libgtsam_RelWithDebInfo.so`, and so the linker isn't able to pick it up when compiling `gtsam_integrator` or `kimera-vio`. This should be fixed by changing the GTSAM build system so that it installs the correctly named `.so` in `RelWithDebInfo` mode.

Issue #91, opened 2020-09-04
Improve install_deps.sh


mhuzai:
Print detailed message listing exactly what will be installed.

e3m3:
For the package dependencies, we can print all packages to be sent to 'apt-get'. 'apt-get' subsequently which packages will actually be installed. Another option is to find out which packages will be installed by 'apt-get' using the provided package list, and print that instead.

For the other install scripts, 'cmake' is generally used to build the packages, followed by a 'make install'. What is the correct behavior for these dependencies? We already prompt the user asking whether each dependent library should be installed.

mhuzai:
The original intent of this was that we should tell the user what the intent of the script is _before_ we start prompting them for individual dependencies. This would be more detailed information than just "install DBoW2?" It would say what we need DBoW2 for, for instance.
But this may be unnecessary now. The queries give enough high-level information, and if we print the apt packages, that might be sufficient. Open to ideas.

What did you think was lacking in install_deps when you ran it?

e3m3:
Based on what you said, I could see us printing list containing the relevant name for each prompt, and then subsequently do the prompts.

Regarding the apt packages, the package list is currently printed after the user prompt, but before installation (without an additional prompt). Given the current structure of the install scripts, this behavior is okay with me.

In general, I dislike prompts in scripts, but I could see the need for them for ILLIXR users. If the result of the large dependency list is that the project is tied to a handful of Ubuntu LTS distributions only, then I would say a more automated approach that detects missing/conflicting package or sets up a virtual environment would be best.

mhuzai:
Yes, I agree that a more automated approach would be best but given our limited resources, what we have at the moment is alright.

As for the printing behavior, I think we're in agreement.

charmoniumQ:
If we're already printing the packages that `apt-get` installs, I'd suggest also silencing the `stdout`. It's pretty redundant, anything significant goes to `stderr`, and I care about the signal-to-noise ratio of the script, so warnings do not get missed.

PR #185, opened 2020-11-10
Automatically install Intel RealSense packages for Ubuntu 18 LTS only

e3m3:
Package dependencies for IntelRealSense are only automatically installed for Ubuntu 18 LTS users. Users running unsupported distributions receive a warning indicating that IntelRealSense should be installed/built manually. (#176)


charmoniumQ (dismissed):
Thanks for cleaning this up.

scripts/install_apt_deps.sh
```diff
@@ -1,21 +1,92 @@
...
+echo "Detected OS: '${PRETTY_NAME}'"
+if [ "${ID}" == "ubuntu" ] && [ "${VERSION_ID}" == "18.04" ]; then
+    use_realsense="yes"
+    pkg_dep_list+="${pkg_dep_list_realsense} "
+elif [ "${ID}" == "ubuntu" ] && [ "${VERSION_ID}" == "20.04" ]; then
+    use_realsense="no"
+    print_warning "${pkg_warn_msg_realsense}"
+else
+    use_realsense="no"
+    print_warning "${pkg_warn_msg_realsense}"
```
mhuzai, line 74: Both branch paths are functionally equivalent, so why the distinction?

scripts/install_apt_deps.sh
```diff
@@ -1,21 +1,92 @@
...
+pkg_build_url_realsense="https://github.com/IntelRealSense/librealsense/blob/master/doc/installation.md"
+pkg_warn_msg_realsense="Currently, IntelRealSense does not support binary package installations for Ubuntu 20 LTS kernels, or other non-Ubuntu Linux distributions. If your project requires IntelRealSense support, please build and install the IntelRealSense SDK from source. For more information, visit '${pkg_install_url_realsense}' and '${pkg_build_url_realsense}'."
+
+# Check for distribution support of IntelRealSense
+# For supported distributions, automatically add the required install
+# package dependencies
+echo "Detected OS: '${PRETTY_NAME}'"
+if [ "${ID}" == "ubuntu" ] && [ "${VERSION_ID}" == "18.04" ]; then
+    use_realsense="yes"
+    pkg_dep_list+="${pkg_dep_list_realsense} "
```
mhuzai, line 68: Minor: trailing space after `}`.

scripts/install_apt_deps.sh
```diff
@@ -1,21 +1,92 @@
...
+
+
+### IntelRealSense package dependencies setup ###
+
+pkg_dep_list_realsense="librealsense2-dkms librealsense2-utils"
+pkg_install_url_realsense="https://github.com/IntelRealSense/librealsense/blob/master/doc/distribution_linux.md"
+pkg_build_url_realsense="https://github.com/IntelRealSense/librealsense/blob/master/doc/installation.md"
+pkg_warn_msg_realsense="Currently, IntelRealSense does not support binary package installations for Ubuntu 20 LTS kernels, or other non-Ubuntu Linux distributions. If your project requires IntelRealSense support, please build and install the IntelRealSense SDK from source. For more information, visit '${pkg_install_url_realsense}' and '${pkg_build_url_realsense}'."
+
+# Check for distribution support of IntelRealSense
```
mhuzai, line 62: Minor: Intel RealSense

scripts/install_apt_deps.sh
```diff
@@ -1,21 +1,92 @@
...
+### Initial setup ###
+
+# Source environment variables describing the current OS distribution
 . /etc/os-release
 
+
+### Package dependencies setup ###
+
+# List of common package dependencies for the ILLIXR project
+pkg_dep_list_common="
```
mhuzai, line 13: Curious, any reason behind the ordering here?

scripts/install_apt_deps.sh
```diff
@@ -1,21 +1,92 @@
...
+
+### Package dependencies installation ###
+
 sudo apt install -y software-properties-common curl gnupg2
 curl https://apt.kitware.com/keys/kitware-archive-latest.asc | sudo apt-key add -
 sudo apt-add-repository -u -y "deb https://apt.kitware.com/ubuntu/ ${UBUNTU_CODENAME} main"
-sudo apt-key adv --keyserver keys.gnupg.net --recv-key F6E65AC044F831AC80A06380C8B3A55A6F3EFCDE || sudo apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-key F6E65AC044F831AC80A06380C8B3A55A6F3EFCDE
-sudo add-apt-repository "deb http://realsense-hw-public.s3.amazonaws.com/Debian/apt-repo bionic main" -u
+
+# If supported, add the gpg keys and repository for IntelRealSense
```
mhuzai, line 84: Minor: Intel RealSense

scripts/install_apt_deps.sh
```diff
@@ -1,21 +1,92 @@
...
-	 git clang-10 make cmake libc++-dev libc++abi-dev unzip \
-	 libsqlite3-dev libeigen3-dev libboost-all-dev libatlas-base-dev libsuitesparse-dev libblas-dev libtbb-dev \
-	 glslang-tools libsdl2-dev libglu1-mesa-dev mesa-common-dev freeglut3-dev libglew-dev glew-utils libglfw3-dev \
-	 libusb-dev libusb-1.0 libudev-dev libv4l-dev libhidapi-dev \
-	 build-essential libx11-xcb-dev libxcb-glx0-dev libxcb-randr0-dev libxrandr-dev libxkbcommon-dev libwayland-dev \
-	 libepoxy-dev libdrm-dev libgbm-dev libx11-dev libvirglrenderer-dev libpulse-dev \
-	 libgtest-dev pkg-config libgtk2.0-dev wget xvfb librealsense2-dkms librealsense2-utils \
-     libjpeg-dev libpng-dev libtiff-dev libvtk6-dev libgtk-3-dev libparmetis-dev gfortran \
-	 libgflags-dev libgoogle-glog-dev \
+sudo apt-get install -y ${pkg_dep_list}
```
mhuzai, line 92: Let's just move to `apt` instead of `apt-get`. We use `apt` on line 80.

scripts/install_apt_deps.sh
e3m3: I tried not to change it too much but I tried to follow a few rules:
- If a set of packages are logically grouped, I put them on the same line. May have missed some groups due to unfamiliarity of when/why they were added.
- General dev dependencies like git I tried putting at the top.
- Specialized dev dependencies (e.g. lib.*-dev) and new dependencies at the bottom. These can be broken out for specific features/dependencies (a la realsense)  .
- Further rearrangement can be made with some hints as to what should be grouped.

scripts/install_apt_deps.sh
e3m3: I assumed this would change for 20.04 in the near future (I believe the Intel RealSense SDK will be getting an update based on the external repo issues referenced in #176, so I left that branching path in. Remove still?

scripts/install_apt_deps.sh
e3m3: The space is required for future concatenations. I used a suffix space instead of prefix.

scripts/install_apt_deps.sh
e3m3: Also, we may have the option of doing the "backend" build process specifically only for Ubuntu 20.04 in the branch,

mhuzai: If they add 20.04 support, the `if` condition changes, not the `elif`. But I see your point about manually building on 20.04. How hard is the 20.04 build? Is it robust enough that we can safely put it in a script? If yes, leave the `elif` in. If no, remove it -- best not to waste time doing error-prone builds.

scripts/install_apt_deps.sh
charmoniumQ: I thought `apt` wasn't safe to use within scripts, because the interface is not standardized.

```
SCRIPT USAGE
       The apt(8) commandline is designed as a end-user tool and it may change
       the output between versions. While it tries to not break backward
       compatibility there is no guarantee for it either. All features of
       apt(8) are available in apt-cache(8) and apt-get(8) via APT options.
       Please prefer using these commands in your scripts.
```

mhuzai: Sounds good.

scripts/install_apt_deps.sh
mhuzai: Yeah, that sounds reasonable. Groupings are tricky and its easy to lose track of what is what. But let's take a crack at the easy ones. Add comments in the script to name the groups for future reference.

Display / window manager: `libx11-xcb-dev libx11-dev libxcb-glx0-dev libxcb-randr0-dev libxrandr-dev libxkbcommon-dev libwayland-dev libsdl2-dev`

You already have the "mesa", "GL", and "USB" groups. Should we make one each for threading and linear algebra?

mhuzai: And maybe also an "essentials" or "general" group with stuff like git, cmake, etc.

mhuzai:
@e3m3, while you're at it, can you also tackle #91 in this PR?

mhuzai (approved):
Looks great!

scripts/install_apt_deps.sh
```diff
@@ -1,21 +1,209 @@
...
+function print_warning() { echo -e "\e[31m*Warning* ${1}\e[39m"; }
+
+# Function for flattening a string with items seperated by new lines
+# and white space to a single line seperated by a single space.
+# When calling flatten list, the multi-line list (string) passed as
+# the first argument must be protected by double quotes when expanding
+# a variable.
+# Good sample call site: flat_string=$(flatten_list "${multi_line_string}")
+# Bad sample call site:  flat_string="$(flatten_list ${multi_line_string})"
+function flatten_list() { echo "${1}" | xargs; }
```
charmoniumQ, line 22: Just wondering, do you think it's cleaner to use Bash arrays like:
```
array=(
    elem1
    elem2
)
...
sudo apt-get install -y "${array[@]}"
```

e3m3: Good call! Always forget about bash arrays. :P

e3m3: Unfortunately, it seems that the real issue was passing in the delimiter separated list to a bash function, resulting in the same gross workaround at call sites even when using bash array. I did keep the bash arrays though, since it was still a bit cleaner :P.

scripts/install_apt_deps.sh
```diff
@@ -1,21 +1,252 @@
...
+    libparmetis-dev
+    libatlas-base-dev
+) # End list
 
-sudo apt install -y software-properties-common curl gnupg2
-curl https://apt.kitware.com/keys/kitware-archive-latest.asc | sudo apt-key add -
-sudo apt-add-repository -u -y "deb https://apt.kitware.com/ubuntu/ ${UBUNTU_CODENAME} main"
-sudo apt-key adv --keyserver keys.gnupg.net --recv-key F6E65AC044F831AC80A06380C8B3A55A6F3EFCDE || sudo apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-key F6E65AC044F831AC80A06380C8B3A55A6F3EFCDE
-sudo add-apt-repository "deb http://realsense-hw-public.s3.amazonaws.com/Debian/apt-repo bionic main" -u
+pkg_dep_list_nogroup=(
```
apache8080, line 147: @mhuzai @charmoniumQ should we group these apt packages by the plugins that need them? May help with debugging.

apache8080 (commented):
Nice updates to install deps and the debugging docs. One thing for the debugging docs run `mkdocs serve` locally to make sure they look right.

scripts/install_apt_deps.sh
e3m3: Looks like mkdocs is never installed by the scripts. Seems like a dependency only required for ILLIXR devs, so I think if there are many of these types of packages eventually, we can add a group for them that gets installed with a dev flag.

apache8080: Yeah, currently we only manually install it, but I like the idea of a dev flag for install deps.

mhuzai (commented):
This looks great. My only concern is regarding the CUDA install and its robustness. The user can always say no to the CUDA install prompt, but it still makes me uncomfortable. The probability of something breaking is very high. Thoughts?

charmoniumQ (approved):
LGTM if you address Huzi's past three comments.

scripts/install_apt_deps.sh
```diff
@@ -1,21 +1,301 @@
...
+# If supported, add the keys and repository for CUDA (for GPU plugin support)
+if [ "${use_cuda}" == "yes" ]; then
+    distro_cuda="ubuntu1804"
+    arch_cuda="x86_64"
+    repo_url_cuda="https://developer.download.nvidia.com/compute/cuda/repos/${distro_cuda}/${arch_cuda}"
+    key_srv_url_cuda="${repo_url_cuda}/7fa2af80.pub"
+    add_repo "${key_srv_url_cuda}" "${repo_url_cuda}" "/"
+
+    path_cmd_cuda='export PATH=/usr/local/cuda-11.1/bin${PATH:+:${PATH}}'
+    lib64_cmd_cuda='export LD_LIBRARY_PATH=/usr/local/cuda-11.1/lib64${LD_LIBRARY_PATH:+:${LD_LIBRARY_PATH}}'
```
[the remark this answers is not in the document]
e3m3: I agree this is an issue. Previously we relied on the user to go through the documentation for their system. This path setup is part of the [post installation actions](https://docs.nvidia.com/cuda/cuda-installation-guide-linux/index.html#environment-setup) shared for all configurations. Our scripts runs through most of the current process for an Ubuntu system for CUDA 11, which makes it easier to get started from a new install. Do you think a message stating that this is only for Ubuntu and CUDA 11?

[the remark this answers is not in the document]
mhuzai: Yeah, I think we should add such a message, and also make it bold and clear (all caps or whatever) that the user should *only* do this if they are sure there won't be issues -- out of everything we install, NVIDIA drivers and CUDA have the highest likelihood of bricking someone's machine, and we don't want to be liable for that (our license says this, but who really reads those).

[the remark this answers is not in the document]
e3m3: How do lines `scripts/install_apt_deps.sh:231-236` look to you?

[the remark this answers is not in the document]
mhuzai: They look good. But I was thinking of `install_deps.sh:93`. Instead of a simple query like for the other installs, make the CUDA query big, bold, and red (jk). 😛

charmoniumQ (dismissed):
Note to reviewers: If the CI fails at the _end_ of a run with `Segmentation fault (core dumped)`, then it's a pre-existing issue, not originating in this PR.

charmoniumQ (dismissed):

scripts/install_apt_deps.sh
```diff
@@ -1,21 +1,301 @@
...
+# Source environment variables describing the current OS distribution
 . /etc/os-release
+kernel_version="$(uname -r)"
+echo "Detected OS: '${PRETTY_NAME}' on kernel version '${kernel_version}'"
+
+
+### Helper functions ###
+
+# Function for printing a warning message (with red foreground)
+function print_warning() { echo -e "\e[31m*Warning* ${1}\e[39m"; }
```
mhuzai, line 28: Can we use something like this for the CUDA message?

install_deps.sh
```diff
@@ -54,101 +54,113 @@ fi
...
+    temp_dir=/tmp/ILLIXR_deps
+    mkdir -p "${temp_dir}"
+
+    # For local installs
+    opt_dir=/opt/ILLIXR
+    sudo mkdir -p "${opt_dir}"
+    sudo chown $USER: "${opt_dir}"
+
+    echo "* The user will now be prompted to install the following dependencies and optional features:"
+    echo "  Binary packages (apt-get), Docker, OpenCV, Vulkan, gtest, qemu, OpenXR-SDK, gtsam, opengv, DBoW2, Kimera-RPGO, Conda (miniconda3)"
```
mhuzai, line 87: Add CUDA to this list.

apache8080: Do we check to see if cuda is already on the machine? I don't see it scrolling through install_deps

apache8080 (commented):
Do we check to see if cuda is already on the machine? I don't see it scrolling through install_deps.

e3m3:
> Do we check to see if cuda is already on the machine? I don't see it scrolling through install_deps.

Updated `scripts/install_apt_deps.sh` to check for existing components of an install. The install will be skipped if something is detected.

Issue #179: Piggy backing my rewrite of GTSAM's install script with this PR.

clean.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env sh
```
charmoniumQ, line 1: Fine for now; but when we have CLI arguments overriding the config's action, we should remove this script.

configs/clean.yaml
```diff
@@ -0,0 +1,46 @@
+##
+## Configuration to `make clean` all plugins listed below
```
charmoniumQ, line 2:
```suggestion
## Configuration to `make clean` all plugins listed below
## Note that the version here has to match exactly the version in the configuration you intend to run.
```

mhuzai (approved):

charmoniumQ (approved):

Issue #182, opened 2020-11-09
Kimera build fails


mhuzai:
Kimera dependencies should be installed in `/opt` and not `/tmp` because they need to be persistent, just like gtest.

PR #187, opened 2020-11-10
Issue 182: Build Kimera dependencies in opt

apache8080:
We now build kimera dependencies in opt as explained in #182. Also updated the scripts to only use half the cores, so we don't run out of memory on either virtual machines or under powered machines.

### DIFF
--- a/install_deps.sh
+++ b/install_deps.sh
@@ -46,6 +46,9 @@ function y_or_n() {
 
 if [ "${ID_LIKE}" = debian ] || [ "${ID}" = debian ]
 then
+	# Set nproc to either 1 or half the available cores
+	illixr_nproc=$(python3 -c 'import multiprocessing; print( max(multiprocessing.cpu_count() // 2, 1))')
+
 	# For system-wide installs that are not possible via apt
 	temp_dir=/tmp/ILLIXR_deps
 	mkdir -p "${temp_dir}"
--- a/install_deps.sh
+++ b/install_deps.sh
@@ -10,15 +10,38 @@ set -e
 cd "$(dirname "${0}")"
 
 ### Parse args ###
-
+show_help=0
+exit_code=
 assume_yes=
+JOBS=1
 while [[ "$#" -gt 0 ]]; do
     case "${1}" in
         -y|--yes) assume_yes=true ;;
-        *) echo "Unknown parameter passed: ${1}"; exit 1 ;;
+		-h|--help)
+			show_help=1
+			;;
+		-j|--jobs)
+			if [ -n "$2" ] && [ ${2:0:1} != "-" ]; then
+				JOBS=$2
+				shift
+			else
+				echo "Error: Argument for number of jobs is missing"
+				show_help=1
+				exit_code=1
+			fi
+			;;
+        *) 	echo "Error: Unknown parameter passed: ${1}"; show_help=1; exit_code=1 ;;
     esac
     shift
 done
+
+if  [[ "$show_help" -eq 1 ]]; then
+	echo "ILLIXR install_deps:"
+	echo "    -y/--yes - yes to all dependencies"
+	echo "    -j/--jobs - number of jobs/cores/threads for make to use"
+	echo "    -h/--help - help info for install_deps"
+	exit $exit_code
+fi
 
 ### Get OS ###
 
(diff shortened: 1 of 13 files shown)


charmoniumQ (dismissed):

install_deps.sh
```diff
@@ -46,6 +46,9 @@ function y_or_n() {
 
 if [ "${ID_LIKE}" = debian ] || [ "${ID}" = debian ]
 then
+	# Set nproc to either 1 or half the available cores
+	illixr_nproc=$(python3 -c 'import multiprocessing; print( max(multiprocessing.cpu_count() // 2, 1))')
```
mhuzai, line 50: Is there an easy way to make this more intelligent / adaptive? This is only required for machines with low RAM, and possibly even then only for GTSAM, and we would be leaving a lot of performance on the table on machines with more RAM.

install_deps.sh
```diff
@@ -91,19 +94,19 @@ then
...
 		. ./scripts/install_opengv.sh
 	fi
 
-	if [ ! -d "${temp_dir}/DBoW2" ] && y_or_n "Next: Install DBoW2 from source"; then
+	if [ ! -d "${opt_dir}/DBoW2" ] && y_or_n "Next: Install DBoW2 from source"; then
 		. ./scripts/install_dbow2.sh
 	fi
 
-	if [ ! -d "${temp_dir}/Kimera-RPGO" ] && y_or_n "Next: Install Kimera-RPGO from source"; then
+	if [ ! -d "${opt_dir}/Kimera-RPGO" ] && y_or_n "Next: Install Kimera-RPGO from source"; then
```
mhuzai, line 109: I went through the [Kimera installation documentation](https://github.com/MIT-SPARK/Kimera-VIO/blob/master/docs/kimera_vio_install.md) and noticed that all the dependencies except RPGO do `make install`. So only RPGO needs to go in `opt`. 
One edge case is:
1. Install gtsam
2. Reboot machine
3. Install opengv

In this case step 3 will fail because opengv needs gtsam's Eigen and it would have been wiped out from `tmp`. So maybe we can keep gtsam in `opt` just to be safe, but `opengv` and `DBoW2` can easily go in `tmp`.
On the other hand, `tmp` is just a disk space saving optimization, so maybe it is not entirely necessary.

apache8080: I think we should just keep everything in opt for kimera for simplicity.

install_deps.sh
charmoniumQ: I share your frustration, though I don't know of any examples which do something smarter here.

install_deps.sh
```diff
@@ -10,16 +10,39 @@ set -e
...
+				show_help=1
+				exit_code=1
+			fi
+			;;
+        *) 	echo "Error: Unknown parameter passed: ${1}"; show_help=1; exit_code=1 ;;
     esac
     shift
 done
 
+if  [[ "$show_help" -eq 1 ]]; then
```
charmoniumQ, line 38: Beautiful

install_deps.sh
```diff
@@ -10,16 +10,39 @@ set -e
...
+show_help=0
+exit_code=
 assume_yes=
+JOBS=1
 while [[ "$#" -gt 0 ]]; do
     case "${1}" in
         -y|--yes) assume_yes=true ;;
-        *) echo "Unknown parameter passed: ${1}"; exit 1 ;;
+		-h|--help)
+			show_help=1
```
charmoniumQ, line 21: I think setting `exit_code=1` here as well is less surprising behavior.

install_deps.sh
```diff
@@ -91,19 +117,19 @@ then
 		. ./scripts/install_openxr.sh
```
charmoniumQ, line 117: You've only utilized `illixr_nproc` in some of the builds. Is this because you don't expect `OpenXR-SDK`. `gtest`, ... to run out of memory?

install_deps.sh
```diff
@@ -10,16 +10,39 @@ set -e
 cd "$(dirname "${0}")"
 
 ### Parse args ###
-
+show_help=0
+exit_code=
 assume_yes=
+JOBS=1
```
charmoniumQ, line 16: Why not set `illixr_nproc` directly? People used to say about Apple, `JOBS` was superfluous anyway, right xD

install_deps.sh
apache8080: forgot about those, lol

charmoniumQ (approved):

install_deps.sh
```diff
@@ -10,16 +10,42 @@ set -e
...
 while [[ "$#" -gt 0 ]]; do
     case "${1}" in
         -y|--yes) assume_yes=true ;;
-        *) echo "Unknown parameter passed: ${1}"; exit 1 ;;
+		-h|--help)
+			show_help=1
+			exit_code=1
+			;;
+		-j|--jobs)
+			if [ -n "$2" ] && [ ${2:0:1} != "-" ]; then
```
e3m3, line 27: This is fancy :). Maybe a comment here

charmoniumQ (approved):

e3m3 (approved):
Looks good

PR #195, opened 2020-11-17
Fix runner for monado

Hyjale:
- The runner was not building a shared object plugin for the runtime due to a missed line. 
- I have added OpenVINS to SLAM config file to allow for more accessibility (digging through to find the correct version would be a pain for most people).

### DIFF
--- a/configs/rt_slam_plugins.yaml
+++ b/configs/rt_slam_plugins.yaml
@@ -5,5 +5,9 @@ plugin_group:
     path:
       git_repo: https://github.com/ILLIXR/Kimera-VIO
       version: 6ea99d33fc2435a8c514049416dcb2f8a0e31692
+  #- name: OpenVINS
+  #  path:
+  #    git_repo: https://github.com/ILLIXR/open_vins.git
+  #    version: c6a4f8835851d21f87add38260e73df63c6354a1
   - path: gtsam_integrator/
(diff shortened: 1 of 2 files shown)


charmoniumQ:
Good find.

charmoniumQ (approved):

configs/rt_slam_plugins.yaml
```diff
@@ -5,5 +5,9 @@ plugin_group:
     path:
       git_repo: https://github.com/ILLIXR/Kimera-VIO
       version: 6ea99d33fc2435a8c514049416dcb2f8a0e31692
+  #- name: OpenVINS
+  #  path:
+  #    git_repo: https://github.com/ILLIXR/open_vins.git
+  #    version: c6a4f8835851d21f87add38260e73df63c6354a1
```
apache8080, line 11: @Hyjale can you remove this. We decided to not have openvins in the config file. A user can add this themselves.

@mhuzai @charmoniumQ unless we want to change this.

apache8080: We had a separate file for OV before.

apache8080: We can just let this through now and change it later.

apache8080 (approved):

Issue #94, opened 2020-09-04
Configurable ILLIXR timeout


therishidesai:
Currently, ILLIXR timesout with 60 seconds. We need a way to configure this time or turn the timeout off. This should probably be set in the config yaml file for runner.py.

charmoniumQ:
See `ILLIXR_RUN_DURATION` in `runtime/main.cpp`.

astro-friedel:
I believe this has been addressed for some time.

PR #198, opened 2020-11-19
Module parameters now configurable

e3m3:
Configurable parameters passed via yaml to the application as environment variables. Now configurable: refresh rate, framebuffer size, experiment run duration (#94)

### DIFF
--- a/common/gl_util/lib/tiny_obj_loader.h
+++ b/common/gl_util/lib/tiny_obj_loader.h
@@ -2168,6 +2168,7 @@ bool LoadObj(attrib_t *attrib, std::vector<shape_t> *shapes,
   std::stringstream errss;
 
   std::ifstream ifs(filename);
+
   if (!ifs) {
     errss << "Cannot open file [" << filename << "]" << std::endl;
     if (err) {
--- a/gldemo/plugin.cpp
+++ b/gldemo/plugin.cpp
@@ -23,10 +23,10 @@ using namespace ILLIXR;
 static constexpr int   EYE_TEXTURE_WIDTH   = ILLIXR::FB_WIDTH;
 static constexpr int   EYE_TEXTURE_HEIGHT  = ILLIXR::FB_HEIGHT;
 
-static const int EYE_TEXTURE_WIDTH   { ILLIXR::DEFAULT_FB_WIDTH  };
-static const int EYE_TEXTURE_HEIGHT  { ILLIXR::DEFAULT_FB_HEIGHT };
+static const int EYE_TEXTURE_WIDTH   { ILLIXR::FB_WIDTH  };
+static const int EYE_TEXTURE_HEIGHT  { ILLIXR::FB_HEIGHT };
 
-static const double DISPLAY_REFRESH_RATE { ILLIXR::DEFAULT_REFRESH_RATE };
+static const double DISPLAY_REFRESH_RATE { ILLIXR::REFRESH_RATE };
 static const std::chrono::nanoseconds VSYNC_PERIOD { static_cast<size_t>(NANO_SEC/DISPLAY_REFRESH_RATE) };
 static constexpr std::chrono::milliseconds VSYNC_DELAY_TIME {std::size_t{2}};
 
--- a/gldemo/plugin.cpp
+++ b/gldemo/plugin.cpp
@@ -13,6 +13,7 @@
 #include "common/math_util.hpp"
 #include "common/pose_prediction.hpp"
 #include "common/gl_util/obj.hpp"
+#include "common/global_module_defs.hpp"
 #include "shaders/demo_shader.hpp"
 #include "common/global_module_defs.hpp"
 #include "common/error_util.hpp"
@@ -22,7 +23,11 @@ using namespace ILLIXR;
 static constexpr int   EYE_TEXTURE_WIDTH   = ILLIXR::FB_WIDTH;
 static constexpr int   EYE_TEXTURE_HEIGHT  = ILLIXR::FB_HEIGHT;
 
-static constexpr std::chrono::nanoseconds vsync_period {std::size_t(NANO_SEC/60)};
+static const int EYE_TEXTURE_WIDTH   { ILLIXR::DEFAULT_FB_WIDTH  };
+static const int EYE_TEXTURE_HEIGHT  { ILLIXR::DEFAULT_FB_HEIGHT };
+
+static const double DISPLAY_REFRESH_RATE { ILLIXR::DEFAULT_REFRESH_RATE };
+static const std::chrono::nanoseconds VSYNC_PERIOD { static_cast<size_t>(NANO_SEC/DISPLAY_REFRESH_RATE) };
 static constexpr std::chrono::milliseconds VSYNC_DELAY_TIME {std::size_t{2}};
 
 // Monado-style eyebuffers:
@@ -60,7 +65,7 @@ public:
 		if (next_vsync == nullptr) {
 			// If no vsync data available, just sleep for roughly a vsync period.
 			// We'll get synced back up later.
-			std::this_thread::sleep_for(vsync_period);
+			std::this_thread::sleep_for(VSYNC_PERIOD);
 			return;
 		}
 
@@ -71,7 +76,7 @@ public:
 		}
 #endif
 		
-		bool hasRenderedThisInterval = (now - lastFrameTime) < vsync_period;
+		bool hasRenderedThisInterval = (now - lastFrameTime) < VSYNC_PERIOD;
 
 		// If less than one frame interval has passed since we last rendered...
 		if (hasRenderedThisInterval)
@@ -84,7 +89,7 @@ public:
 			// by a vsync period, so it's always in the future.
 			while(wait_time < now)
 			{
-				wait_time += vsync_period;
+				wait_time += VSYNC_PERIOD;
 			}
 
 #ifndef NDEBUG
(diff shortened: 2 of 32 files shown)


runner/runner/util.py
```diff
@@ -57,6 +57,18 @@ def unflatten(it: Iterable[V]) -> Iterable[Iterable[V]]:
         yield (elem,)
 
 
+def flatten_maps_list(maps: List[Mapping[str, str]], key_prefix: Optional[str] = "") -> Mapping[str, str]:
+    """
+    Take a list of maps, returning a single map with every mapping found with
+    an each key name updated with the (optionally) provided key prefix
+    """
+    flat_map = dict()
```
charmoniumQ, line 65:
```suggestion
    flat_map: Mapping[str, str] = dict()
```

Mypy will probably get mad at you because it can't deduce the type here.

runner/runner/util.py
```diff
@@ -57,6 +57,18 @@ def unflatten(it: Iterable[V]) -> Iterable[Iterable[V]]:
         yield (elem,)
 
 
+def flatten_maps_list(maps: List[Mapping[str, str]], key_prefix: Optional[str] = "") -> Mapping[str, str]:
```
charmoniumQ, line 60:
```suggestion
def flatten_maps_list(maps: List[Mapping[str, V]], key_prefix: str = "") -> Mapping[str, V]:
```
`Optional` means can be `None`, not can be omitted from the arg-list. Mypy will figure out that it can be omitted from the arg-list because there's a default.

runner/runner/main.py
```diff
@@ -149,6 +153,7 @@ def load_tests(config: Mapping[str, Any]) -> None:
     subprocess_run(
         ["xvfb-run", str(runtime_exe_path), *map(str, plugin_paths)],
         env_override=dict(ILLIXR_DATA=str(data_path), ILLIXR_DEMO_DATA=str(demo_data_path), ILLIXR_RUN_DURATION="10"),
+        env_constants=consts_map,
```
charmoniumQ, line 156:
```suggestion
        env_override=dict(ILLIXR_DATA=str(data_path), ILLIXR_DEMO_DATA=str(demo_data_path), ILLIXR_RUN_DURATION="10", **consts_map),
```
I don't think this merits being a new argument because it is specific to this use case.

runner/runner/util.py
```diff
@@ -164,6 +177,8 @@ def subprocess_run(
     cwd = (cwd if isinstance(cwd, Path) else Path(cwd)) if cwd is not None else Path()
     if env_override:
         env.update(env_override)
+    if env_constants:
+        env.update(env_constants)
```
charmoniumQ, line 181:
```suggestion
```
See comment on the call-site of this function.

runner/runner/main.py
charmoniumQ: Likewise for the other invocations.

charmoniumQ (commented):
Thanks for having good comments too.

We currently call `getenv` a lot for `ILLIXR_DATA` and `ILLIXR_DEMO_DATA` (horribly named). We could use this PR (or another; It doesn't matter to me) to move those to your new configuration system. The only catch is that these have to be `pathify`ed by the runner instead of directly passed in from the config (grep for `data_path` or `demo_data_path` in the runner to see what I mean). So, I was thinking of using this as an opportunity to correct some bad var names and also calling `pathify` on any config var name that ends with `_path`.

We should also replace `ILLIXR_RUN_DURATION` with your new conf system (this PR or a different one).

charmoniumQ:
In general, it can be useful to run:
```
isort *.py
black --line-length 90 *.py
mypy --strict --ignore-missing-imports *.py
```
but we have no formalized process for this (no requirement).

runner/runner/util.py
```diff
@@ -228,26 +249,26 @@ def worker(chunk: List[T], results: queue.Queue[V]) -> None:
     for thread in threads:
         thread.start()
 
-    def get_results(results: queue.Queue[V]) -> Iterable[V]:
+    def get_results(
```
charmoniumQ, line 252: You beat me to the Mypy fixes here ;)

runner/runner/main.py
```diff
@@ -128,14 +130,17 @@ def load_native(config: Mapping[str, Any]) -> None:
     )
     subprocess_run(
         command_lst_sbst,
-        env_override=dict(ILLIXR_DATA=str(data_path), ILLIXR_DEMO_DATA=str(demo_data_path), KIMERA_ROOT=config["loader"]["kimera_path"]),
+        env_override=dict(
+            KIMERA_ROOT=config["loader"]["kimera_path"], **consts_map_pathified,
```
charmoniumQ, line 134: Can we factor `kimera_path` into a global module def too?

charmoniumQ (approved):

runner/runner/main.py
e3m3: Definitely! I wasn't sure what the desired behavior was for kimera, since the path and the plugin is/was not used in all of the yaml configurations. Should it be a default global definition?

charmoniumQ: Actually, it's probably better to handle this a different way.

e3m3: Got it. Will discuss more about `__FILE__` then

e3m3:
Getting the segfault in CI from the logger issue

charmoniumQ (approved):
I Agree: CI is failing due to pre-existing issue.

You should force this to merge when you get a second approval. Lmk if you don't have the "force to merge anyway" button.

runner/runner/main.py
apache8080: Did you guys file an issue for this?

apache8080 (approved):
LGTM

timewarp_gl/plugin.cpp
```diff
@@ -59,18 +67,18 @@ class timewarp_gl : public threadloop {
 	const std::shared_ptr<switchboard> sb;
 	const std::shared_ptr<pose_prediction> pp;
 
-	static constexpr int   SCREEN_WIDTH    = 550*2;
-	static constexpr int   SCREEN_HEIGHT   = 320*2;
+	//static constexpr int   SCREEN_WIDTH    = 550*2;
```
mhuzai, line 70: Remove old code here and elsewhere.

runtime/runtime_impl.hpp
```diff
@@ -4,20 +4,27 @@
...
+
 class runtime_impl : public runtime {
 public:
 	runtime_impl(GLXContext appGLCtx) {
 		pb.register_impl<record_logger>(std::make_shared<sqlite_record_logger>());
 		pb.register_impl<gen_guid>(std::make_shared<gen_guid>());
 		pb.register_impl<switchboard>(create_switchboard(&pb));
-		pb.register_impl<xlib_gl_extended_window>(std::make_shared<xlib_gl_extended_window>(448*2, 320*2, appGLCtx));
+		pb.register_impl<xlib_gl_extended_window>(std::make_shared<xlib_gl_extended_window>(GLX_FB_WIDTH, GLX_FB_HEIGHT, appGLCtx));
+		//pb.register_impl<xlib_gl_extended_window>(std::make_shared<xlib_gl_extended_window>(448*2, 320*2, appGLCtx));
```
mhuzai, line 27: Remove.

gldemo/plugin.cpp
```diff
@@ -13,14 +13,19 @@
...
-static constexpr int   EYE_TEXTURE_HEIGHT  = 1024;
+static const std::string OBJ_DIR { ILLIXR::DEMO_OBJ_PATH };
 
-static constexpr std::chrono::nanoseconds vsync_period {std::size_t(NANO_SEC/60)};
+static const int EYE_TEXTURE_WIDTH   { ILLIXR::FB_WIDTH  };
+static const int EYE_TEXTURE_HEIGHT  { ILLIXR::FB_HEIGHT };
+
+static const double DISPLAY_REFRESH_RATE { ILLIXR::REFRESH_RATE };
+static const std::chrono::nanoseconds VSYNC_PERIOD { static_cast<size_t>(NANO_SEC/DISPLAY_REFRESH_RATE) };
+//static constexpr std::chrono::nanoseconds VSYNC_PERIOD {std::size_t(NANO_SEC/60)};
```
mhuzai, line 28: Remove.

mhuzai (commented):
Left a few small comments but otherwise looks fantastic.

therishidesai (dismissed):

mhuzai (dismissed):

mhuzai:
CI is failing with `Illegal instruction (core dumped)`.

common/gl_util/lib/tiny_obj_loader.h
```diff
@@ -2168,6 +2168,7 @@ bool LoadObj(attrib_t *attrib, std::vector<shape_t> *shapes,
   std::stringstream errss;
 
   std::ifstream ifs(filename);
+
```
JeffreyZh4ng, line 2171: Remove file from commit

JeffreyZh4ng (commented):
I thought the code was clean and easy to understand. Most of the other changed files were just changing the call sites of where the (now configurable) params were used.

charmoniumQ:
Let's also specify whether to use ZED's parameters in OpenVINS ([here](https://github.com/ILLIXR/open_vins/blob/issue-208-cpu-timer3/ov_msckf/src/slam2.cpp#L23)) from this module.

In the long run, we would like to read these params from a file, and that file should be specified by a runtime parameter.

astro-friedel:
Much of this has been previously addressed or is addressed in the new cmake build system.

PR #204, opened 2020-12-03
Fix CI

charmoniumQ:
[Master](https://github.com/ILLIXR/ILLIXR/runs/1416725950) has a green checkmark despite:
```
Please define KIMERA_ROOT
Aborted (core dumped)
```

This is because we do not pass `check=True` the `subprocess.run`, so the return-code of ILLIXR was never checked by the runner.

At first CI failed when it should have succeeded, but in the immortalized words of Batman, eventually you live long enough to see yourself become the hero or die young.


charmoniumQ:
This should be accepted despite the failing check, because the previous system would not report a genuine failure and this one will.

```
Joined timewarp_gl
Joined audio_encoding
Joined audio_decoding
Joined gldemo
I1203 14:07:22.480195  1317 Pipeline.cpp:261] Pipeline destructor called.
I1203 14:07:22.480234  1317 Pipeline.cpp:423] Shutting down VIO pipeline.
I1203 14:07:22.480321  1317 MesherModule.cpp:65] Shutting down queues for: Mesher
I1203 14:07:22.480371  1317 Pipeline.cpp:443] VIO Pipeline's threads shutdown successfully.
VIO Pipeline successful shutdown.
I1203 14:07:22.481595  1317 VioBackEnd.h:88] Backend destructor called.
I1203 14:07:22.483816  1317 StereoVisionFrontEnd.h:64] StereoVisionFrontEnd destructor called.
Segmentation fault (core dumped)
```

e3m3 (approved):

mhuzai:
The changes look good. The failure is because of the sqlite issue. We can merge this in, but then we would have a broken master until sqlite gets fixed.

Idea: switch to stdout_logger or noop_logger until sqlite gets fixed? Thoughts?

charmoniumQ:
> We can merge this in, but then we would have a broken master until sqlite gets fixed.
We already have a broken master until SQLite gets fixed.

I'll try moving to `noop_record_logger` (I agree this is a reasonable fix until SQLite gets fixed).

charmoniumQ:
After switching to the `noop_record_logger`, it's still failing.

The segfault might be in some destructor, and that destructor is preventing the SQLite from logging, but SQLite is a symptom not a cause, so removing it does not remove the segfault.

mhuzai:
> 
> 
> After switching to the `noop_record_logger`, it's still failing.
> 
> The segfault might be in some destructor, and that destructor is preventing the SQLite from logging, but SQLite is a symptom not a cause, so removing it does not remove the segfault.

That's very interesting. At least helps rule out sqlite as the offender. But also makes it more cryptic what the error is.

Anyway, this is good to go.

charmoniumQ:
> Anyway, should we revert back to sqlite logger before merging?

Agree; done.

mhuzai (approved):

e3m3 (approved):

PR #209, opened 2021-01-10
Implement CPU Timer v3

charmoniumQ:
This breaks the existing graphing infrastructure (aka "results repo"), so this will remain WIP until complementary changes have taken place there.


charmoniumQ:
Also see [related OpenVINS branch](https://github.com/ILLIXR/open_vins/tree/issue-208-cpu-timer3).

PR #214, opened 2021-01-10
Issue 75 mutex

### LINKED ISSUE #75: Replace `std::mutex` with `std::shared_mutex`

charmoniumQ:
Fixes #75, supersedes #158.

### DIFF
--- a/runtime/sqlite_record_logger.hpp
+++ b/runtime/sqlite_record_logger.hpp
@@ -5,6 +5,7 @@
 #include <chrono>
 #include <atomic>
 #include <thread>
+#include <shared_mutex>
 #include <experimental/filesystem>
 #include "concurrentqueue/blockingconcurrentqueue.hpp"
 #include "sqlite3pp/sqlite3pp.hpp"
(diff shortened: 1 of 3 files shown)


e3m3 (approved):
Lgtm

runtime/sqlite_record_logger.hpp
```diff
@@ -176,14 +177,16 @@ class sqlite_record_logger : public record_logger {
...
+		{
+			const std::shared_lock<std::shared_mutex> lock {_m_registry_lock};
+			auto result = registered_tables.find(rh.get_id());
+			if (result != registered_tables.cend()) {
+				return result->second;
+			}
 		}
+		const std::unique_lock<std::shared_mutex> lock{_m_registry_lock};
+		auto pair = registered_tables.try_emplace(rh.get_id(), rh);
+		return pair.first->second;
```
mhuzai, line 189: Just to make sure, this locking is required to support the case where two different SB threads may try to log the same record type at the same time? Is this use case being exercised in our code at the moment?

mhuzai (approved):

runtime/sqlite_record_logger.hpp
charmoniumQ: Yeah; All switchboard threads submit `switchboard_callback` records to the same table. Only one of them should do the SQL `CREATE` incantation.

The SQL `INSERT` incantation is already serialized because all producers put in the same queue, and only one thread dequeues from that, which is responsible for actually doing the insertion,

mhuzai (approved):

e3m3 (approved):

PR #219, opened 2021-01-20
Clear errno after hazard and check errno for faults

### LINKED ISSUE #212: Figure out why `errno` is set
`errno` seems to be set to non-zero at startup. Someone must be doing a dangerous operation, and not checking `errno`. We need to know who.

Could be related to #173 .

charmoniumQ:
- Apparently, `XOpenDisplay` can set `errno` even if there is no error ([source](https://cboard.cprogramming.com/linux-programming/119957-xlib-perversity.html)). The true indication of error is whether its return is non-null. This theoretical possibility was actually happening on two systems I tested with. Therefore, we need to clear `errno`.
- To catch this in the future, I will `assert(errno == 0)`. This is not meant to be a "definitive" check. It could be set despite no error (as in the previous case), and it could be unset despite an error (if some intermediate call clears it). For a definitive check, we would need to check every call that could set `errno`. For now, I am satisfied with "partial" `errno` checking for now.

Fixes #212.

### DIFF
--- a/runtime/runtime_impl.hpp
+++ b/runtime/runtime_impl.hpp
@@ -73,7 +73,18 @@ public:
 			std::cerr << "You didn't call stop() before destructing this plugin." << std::endl;
 			abort();
 		}
-		assert(errno == 0);
+		assert(errno == 0 && "errno was set during run. Maybe spurious?");
+		/*
+		  Note that this assertion can have false positives AND false negatives!
+		  - False positive because the contract of some functions specifies that errno is only meaningful if the return code was an error [1].
+		    - We will try to mitigate this by clearing errno on success in ILLIXR.
+		  - False negative if some intervening call clears errno.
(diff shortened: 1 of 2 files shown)


runtime/runtime_impl.hpp
```diff
@@ -71,6 +73,7 @@ class runtime_impl : public runtime {
 			std::cerr << "You didn't call stop() before destructing this plugin." << std::endl;
 			abort();
 		}
+		assert(errno == 0);
```
mhuzai, line 76: Can you please add the "this is not a definitive check" statement from the pull request in a comment here for clarity?

mhuzai (commented):
Holy.

Code looks good. Please just add requested comment.

mhuzai (approved):

e3m3 (approved):

Boyuan-Tian (approved):

Issue #220, opened 2021-01-24
Base threadloop off of managed_thread


charmoniumQ:
All threads that run a tight-loop and check for termination in ILLIXR should have a common base class.

This eases the implementation of other features:
- #208: Logging code can go in the common base class.
- #217: "pausing" would be written once in the common base class and used everywhere.

This also minimizes the possibility of error, because `managed_thread` automatically stops itself in its destructor.

Issue #211, opened 2021-01-10
Consider logging over Switchboard


charmoniumQ:
Consider sending structured logs to the database over switchboard instead of over a custom interthread queue.

How could we log performance data inside Switchboard, if logging uses Switchboard?

Would we need the advanced Switchbaord interface so that it can process post-realtime?

charmoniumQ:
Relevant parts of `project-scheduling`:
- [runtime/main.cpp:49](https://github.com/ILLIXR/ILLIXR/blob/91c6ff35db9fc48a6407fea2131aa35429383095/runtime/main.cpp#L49)
- [runtime/frame_logger2.hpp](https://github.com/ILLIXR/ILLIXR/blob/91c6ff35db9fc48a6407fea2131aa35429383095/runtime/frame_logger2.hpp)

PR #221, opened 2021-01-24
Fix timing infrastructure

### LINKED ISSUE #208: Fix CPU timing infrastructure
- The existing infrastructure uses a different collection method for ILLIXR-native components, external components such as OpenVINS, and libraries such as OpenCV.
- The existing infrastructure is not thread safe: lines could get spliced in the log.
- The existing infrastructure is limited to capturing at a specific level: a plugin's `_p_one_iteration` or switchboard callback. This necessitates a lot of complex, hardcoded logic in the analysis scripts. This complexity was error-prone (this is why periods were wrong in OpenVINS when I was showing it during the scheduling meeting).
  - It requires hardcoded logic to split `offline_imu_cam` and `OpenVINS` into the times it had camera data vs just IMU data.
  - It requires hardcoded logic to sum OpenVINS's component threads (durations should be summed, but what about start times?).
  - It requires hardcoded logic to extract the logging overhead time.
- The existing infrastructure cannot distinguish when component A calls a library function that starts a thread and when B calls the same function. Since it starts a thread, its CPU time won't be accounted in A and B. This gives us trouble in the existing infrastructure that OpenVINS and the camera loader both called OpenCV.

charmoniumQ:
Closes #208. Dependent on pending PRs for #220 and [open_vins #2](https://github.com/ILLIXR/open_vins/issues/2).

I have broken off the CPU timer into a separate repo [here](https://github.com/charmoniumQ/cpu_timer) so that it can be reused between projects.

It will be the responsibility of a future PR to log that data to disk (issue #211).

This PR removes the old SQLite logging infrastructure. Logs that use or add-to the stack-frame's context should be added into the "extra info" of the CPU timer. Logs that do not should be converted to Switchboard topics and logged into a bag.

This PR is for the interface of the CPU timer. The reviewer should verify:
- The interface is easy to use.
- The documentation is sufficient enough for others to use the CPU timer.
- No spurious changes were erroneously committed.
- ILLIXR still runs at a normal speed and prints something like `Got 75493 finished frames` a few times (once for each thread) at the finish.


astro-friedel:
Superceded by PR 383

Issue #225, opened 2021-02-05
Errno and error-checking


charmoniumQ:
Same as #212, but electric boogaloo.

We should seek to clear errno. While checking errno can have false positives, this seeks to reduce those false positives.

```
maybe_set_errno_spuriously();
errno = 0;
```

We should check errno before clearing errno, since `errno = 0` could mask a prior error. This reduces the probability of false negatives.

```
assert(errno == 0);
maybe_set_errno_spuriously();
errno = 0;
```

Usually, when errno is set in a false positive, we should be checking for errors in other ways.

```
assert(errno == 0);
int rc = maybe_set_errno_spuriously();
if (rc) { abort(); }
errno = 0;
```

Issue #213, opened 2021-01-10
Preserve partial progress in our `Makefile`


charmoniumQ:
We compile-and-link in one step like so:

```
${CXX} blah.cpp blah2.cpp ...
```

This means that if _any_ of the `.cpp` files change, we recompile all of them. Instead, we should compile and link in two steps.

```
# For each CPP file
${CXX} blah.cpp -o blah.dbg.o

${CXX} blah.dbg.o blah2.dbg.o ... -o plugin.dbg.so
```

This might require converting some things in `common/` from `.hpp` to `.cpp`.

Also see why OpenVINS recompilation takes so long.

charmoniumQ:
See #42.

astro-friedel:
The new CMake build system is much smarter about only building file which have changed and linking is a separate step as well.

Issue #54, opened 2020-07-26
Use a C++ linter


charmoniumQ:
- Possibly OCLint
- Possibly [cppclean](https://code.google.com/archive/p/cppclean/)
- Should detect unused `#include`

astro-friedel:
clang-tidy was implemented a while ago.

Issue #44, opened 2020-07-26
Add human-readable logging infrastructure


charmoniumQ:
Our `record_logger` and the CPU timers in #209/#211 are good for computer-readable telemetry. However, we need an infrastructure for data intended to get to a user, such as warnings, error descriptions, and debugging messages.

Perhaps, we should use [glog](https://github.com/google/glog).

- By default, ILLIXR should not print anything except for really critical stuff ([why this is considered "good UNIX style"](http://www.catb.org/~esr/writings/taoup/html/ch11s09.html))
  - Logs in a loop can use something like glog's `LOG_EVERY_N`.
- Many of the `assert(thing && msg)` can be replaced with glog's `CHECK(thing) << msg`.
  - This permits the message to be a dynamically generated string, like `CHECK(thing) << getErrorMessage(msg)` which would not work with a raw `assert`.
  - `CHECK` is *not* suppressed by `NDEBUG`, which has caused us bugs in the past (and pesky `[[maybe_unused]]` annotations).
    - One can still explicitly reserve a check for debug builds by surrounding the `CHECK` with `#ifndef NDEBUG`, as we do already in many places (e.g. `common/record_logger.hpp`).
- Glog can dump the stacktrace on segfault.
  - We do already have `catchsegv`. There's not a whole lot of benefit to switching over to glog for that.
- Glog embeds the filename and line number in all logs.
- Perhaps we want to "selectively enable" debug flags for different plugins, as in LLVM. This is the only thing not readily avilable in glog.

PR #227, opened 2021-02-05
Issue 225 error checking

JeffreyZh4ng:
https://github.com/ILLIXR/ILLIXR/issues/225


configs/core_plugins.yaml
```diff
@@ -3,7 +3,7 @@ plugin_group:
   - name: audio
     path:
       git_repo: https://github.com/ILLIXR/audio_pipeline.git
-      version: 3433bb452b2ec661c9d3ef65d9cf3a2805e94cdc
+      version: 846196a4ebbf7be084b6e5f1d20f709d26bb75f4
```
e3m3, line 6:
```suggestion
      version: eb0b3ded1918567b4ff5bac1078e6bdd4a8baee1
```

configs/rt_slam_plugins.yaml
```diff
@@ -4,10 +4,10 @@ plugin_group:
   - name: Kimera-VIO/
     path:
       git_repo: https://github.com/ILLIXR/Kimera-VIO
-      version: 6ea99d33fc2435a8c514049416dcb2f8a0e31692
-  #- name: OpenVINS
-  #  path:
-  #    git_repo: https://github.com/ILLIXR/open_vins.git
-  #    version: c6a4f8835851d21f87add38260e73df63c6354a1
+      version: 3f69ea19e063b0b399cba005067b875bd7373deb
```
e3m3:
```suggestion
      version: 648ac892543701abd78a6f4bc56a6b6001f8413f
```

e3m3 (changes requested):
Updated the commit numbers to reflect updated branches for Kimera-VIO and audio_pipeline

configs/rt_slam_plugins.yaml
charmoniumQ: I suggest we use a better strategy for managing these commit references.

ILLIXR/ILLIXR (what I'll call "main repo") needs to refer to ILLIXR/audio_pipeline and friends (what I'll call "subrepos"). Currently, we do this by pinning an exact version (commit hash). If the subrepo updates, the main repo does not automatically get that change. This is good for long-term stability, bad for rapid development. If you do want those changes, one has to do these small edits. However, these tiny changes are easily overlooked (as they were here) and can be easily wrong.

Referring to "master" of the subrepos is on the other end of the spectrum. It's great for rapid development and bad for stability. We know that branches of main repo can fall months out-of-date. If the main repo references master of the subrepo, the subrepos will have to support a wide range of main repo versions. If a subrepo updates master, it could break old branches of the main repo, which get the update automatically.

The in-between solution I propose is managing these by semver. Whenever master in a subrepo updates, its maintainer may choose to move the tag x.y to point to the new commit or create a new tag, x.(y+1) or (x+1).0. This way, we explicitly say when we are breaking backwards compatibility (when the minor or major version changes), and non-breaking changes can be pulled in for free, enabling rapid development without compromising deterministic behavior of old branches.

charmoniumQ: image

e3m3: I agree that we should move to a tag-based system like you suggest. @mhuzai Should the versioning system we use match the milestone release versions to avoid confusion? In that case, we would start from a version `2.X` tag?

mhuzai: Well, the components are independent of each other, so they should be able to bump their versions independently. We can tag the last standalone commit 1.0, the current latest commit 2.0, and then bump versions in each subrepo as required.

mhuzai:
Please remove duplicate commits from other branches.

e3m3 (approved):

charmoniumQ (approved):

timewarp_gl/plugin.cpp
```diff
@@ -620,10 +662,43 @@ class timewarp_gl : public threadloop {
...
+            std::exit(1);
+		}
+
+        RAC_ERRNO_MSG("timewarp_gl before closing X");
  		glXDestroyContext(xwin->dpy, xwin->glc);
  		XDestroyWindow(xwin->dpy, xwin->win);
  		XCloseDisplay(xwin->dpy);
+        RAC_ERRNO_MSG("timewarp_gl after closing X");
+
+	    if (distortion_indices != nullptr) {
```
charmoniumQ, line 678: `free` is good, but `std::vector` is better for these variables.

I think we want them to be a vector of vectors, one for each eye. This eliminates `vec[eye*NUM_ELEM + elem]`.

runtime/sqlite_record_logger.hpp
```diff
@@ -115,33 +130,35 @@ class sqlite_thread {
 			// This currently has to be 'reinterpreted' for every iteration.
 			sqlite3pp::command cmd{db, insert_str.c_str()};
 			const record& r = record_batch[i];
-			for (unsigned i = 0; i < rh.get_columns(); ++i) {
+			for (unsigned int j = 0; j < rh.get_columns(); ++j) {
```
charmoniumQ, line 133: Yup. Good call.

offline_imu_cam/plugin.cpp
```diff
@@ -63,20 +65,42 @@ class offline_imu_cam : public ILLIXR::threadloop {
...
+                default:
+                    break;
+            }
+            switch (num_ch1) {
+                case 3:
+                    cv::cvtColor(*cam1.value(), *cam1.value(), cv::COLOR_BGR2GRAY);
+                case 1:
+                default:
+                    break;
+            }
```
charmoniumQ, line 101: IMO unnecessary as `offline_imu_cam` is tightly-coupled to `lazy_load` already. It can be part of the "contract" that lazy loud emits grayscale images. I would rather just `assert` that this contract is met in `offline_imu_cam`.

```
assert(blah->channels() == 1 && "Got a color image where grayscale is expected");
```

gldemo/plugin.cpp
```diff
@@ -363,20 +398,26 @@ class gldemo : public threadloop {
 		// Load/initialize the demo scene.
 
 		char* obj_dir = std::getenv("ILLIXR_DEMO_DATA");
-		if(obj_dir == NULL) {
+		if (obj_dir == nullptr) {
 			std::cerr << "Please define ILLIXR_DEMO_DATA." << std::endl;
-			abort();
+            std::exit(1);
```
charmoniumQ, line 403: At faults, I think `abort` is better because it fires a `SIGABRT` signal, which can be caught by a debugger. `std::exit` and `std::terminate` cannot, AFAIK. I suppose this is why `assert` sends a `SIGABRT`.

The question is, should destructors be run if the exit is due to a fatal error? I think the answer is no. In the case of a fatal error, the output is likely "tainted" by the error. In the _special case_ that the operator does want it, they can replace that particular `abort` with an `std::exit`, but I contend this should not be the default,

gldemo/plugin.cpp
```diff
@@ -320,20 +348,23 @@ class gldemo : public threadloop {
...
 	// it will be replaced by a real, Monado-interfaced application.
 	virtual void start() override {
-		glXMakeCurrent(xwin->dpy, xwin->win, xwin->glc);
+		assert(errno == 0 && "Errno should not be set at start of gldemo start function");
 
-		// Initialize the GLFW library, still need it to get time
-		if(!glfwInit()){
-			printf("Failed to initialize glfw\n");
+		if (!glXMakeCurrent(xwin->dpy, xwin->win, xwin->glc)) {
+			std::cerr << "glXMakeCurrent returned false in " << __FILE__ << ":" << __LINE__ << std::endl;
```
charmoniumQ, line 354: I think it's best to `assert(0 && "something bad happened")`. It's less typing anyway.

Unfortunately, `assert(glXstuff() ...)` (eliding the if-statement) won't work because it won't run the expression when `profile = opt`.

gldemo/plugin.cpp
```diff
@@ -65,8 +65,8 @@ class gldemo : public threadloop {
 
 #ifndef NDEBUG
 		if (log_count > LOG_PERIOD) {
-			std::chrono::duration<double, std::milli> vsync_in = *next_vsync - now;
-			printf("\033[1;32m[GL DEMO APP]\033[0m First vsync is in %4fms\n", vsync_in.count());
+			const double vsync_in = std::chrono::duration_cast<std::chrono::milliseconds>(*next_vsync - now).count();
```
charmoniumQ, line 68: :+1:

debugview/plugin.cpp
```diff
@@ -47,6 +48,16 @@ Eigen::Matrix4f lookAt(Eigen::Vector3f eye, Eigen::Vector3f target, Eigen::Vecto
 
 }
 
+
+/**
+ * @brief Callback function to handle glfw errors
+ */
+static void glfw_error_callback(int error, const char* description) {
+    std::cerr << "|| glfw error_callback: " << error << std::endl
+              << "|> " << description << std::endl;
```
charmoniumQ, line 57: Shall we exit/abort in this case?

common/shader_util.hpp
```diff
@@ -19,13 +20,14 @@ static void GLAPIENTRY
...
     }
-	fprintf( stderr, "GL CALLBACK: %s type = 0x%x, severity = 0x%x, message = %s\n",
-			( type == GL_DEBUG_TYPE_ERROR ? "** GL ERROR **" : "" ),
-				type, severity, message );
+    std::cerr << "GL CALLBACK: " << (type == GL_DEBUG_TYPE_ERROR ? "** GL ERROR **" : "")
+              << " type = 0x" << std::hex << type << std::dec
+              << ", severity = 0x" << std::hex << severity << std::dec
+              << ", message = " << message
+              << std::endl;
 #endif
```
charmoniumQ, line 31: Shall we exit/abort in this case?

common/extended_window.hpp
```diff
@@ -6,42 +6,44 @@
...
+                std::cerr << "Cannot connect to X server" << std::endl;
+                std::exit(1);
             } else {
 				// Apparently, XOpenDisplay's _true_ error indication is whether dpy is nullptr.
 				// https://cboard.cprogramming.com/linux-programming/119957-xlib-perversity.html
 				// if (errno != 0) {
 				// 	std::cerr << "XOpenDisplay succeeded, but errno = " << errno << "; This is benign, so I'm clearing it now.\n";
 				// }
-				errno = 0;
+				RAC_ERRNO_MSG("extended_window after XOpenDisplay : value returned not null");
```
charmoniumQ, line 43: The point of this is that non-zero `errno` here doesn't actually denote an erroneous condition.

Any output caused by this is noise. Let's go back to `errno = 0`.

offline_imu_cam/plugin.cpp
e3m3: In previous testing, calling `cv:cvtColor` was an issue because the input was not color. In testing this branch, I had the opposite issue- removing the call was an issue because the input was indeed in color. Did the behavior of `lazy_load` change recently?

gldemo/plugin.cpp
e3m3: `abort` does not allow the C++ runtime to perform cleanup:
>  The difference between exit and abort is that exit allows the C++ run-time termination processing to take place (global object destructors get called), but abort terminates the program immediately. The abort function bypasses the normal destruction process for initialized global static objects. It also bypasses any special processing that was specified using the atexit function.
[Source](https://docs.microsoft.com/en-us/cpp/cpp/program-termination?view=msvc-160). It seems to me that `abort` could be useful for a debugging environment, though. Perhaps we can use a small utility that switches the `std::exit` to `abort` if `#ifdndef NDEBUG`?

e3m3: In terms of the destructors, we can be holding onto stateful resources (e.g., file handles) which should be closed normally if possible to avoid odd side-effects.

e3m3: That is, I think debugging via `SIGABRT` is the special case, given that the user will need to resolve the failure.

gldemo/plugin.cpp
e3m3: Agreed. The value can be stored temporarily.

common/extended_window.hpp
e3m3: Agreed the output is noise, but currently, `RAC_ERRNO` and `RAC_ERRNO_MSG` only print and clear when `NDEBUG` is on. Errnos throughout execution during a run with `opt` are ignored. Clearing errno here would be the exception.

common/global_module_defs.hpp
```diff
@@ -4,6 +4,47 @@ This is the file where default values are defined
...
+
+
+/**
+ * @brief Support function to debug.
+ *
+ * If errno is set, this function will report errno's value and the calling context.
+ * It will subsequently clear errno (reset value to 0).
+ * Otherwise, this function does nothing.
+ */
+inline void report_and_clear_errno(
```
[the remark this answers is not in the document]
e3m3: The meaning for `inline` has somewhat changed to mean that multiple definitions are allowed, but `static` probably results in the same thing but with a cleaner output.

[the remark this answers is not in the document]
e3m3: The meaning for `inline` has somewhat changed to mean that multiple definitions are allowed, but `static` probably results in the same thing but with a cleaner output.

common/extended_window.hpp
charmoniumQ: Debug mode will be active when someone is hunting for the source of a true bug. Getting a print for a not-genuine-error in such case would be misleading. The reader has to read the message and understand the source to determine that it is not genuine.

offline_imu_cam/plugin.cpp
charmoniumQ: No, I don't think it changed recently.

Since the component is only used here, the behavior "should" be whatever makes it easier for `offline_imu_cam` and its downstream, `open_vins`. I believe loading grayscale images from `cv::imread` is the right way to go.

common/common.mk
```diff
@@ -3,7 +3,7 @@
 CXX := clang++-10
 STDCXX ?= c++17
 CFLAGS := $(CFLAGS) -DGLSL_VERSION='"330 core"'
-DBG_FLAGS ?= -Og -g -Wall -Wextra -Werror
+DBG_FLAGS ?= -Og -g -Wall -Wextra -Werror -rdynamic
```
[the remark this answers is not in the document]
e3m3: Yep, I think `catchsegv` uses `backtrace`.

charmoniumQ (dismissed):

gldemo/plugin.cpp
```diff
@@ -320,20 +346,21 @@ class gldemo : public threadloop {
...
-		if(!glfwInit()){
-			printf("Failed to initialize glfw\n");
-		}
 		// Init and verify GLEW
-		if(glewInit()){
-			printf("Failed to init GLEW\n");
-			exit(0);
+		const GLenum glew_err = glewInit();
+		if (glew_err != GLEW_OK) {
+            ILLIXR::abort("Failed to init GLEW");
```
charmoniumQ, line 358: std::cerr << "GLEW error: " << glewGetErrorString(err) << std::endl;

common/shader_util.hpp
```diff
@@ -98,8 +106,9 @@ static GLuint init_and_link (const char* vertex_shader, const char* fragment_sha
 		std::cout << error_msg;
     }
 
-    if(glGetError()){
-        printf("initGL, error at end of initGL");
+    if (glGetError()) {
+        std::cerr << "initGL, error at end of initGL" << std::endl;
+        std::exit(1);
```
charmoniumQ, line 111: I believe these `std::exit`s should be `illixr::abort`s.

offline_imu_cam/data_loading.hpp
```diff
@@ -23,7 +23,7 @@ class lazy_load_image {
 		: _m_path(path)
 	{ }
 	std::unique_ptr<cv::Mat> load() const {
-		auto img = std::unique_ptr<cv::Mat>{new cv::Mat{cv::imread(_m_path, cv::IMREAD_COLOR)}};
+		auto img = std::unique_ptr<cv::Mat>{new cv::Mat{cv::imread(_m_path, cv::IMREAD_GRAYSCALE)}};
 		/* TODO: make this load in grayscale */
```
charmoniumQ, line 27: Remove `TODO`

charmoniumQ:
Also note: #44 relates to using a "logging infrastructure for human-readable messages" such as [glog](https://github.com/google/glog). Glog has `PLOG` which prints your message, the line number, and the human-readable message associated with errno.

I vote to accept this now, but have an eye towards upgrading these facilities in #44. We could upgrade the implementation of your error-reporting macros without changing their call-sites.

offline_imu_cam/data_loading.hpp
```diff
@@ -24,7 +24,6 @@ class lazy_load_image {
 	{ }
 	std::unique_ptr<cv::Mat> load() const {
 		auto img = std::unique_ptr<cv::Mat>{new cv::Mat{cv::imread(_m_path, cv::IMREAD_GRAYSCALE)}};
-		/* TODO: make this load in grayscale */
 		assert(!img->empty());
 
 		// Sam's note: I am moving cvtColor into slam2
```
charmoniumQ, line 29: One more request :laughing: :
Delete this comment and the commented code; it is now out-of-date.

e3m3: You got it

charmoniumQ (dismissed):

timewarp_gl/plugin.cpp
```diff
@@ -433,69 +438,109 @@ class timewarp_gl : public threadloop {
...
+		glBufferData(GL_ELEMENT_ARRAY_BUFFER, num_distortion_indices * sizeof(GLuint), distortion_indices_data, GL_STATIC_DRAW);
 
-                // Config PBO for texture image collection
-                glGenBuffers(1, &PBO_buffer);
-                glBindBuffer(GL_PIXEL_PACK_BUFFER, PBO_buffer);
-                glBufferData(GL_PIXEL_PACK_BUFFER, SCREEN_WIDTH * SCREEN_HEIGHT * 3, 0, GL_STREAM_DRAW);
+        // Config PBO for texture image collection
+        glGenBuffers(1, &PBO_buffer);
+        glBindBuffer(GL_PIXEL_PACK_BUFFER, PBO_buffer);
+        glBufferData(GL_PIXEL_PACK_BUFFER, SCREEN_WIDTH * SCREEN_HEIGHT * 3, 0, GL_STREAM_DRAW);
```
mhuzai, line 506: Only do this when collecting data / offload is enabled.

timewarp_gl/plugin.cpp
```diff
@@ -433,69 +438,109 @@ class timewarp_gl : public threadloop {
...
 
 		glBindFramebuffer(GL_FRAMEBUFFER,0);
 		glViewport(0, 0, SCREEN_WIDTH, SCREEN_HEIGHT);
 		glClearColor(0, 0, 0, 0);
-    	glClear(GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT | GL_STENCIL_BUFFER_BIT);
+
+        glClear(GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT | GL_STENCIL_BUFFER_BIT);
+        RAC_ERRNO_MSG("timewarp_gl after glClear");
+        /// RAC is triggered by glClear consistently. Checking for errors here to be safe.
+        const GLenum glClearError = glGetError();
```
mhuzai, line 527: Why is this necessary? Won't GL errors get reported via the debug callback?

timewarp_gl/plugin.cpp
```diff
@@ -615,17 +655,22 @@ class timewarp_gl : public threadloop {
...
 
-		if(delta > vsync_period)
-		{
-			printf("\033[0;31m[TIMEWARP: CRITICAL]\033[0m Stale frame!\n");
+		if (time_since_render > vsync_period) {
+            std::cout << "\033[0;31m[TIMEWARP: CRITICAL]\033[0m Stale frame!" << std::endl;
 		}
+
+		// Even in debug builds, this is quite verbose
+		// std::cout << "\033[1;36m[TIMEWARP]\033[0m Warping from swap " << most_recent_frame->swap_indices[0] << std::endl;
```
mhuzai, line 673: Why is this necessary, given line 665?

timewarp_gl/plugin.cpp
```diff
@@ -677,28 +724,37 @@ class timewarp_gl : public threadloop {
...
+            const double latency_ptd = predict_to_display.count()/1e6;
+            const double latency_rtd = render_to_display.count()/1e6;
+            const time_type time_next_swap = GetNextSwapTimeEstimate();
+            const double timewarp_estimate = std::chrono::duration_cast<std::chrono::milliseconds>(time_next_swap - time_last_swap).count();
+
+            std::cout << "\033[1;36m[TIMEWARP]\033[0m Swap time: " << time_swap << "ms" << std::endl
+			          << "\033[1;36m[TIMEWARP]\033[0m Motion-to-display latency: " << latency_mtd << "ms" << std::endl
+			          << "\033[1;36m[TIMEWARP]\033[0m Prediction-to-display latency: " << latency_ptd << "ms" << std::endl
+		              << "\033[1;36m[TIMEWARP]\033[0m Render-to-display latency: " << latency_rtd << "ms" << std::endl
+			          << "Timewarp estimating: " << timewarp_estimate << "ms in the future" << std::endl;
```
mhuzai, line 738: This wording has always confused me. Can you make it something like "Next swap in: xyz ms"?

timewarp_gl/plugin.cpp
```diff
@@ -719,10 +775,16 @@ class timewarp_gl : public threadloop {
 
 	virtual ~timewarp_gl() override {
 		// TODO: Need to cleanup resources here!
```
mhuzai, line 777: Don't we take care of everything at this point? If so, remove this comment.

mhuzai (commented):
I am still not sure what the exact combo of asserts and RACs should be for errno checking. Let's discuss that via chat.

common/extended_window.hpp
```diff
@@ -6,31 +6,34 @@
...
+            std::cout << "Opening display" << std::endl;
 #endif
-            dpy = XOpenDisplay(NULL);
+            assert(errno == 0 && "Errno should not be set at start of xlib_gl_extended_window constructor");
+
+            dpy = XOpenDisplay(nullptr);
             if (!dpy) {
-                printf("\n\tcannot connect to X server\n\n");
-                exit(1);
+                ILLIXR::abort("Cannot connect to X server");
```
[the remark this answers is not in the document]
e3m3: Its better to be explicit, and avoid relying on `using namespace ...` elsewhere in the code. If the `using` line gets deleted or moved, the code breaks. Also, even with `using`, the code does not need any context for the reader to understand that it's not a call to `std::abort`. Also see [SO](https://stackoverflow.com/questions/1452721/why-is-using-namespace-std-considered-bad-practice) and this [article](https://www.geeksforgeeks.org/using-namespace-std-considered-bad-practice/).

common/extended_window.hpp
```diff
@@ -6,31 +6,34 @@
...
         xlib_gl_extended_window(int _width, int _height, GLXContext _shared_gl_context){
             width = _width;
             height = _height;
 
 #ifndef NDEBUG
-            printf("Opening display\n");
+            std::cout << "Opening display" << std::endl;
 #endif
-            dpy = XOpenDisplay(NULL);
+            assert(errno == 0 && "Errno should not be set at start of xlib_gl_extended_window constructor");
```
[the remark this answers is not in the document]
e3m3: Agreed. We will tackle this in #44.

> Also note: #44 relates to using a "logging infrastructure for human-readable messages" such as glog. Glog has PLOG which prints your message, the line number, and the human-readable message associated with errno.

> I vote to accept this now, but have an eye towards upgrading these facilities in #44. We could upgrade the implementation of your error-reporting macros without changing their call-sites.

configs/rt_slam_plugins.yaml
```diff
@@ -4,10 +4,10 @@ plugin_group:
   - name: Kimera-VIO/
     path:
       git_repo: https://github.com/ILLIXR/Kimera-VIO
-      version: 6ea99d33fc2435a8c514049416dcb2f8a0e31692
+      version: "2.1"
```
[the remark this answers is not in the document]
e3m3: It hasn't but this is an update to be merged that exists in the Kimera-VIO repo now to accompany this PR. Since the tag for the corresponding commit exists, we can test this PR in CI, and then merge the equivalent Kimera-VIO PR somewhat independently.

All versions will be easier to track once we integrate the refactored configurations for #231 and #54. Each external module has its own `.yaml` configuration specifying the tag so we can update versions in a consistent manner across plugins and configurations.

runtime/sqlite_record_logger.hpp
```diff
@@ -22,20 +25,30 @@ namespace ILLIXR {
...
 	std::string prep_insert_str() {
+		assert(errno == 0 && "Errno should not be set at start of prep_insert_str");
+
 		std::string drop_table_string = std::string{"DROP TABLE IF EXISTS "} + table_name + std::string{";"};
 		db.execute(drop_table_string.c_str());
+		RAC_ERRNO_MSG("sqlite_record_logger after drop table execute.");
 
 		std::string create_table_string = std::string{"CREATE TABLE "} + table_name + std::string{"("};
-		for (unsigned i = 0; i < rh.get_columns(); ++i) {
+		for (unsigned int i = 0; i < rh.get_columns(); ++i) {
```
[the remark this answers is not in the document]
e3m3: It's just not clear. Not everyone is madly in love with C/C++ like we are :smiley:

runner/runner/main.py
```diff
@@ -102,6 +102,7 @@ def load_native(config: Mapping[str, Any]) -> None:
         ILLIXR_DEMO_DATA=str(demo_data_path),
         ILLIXR_OFFLOAD_ENABLE=str(enable_offload_flag),
         ILLIXR_ALIGNMENT_ENABLE=str(enable_alignment_flag),
+        ILLIXR_RUN_DURATION=str(config["action"].get("ILLIXR_RUN_DURATION", 60)),
```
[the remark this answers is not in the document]
e3m3: This gets reverted in #198.

timewarp_gl/plugin.cpp
e3m3: We can remove it. This one was just for testing/abundance of caution.

common/shader_util.hpp
```diff
@@ -17,78 +19,82 @@ static void GLAPIENTRY
...
     glCompileShader(vertex_shader_handle);
     glGetShaderiv(vertex_shader_handle, GL_COMPILE_STATUS, &result);
-    if ( result == GL_FALSE )
-    {
+    if (result == GL_FALSE) {
         GLchar msg[4096];
         GLsizei length;
         glGetShaderInfoLog( vertex_shader_handle, sizeof( msg ), &length, msg );
-        printf( "1 Error: %s\n", msg);
+        std::cerr << "1 Error: " << msg << std::endl;
```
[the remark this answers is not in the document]
e3m3: I removed the numbers, but left the print so that we can get info from `msg`. Can remove also.

common/shader_util.hpp
```diff
@@ -17,78 +19,82 @@ static void GLAPIENTRY
...
     }
-	fprintf( stderr, "GL CALLBACK: %s type = 0x%x, severity = 0x%x, message = %s\n",
-			( type == GL_DEBUG_TYPE_ERROR ? "** GL ERROR **" : "" ),
-				type, severity, message );
+    std::cerr << "GL CALLBACK: " << (type == GL_DEBUG_TYPE_ERROR ? "** GL ERROR **" : "")
+              << " type = 0x" << std::hex << type << std::dec
+              << ", severity = 0x" << std::hex << severity << std::dec
+              << ", message = " << message
+              << std::endl;
+    ILLIXR::abort();
```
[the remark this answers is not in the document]
e3m3: We check more explicitly using the `severity` interface above. Continuing to use `severity` over `type` might be more robust.

Either we can:
- Return early for `GL_DEBUG_SEVERITY_LOW` and `GL_DEBUG_SEVERITY_MEDIUM` like for `GL_DEBUG_SEVERITY_NOTIFICATION`, or
- Only return early for `GL_DEBUG_SEVERITY_NOTIFICATION`, and guard the `abort` for `GL_DEBUG_SEVERITY_LOW` and `GL_DEBUG_SEVERITY_MEDIUM` so that we get the print out for low, medium, and high, but only crash when it's high.

timewarp_gl/plugin.cpp
```diff
@@ -2,8 +2,8 @@
 #include <future>
 #include <iostream>
 #include <thread>
+#include <vector>
```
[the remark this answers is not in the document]
e3m3: Even if it was, we should try to include everything used within a compilation unit (file) so that we can easily implement #213 and #54.

charmoniumQ (dismissed):

common/shader_util.hpp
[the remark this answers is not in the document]
mhuzai: Option 2 sounds great, but let's crash with medium too. This is just to err on the side of caution.

configs/rt_slam_plugins.yaml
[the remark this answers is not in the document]
mhuzai: Nice!

charmoniumQ (dismissed):

mhuzai (approved):

runner.sh
```diff
@@ -7,4 +7,6 @@ cd "$(dirname ${0})/"
 conda deactivate
 conda activate illixr-runner
 
+export ILLIXR_INTEGRATION=yes
```
charmoniumQ, line 10: Let's put this in the `build_one_plugin`. If someone wants to know how to "build one plugin," they will look there, so it is more discoverable there.

charmoniumQ (commented):
Also, how many "extra" stdout messages does this generate?

e3m3:
> Also, how many "extra" stdout messages does this generate?

Every page or so of the `dbg`-enabled output will have about the following amount of text:
```
|| Errno was set: 110 @ ./switchboard_impl.hpp:307[check_queues]
|> Message: switchboard_impl after wait_dequeue_timed
|| Errno was set: 110 @ ./switchboard_impl.hpp:307[check_queues]
|> Message: switchboard_impl after wait_dequeue_timed
|| Errno was set: 110 @ ./switchboard_impl.hpp:307[check_queues]
|> Message: switchboard_impl after wait_dequeue_timed
|| Errno was set: 110 @ ./switchboard_impl.hpp:307[check_queues]
|> Message: switchboard_impl after wait_dequeue_timed
|| Errno was set: 110 @ ./switchboard_impl.hpp:307[check_queues]
|> Message: switchboard_impl after wait_dequeue_timed
thread,140643395774208,opencv threadpool worker,0
|| Errno was set: 110 @ ./switchboard_impl.hpp:307[check_queues]
```
I'm not too concerned about the amount of output here, but it will be dependent on how many future `errno` setting calls we make. I can add a quieting flag here if we want.

charmoniumQ (approved):
Approval, conditioned on it passing CI some day.

charmoniumQ (approved):

mhuzai (approved):

PR #200, opened 2021-02-11
Dump image and pose for quality evaluation

Boyuan-Tian:
A feature that enables ILLIXR to dump the generated image for quality assessment.

Due to the fresh rate limitation, we could only dump the texture image before the timewarp and the corresponding poses. 

An offline timewarp is applied to dumped images to generate the final result before fed into any image quality assessment algorithms such as SSIM, FILP, and PSNR, etc.

To-do:
[ ] Add switch to enable/disable data collecting in timewarp_gl.


common/common.mk
```diff
@@ -3,8 +3,8 @@
 CXX := clang++-10
 STDCXX ?= c++17
 CFLAGS := $(CFLAGS) -DGLSL_VERSION='"330 core"'
-DBG_FLAGS ?= -Og -g -Wall -Wextra -Werror
-OPT_FLAGS ?= -O3 -DNDEBUG -Wall -Wextra -Werror
+DBG_FLAGS ?= -Og -g -Wall -Wextra -Werror -Wno-missing-field-initializers
```
Boyuan-Tian, line 6: `common/gl_util/lib/stb_image_write.h` initializes structs with { 0 }, which results in the missing-field-initializers warning.
If we don't want to put too many flags here, I can add each structure's initial value.

configs/native-lookup.yaml
```diff
@@ -4,6 +4,7 @@ plugin_groups:
     - path: pose_lookup
     - path: gldemo/
     - path: debugview/
+    - path: offload_data/
```
Boyuan-Tian, line 7: This plugin can always be loaded, while the `enable_offload` flag handles the real activation/deactivation.
This plugin introduces barely zero overhead if we set the flag to 0. This simplifies the procedure from `activating the flag + load the plugin` to `activate the flag` only.

offload_data/plugin.cpp
```diff
@@ -0,0 +1,187 @@
...
+#include <iomanip>
+#include <fstream>
+#include <numeric>
+#define STB_IMAGE_WRITE_IMPLEMENTATION
+#include "common/gl_util/lib/stb_image_write.h"
+
+using namespace ILLIXR;
+
+std::string
+getenv_or(std::string var, std::string default_) {
```
Boyuan-Tian, line 17: This function is used multiply times by timewarp_gl, offload_data, and pose-lookup. It should be put somewhere in common.

I saw a comment in timewarp_gl that says we will use #198 instead of this, so I temporally copied this function to offload_data and pose-lookup.

offload_data/plugin.cpp
```diff
@@ -0,0 +1,187 @@
...
+	std::unique_ptr<reader_latest<texture_pose>> _offload_data_reader;
+	long long _seq_expect, _stat_processed, _stat_missed;
+	std::vector<int> _time_seq;
+	std::vector<const texture_pose*> _offload_data_container;
+
+	int percent = 0;
+	int img_idx = 0;
+	bool enable_offload;
+	bool is_success = false;
+	std::string obj_dir = "metrics/offloaded_data/";
```
Boyuan-Tian, line 82: We might not need to expose the flexibility to set output path to users, as analysis scripts will consume the output data and return the final result that users are interested.

Per the discussion with Sam, the offloaded data is written to the same place (metrics/offloaded_data/) every time running, but the path can also be changed using environment variable if necessary. (see line 35)

pose_lookup/plugin.cpp
```diff
@@ -115,21 +159,27 @@ class pose_lookup_impl : public pose_prediction {
...
 	time_type _m_start_of_time;
 	std::unique_ptr<reader_latest<time_type>> _m_vsync_estimate;
+
+	bool enable_alignment;
+	Eigen::Vector3f init_pos_offset {0};
+	Eigen::Matrix3f align_rot;
+	Eigen::Vector3f align_trans;
+	Eigen::Vector4f align_quat;
+	double align_scale;
+	std::string path_to_alignment = "./metrics/alignMatrix.txt";
```
Boyuan-Tian, line 176: Similar to the `offload_data` plugin, I hard-coded the path to alignment. 

I think uses should be agnostic to these kinds of intermediate data. 

In the future, we could either provide the pre-computed alignment files for each data sequence, or set the output path of alignment computation that is pointing here.

pose_lookup/plugin.cpp
```diff
@@ -5,83 +5,127 @@
 #include "common/data_format.hpp"
 #include "common/plugin.hpp"
 
-/*pyh: reusing data_loading from ground_truth_slam*/
+#include "utils.hpp"
 #include "data_loading.hpp"
 
 using namespace ILLIXR;
 
+std::string
```
Boyuan-Tian, line 13: Similar to the comment in the `offload_data` plugin.

pose_lookup/plugin.cpp
```diff
@@ -5,83 +5,127 @@
...
-        return offset;
-    }
+		return offset;
+	}
 
 	virtual pose_type correct_pose(const pose_type pose) const override {
 		pose_type swapped_pose;
+		pose_type input_pose;
+
+		// Step 1: Compensate starting point to (0, 0, 0), pos only
```
Boyuan-Tian, line 73: An elegant way to get rid of the data copy here is to make the input pose non-const, but I am not sure if that is an acceptable solution.

Boyuan-Tian:
Also, I noticed that the programming style is quite different across multiple files, like the brace style and indentation.

Is there any unified style or guidance I can follow? I can tweak them later.

e3m3:
> Also, I noticed that the programming style is quite different across multiple files, like the brace style and indentation.
> 
> Is there any unified style or guidance I can follow? I can tweak them later.

Coming very soon once I get around to #45! [This](https://github.com/WebKit/WebKit/blob/main/.clang-format) is the styling I am thinking of basing the format on.

common/global_module_defs.hpp
```diff
@@ -19,4 +19,23 @@ static constexpr int FB_WIDTH = 2560; //Pixels
...
+}
+
+/// Temporary environment variable getter. Not needed once #198 is merged.
+std::string getenv_or(std::string var, std::string default_) {
+    if (std::getenv(var.c_str())) {
+        return {std::getenv(var.c_str())};
+    } else {
+        return default_;
+    }
+}
```
e3m3, line 39: Moved these here to reduce clutter and changes once #198 is merged.

configs/ci.yaml
```diff
@@ -11,6 +11,10 @@ data:
     archive_path:
       download_url: 'http://robotics.ethz.ch/~asl-datasets/ijrr_euroc_mav_dataset/vicon_room1/V1_02_medium/V1_02_medium.zip'
 demo_data: demo_data/
+
+enable_offload:   False
+enable_alignment: False
+
```
e3m3, line 17: Added missing definitions

configs/clean.yaml
```diff
@@ -44,3 +44,5 @@ action:
 profile: opt
 data: data
 demo_data: demo_data
+enable_offload: False
+enable_alignment: False
```
e3m3, line 48: Added missing definitions

configs/headless.yaml
```diff
@@ -11,6 +11,10 @@ data:
     archive_path:
       download_url: 'http://robotics.ethz.ch/~asl-datasets/ijrr_euroc_mav_dataset/vicon_room1/V1_02_medium/V1_02_medium.zip'
 demo_data: demo_data/
+
+enable_offload:   False
+enable_alignment: False
+
```
e3m3, line 17: Added missing definitions

configs/monado.yaml
```diff
@@ -8,6 +8,10 @@ data:
     archive_path:
       download_url: 'http://robotics.ethz.ch/~asl-datasets/ijrr_euroc_mav_dataset/vicon_room1/V1_02_medium/V1_02_medium.zip'
 demo_data: demo_data/
+
+enable_offload:   False
+enable_alignment: False
+
```
e3m3, line 14: Added missing definitions

offload_data/plugin.cpp
```diff
@@ -0,0 +1,184 @@
...
+		, _seq_expect(1)
+		, _stat_processed(0)
+		, _stat_missed(0)
+		, percent{0}
+		, img_idx{0}
+		, enable_offload{ILLIXR::str_to_bool(ILLIXR::getenv_or("ILLIXR_OFFLOAD_ENABLE", "False"))}
+		, is_success{true}
+		/// TODO: Set with #198
+		, obj_dir{ILLIXR::getenv_or("ILLIXR_OFFLOAD_PATH", "metrics/offloaded_data/")}
+		{
```
e3m3, line 32: Moved initialization to constructor to reduce the number of changes needed once #198 is merged.

offload_data/plugin.cpp
```diff
@@ -0,0 +1,184 @@
...
+			/// Remove existing file and create folder for offloading
+			/// TODO: Not portable
+			system(("rm -r " + obj_dir).c_str());
+			system(("mkdir -p " + obj_dir).c_str());
+		}
```
e3m3, line 37: Not portable, but we don't have to address this at the moment.

offload_data/plugin.cpp
```diff
@@ -0,0 +1,184 @@
...
+				meta_file << it << " ";
+			meta_file << std::endl << std::endl << std::endl;
+
+			meta_file << "ordered time: " << std::endl;
+			std::sort(_time_seq.begin(), _time_seq.end(), [](int x, int y) {return x > y;});
+			for (int& it : _time_seq)
+				meta_file << it << " ";
+		}
+		meta_file.close();
+	}
```
e3m3, line 113: `meta_file.close()` may still fail if something goes wrong. We may print a warning here in that scenario, if desired.

offload_data/plugin.cpp
```diff
@@ -0,0 +1,184 @@
...
+				pose_file << container_it->latest_quaternion.y() << " ";
+				pose_file << container_it->latest_quaternion.z() << std::endl;
+
+				pose_file << "render_pose_orientation: ";
+				pose_file << container_it->render_quaternion.w() << " ";
+				pose_file << container_it->render_quaternion.x() << " ";
+				pose_file << container_it->render_quaternion.y() << " ";
+				pose_file << container_it->render_quaternion.z();
+			}
+			pose_file.close();
```
e3m3, line 170: Same as `meta_file.close()`.

pose_lookup/plugin.cpp
```diff
@@ -5,138 +5,185 @@
...
+
+        // Step 1: Compensate starting point to (0, 0, 0), pos only
+        input_pose.position(0) = pose.position(0) - init_pos_offset(0);
+        input_pose.position(1) = pose.position(1) - init_pos_offset(1);
+        input_pose.position(2) = pose.position(2) - init_pos_offset(2);
+        input_pose.orientation.x() = pose.orientation.x();
+        input_pose.orientation.y() = pose.orientation.y();
+        input_pose.orientation.z() = pose.orientation.z();
+        input_pose.orientation.w() = pose.orientation.w();
+
```
e3m3, line 78: I'm not bothered by the extra copying here as `input_pose` is a temporary local. (@charmoniumQ @mhuzai) Thoughts?

e3m3 (approved):

timewarp_gl/plugin.cpp
```diff
@@ -576,6 +641,24 @@ class timewarp_gl : public threadloop {
...
+			auto offload_data = new texture_pose;
+			offload_data->seq = ++_offload_seq;
+			offload_data->image = image;
+			offload_data->pose_time = lastSwapTime;
+			offload_data->offload_time = offload_time;
+			offload_data->position = latest_pose.pose.position;
+			offload_data->latest_quaternion = latest_pose.pose.orientation;
+			offload_data->render_quaternion = most_recent_frame->render_pose.pose.orientation;
+
+			_m_offload_data->put(offload_data);
```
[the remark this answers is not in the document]
e3m3: Actually as pointed out by @mhuzai, this won't be resolved even with #32 as we need to store the metadata without affecting system performance (for accurate timing). We will need to partially address this some other way in the future (e.g., compression).

common/common.mk
charmoniumQ: It's better to use `#pragma GCC diagnostic` around the import site, so the warning is active for the rest of the code. See [this SO post](https://stackoverflow.com/q/3378560/1078199).

e3m3: Nice solution for external dependencies

offload_data/plugin.cpp
```diff
@@ -0,0 +1,184 @@
...
+				} else {
+					_stat_missed = 0;
+				}
+				_stat_processed++;
+				_seq_expect = in->seq+1;
+				return skip_option::run;
+			}
+		}
+
+		void _p_one_iteration() override {
```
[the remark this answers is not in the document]
mhuzai: It is supposed to be synchronous but it is currently implemented this way because until SB 2.0 gets merged in, we don't have parallel synchronous callbacks.

common/common.mk
Boyuan-Tian: Removed the flags in common.mk and added `#pragma GCC diagnostic` in `offload_data/plugin.cpp`

offload_data/plugin.cpp
Boyuan-Tian: Boost Filesystem is applied now.

pose_lookup/plugin.cpp
Boyuan-Tian: Use the functional-style transformation for this part now.

timewarp_gl/plugin.cpp
```diff
@@ -158,8 +158,55 @@ class timewarp_gl : public threadloop {
...
+
+        GLubyte* readTextureImage(){
+
+                GLubyte* pixels = new GLubyte[SCREEN_WIDTH * SCREEN_HEIGHT * 3];
+
+                // Start timer
+                startGetTexTime = std::chrono::high_resolution_clock::now();
+
+                // Enable PBO buffer
+                glBindBuffer(GL_PIXEL_PACK_BUFFER, PBO_buffer);
```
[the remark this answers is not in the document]
Boyuan-Tian: Added glGetError for all the OpenGL calls in the `readTextureImage` function.

e3m3 (approved):
This branch was updated via a merge with master. We will need to use the Update Branch button to merge this in.

charmoniumQ (approved):

pose_lookup/plugin.cpp
```diff
@@ -4,139 +4,189 @@
...
+        , _m_start_of_time{std::chrono::high_resolution_clock::now()}
+        , _m_vsync_estimate{sb->subscribe_latest<time_type>("vsync_estimate")}
+        /// TODO: Set with #198
+        , enable_alignment{ILLIXR::str_to_bool(getenv_or("ILLIXR_ALIGNMENT_ENABLE", "False"))}
+        , init_pos_offset{0}
+        , align_rot{Eigen::Matrix3f::Zero()}
+        , align_trans{0}
+        , align_quat{0}
+        , align_scale{0.0}
+        , path_to_alignment{ILLIXR::getenv_or("ILLIXR_ALIGNMENT_FILE", "./metrics/alignMatrix.txt")}
```
mhuzai, line 31: Who sets this?

offload_data/plugin.cpp
```diff
@@ -0,0 +1,189 @@
...
+		, _offload_data_reader{sb->subscribe_latest<texture_pose>("texture_pose")}
+		, _seq_expect(1)
+		, _stat_processed(0)
+		, _stat_missed(0)
+		, percent{0}
+		, img_idx{0}
+		, enable_offload{ILLIXR::str_to_bool(ILLIXR::getenv_or("ILLIXR_OFFLOAD_ENABLE", "False"))}
+		, is_success{true}
+		/// TODO: Set with #198
+		, obj_dir{ILLIXR::getenv_or("ILLIXR_OFFLOAD_PATH", "metrics/offloaded_data/")}
```
mhuzai, line 35: Who sets this?

offload_data/plugin.cpp
```diff
@@ -0,0 +1,189 @@
...
+				_stat_missed = 0;
+			}
+			_stat_processed++;
+			_seq_expect = in->seq+1;
+			return skip_option::run;
+		}
+	}
+
+	void _p_one_iteration() override {
+		std::cout << "Image index: " << img_idx++ << std::endl;
```
mhuzai, line 57: Guard print.

mhuzai (commented):
Too late given that this has already been merged in, but some of these comments should still be addressed.

Boyuan-Tian:
> Too late given that this has already been merged in, but some of these comments should still be addressed.

Sorry... I thought we are good to go, but I really should have asked everyone before I clicked the button...
Can we withdraw the last merge commit in the main branch and then fix these issues? Hopefully it won't be too late...

pose_lookup/plugin.cpp
e3m3: This is set right here, given that no environment variable will exist until set via #198. A transitional hook.

offload_data/plugin.cpp
e3m3: This is set right here, given that no environment variable will exist until set via #198. A transitional hook.

e3m3:
> > Too late given that this has already been merged in, but some of these comments should still be addressed.
> 
> Sorry... I thought we are good to go, but I really should have asked everyone before I clicked the button...
> Can we withdraw the last merge commit in the main branch and then fix these issues? Hopefully it won't be too late...

We can open a new issue. No worries.

pose_lookup/plugin.cpp
mhuzai: Yeah, I saw the #198 comment but I meant was this (sorry, should have clarified): the enabling/disabling flag is already set via the config and runner. Why not do the same for the path (here and in offload_data)? At a minimum, it would at least save recompilation.

mhuzai:
> 
> 
> > > Too late given that this has already been merged in, but some of these comments should still be addressed.
> > 
> > 
> > Sorry... I thought we are good to go, but I really should have asked everyone before I clicked the button...
> > Can we withdraw the last merge commit in the main branch and then fix these issues? Hopefully it won't be too late...
> 
> We can open a new issue. No worries.

Yeah, the changes are minor, so we can just make a new "minor fixes to offload data" type of PR.

pose_lookup/plugin.cpp
e3m3: These variables can still be set easily by adding an `export ILLIXR_ALIGNMENT_FILE=...` in the user's environment via the shell or in `runner.sh` before the `python` command.

Any flags set pre-#198 by the config need to be added to the schema and the runner environment variable overrides. These changes would then have to be reverted once #198 is merged. I let it be for the flags, since those had already been implemented.

mhuzai: Okay, that makes sense. I guess that just leaves the variable setting for `load_monado` and the print guard for the fix PR. @Boyuan-Tian

Issue #104, opened 2020-09-11
Proposal for using the Nix build system


therishidesai:
## Transitioning to Nix
The following is a proposal on why we should move to Nix, how we should move to Nix, and what are some of the unknowns if we move to Nix.

### Why Nix?
[Nix](https://nixos.org/) is a package management and build system that allows for reproducible builds. For ILLIXR, we have code from many different areas that have a wide variety of dependencies. Currently, we apt-install or build from source all of the dependencies on to the users host machine. This makes a lot of assumptions as to what their underlying system has and there is a chance we may ruin the user's apt repositories and other projects with the compiler changes that we make. Finally, this install process isn't reliably portable to other systems and as we want to move to ARM systems in the future this will be important. Nix solves almost all of these issues because of its design. When you build a package in Nix it is an entirely indpenedent build so it doesn't pollute the rest of the system. In addition, Nix has clean cross compilation support which allow us to easily port to ARM. Nix also has a smart cacheing system that will speed up many of our builds. Finally, we will be able to use Nix to move away from git submodules which continue to cause us a lot of issues.

### How we use Nix?
The first step in moving to Nix is writting a Nix derivation (Nix script with build instructions) for each one of our components. For components like OpenVINS this will be fairly easy since they are standard C++ project. We also need to write a Nix derivation for our ILLIXR common files. The first challenge that comes with Nix is how to use OpenGL and CUDA on the users machine. This becomes a challenge because of the machine specifics with Nvidia and mesa drivers. The solution in the Nix ecosystem is to use something called [nixGL](https://github.com/guibou/nixGL). What nixGL does is find the Nvidia/mesa driver that is installed locally on the user's machine and pulls the same driver in to the Nix environment. The one challenge with nixGL is the user needs to have the most up to date Nvidia driver since Nvidia has a habit of removing the older driver repositories.
The final thing with Nix is how we transition the repository setup. We will remove all of the submodules in our main ILLIXR root repository. Then we will have the root repository contain all of the nix derivations for each component and a src.nix file that points to the git repo for the individual component and information on which branch/commit we want to build from:
```
ILLIXR
  -- components
     -- open_vins
        -- default.nix (Nix build derivation)
        -- src.nix
```
With this repository setup, all of the code changes will be in their individual repos and then the master repo will have a PR where a contributor just bumps the commit version in the src.nix file.

### Unkowns with Nix
[...]

PR #242, opened 2021-03-11
Fixed gtsam, opengv, and Eigen installation

### LINKED ISSUE #239: GTSAM installation script points to wrong shared library
The GTSAM install script runs `make install` before attempting to repair the symlink to `libgtsam.so`. The symlinks installed to the `prefix_dir` may not point to the `.so` file for the build type enabled (e.g. `RelWithDebInfo`).

`make` and `make install` should be called separately.

### LINKED ISSUE #241: Eigen should be common between opengv, gtsam, and kimera-vio
The current install scripts for gtsam point to the source's include directory which may become inconsistent with the Eigen installed at `/usr/local`, especially if the system package is also installed. See `Kimera-VIO/docs/kimera_vio_install.md`. Todo:
- Remove Eigen from the package install script (`scripts/install_apt_deps.sh`)
- Use `/usr/local` as the prefix directory for installation
- Set the Eigen include directories

e3m3:
Closes #241.
Merging with #240. Closes #239.


scripts/install_opengv.sh
```diff
@@ -1,11 +1,54 @@
 #!/bin/bash
 
-git clone https://github.com/laurentkneip/opengv.git "${opt_dir}/opengv"
+#--- Script for installing opengv ---#
 
+
+### Default imported variables setup ###
+
+if [ -z "${opt_dir}" ]; then
```
charmoniumQ, line 8: Will these get annoying to write in every script? The default values are copied into N places. Imagine if we want to change one of these vars (like in the past, when we added `opt_dir`).

I think we should write `set -eux` and not have per-script defaults.

e3m3: That's a good point. I wanted to allow the scripts to run standalone without needing the values to be passed from `install_deps.sh` which isn't run from CI directly. Without these, I think `set -eux` would cause the Dockerfile script to fail. I think the best option for defaults now that these are in multiple files is to source them from a common bash script like the helper functions.

e3m3:
*Dependency hell*
- OpenCV required `libvtk6-dev` which requires `libeigen3-dev`.
- Installing eigen from source before OpenCV is not an option because removing `libvtk6-dev` disables support for `<opencv2/viz/types.hpp>` (needed by Kimera-VIO).
- We cannot install VTK (`v6.3.0`) from source because our compiler toolchain is too new.
- Reverting to installing OpenCV from source using system packages available, _prior_ to removing `libeigen3-dev` and `libvtk6-dev` for the from-source eigen installation used by the rest of our dependencies. This means that, on Ubuntu 18.04 LTS, OpenCV is compiled using Eigen `3.3.4`. The remaining packages are installed using Eigen `3.3.9` from source.

#104 / #246 should be high priority.

charmoniumQ (approved):

scripts/install_gtsam.sh
```diff
@@ -3,30 +3,21 @@
...
 
+## Source the default values for imported variables
+. scripts/default_values.sh
 
-### Helper functions ###
-
-# Source the global helper functions
-. scripts/bash_utils.sh
+## Source the global helper functions
+. scripts/system_utils.sh
```
mhuzai, line 12: Why isn't this being done for all the other install scripts (openxr, qemu, etc.)? They also use the variables sourced in these two bash files, and to further add to the confusion, a subset of these variables is also set in `install_deps.sh` (e.g., line 82 sets `opt_dir`).

mhuzai (commented):
The changes look good in general, but left a few comments.

scripts/install_gtsam.sh
e3m3: > Why isn't this being done for all the other install scripts (openxr, qemu, etc.)?

Other scripts will be moved to a similar structure in dedicated PR's for those installs (to reduce the time to merge for this PR, as well as to improve separation of concerns for debugging). In the future, these scripts should be replaced for portability (e.g., moving dependency handling to the python/Runner), and should target the cross-platform package manager resulting from #104 and #246.

> They also use the variables sourced in these two bash files, and to further add to the confusion, a subset of these variables is also set in install_deps.sh (e.g., line 82 sets opt_dir).

The default values can be overridden by calling scripts like `install_deps.sh`. There are three possible workflows to consider:

1.  Calling script is `Dockerfile`: The values set in the environment take precedent when debugging/testing in CI.

1.  Calling script is `install_deps.sh`: This interactive script sets the values for a local machine install. We don't take user input in this script for the installation locations yet, but the defaults can be overridden here.

1.  There is no calling script (the script is called directly from the CLI or another script): The script being called will not crash due to missing (or erroneously empty) values, and use the default install setup can continue.

I agree these scripts should be cleaned up to unify the structure and reduce confusion. I don't think we should address those issues in this PR.

charmoniumQ (approved):

mhuzai (approved):

PR #248, opened 2021-03-22
Updated contributing guidelines

### LINKED ISSUE #247: Update contributing guidelines
Add details for rebasing and checkpointing branches for pull requests when the `master` branch has been updated.

e3m3:
Closes #247.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -110,7 +110,7 @@ Why are the above steps necessary?
 
 1.  Using rebases keeps the `master` and feature branch histories streamlined (minimizing branching),
         thus making it easier to compose feature branches for integration testing.
-    See this article on [rebasing public brances][3] for more information.
+    See this article on [rebasing public branches][3] for more information.
 
 If your PR has not seen activity from the ILLIXR team after a long period of time (e.g., 2 weeks),
     feel free to contact the team directly on the GitHub Issue Conversation tab or at
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -24,19 +24,79 @@ Please follow these steps when making pull requests (PRs):
     Please give the PR a descriptive name.
 
 1.  As you make progress on your PR, keep your branch up-to-date with the `master` branch which
-        may have been updated after starting your PR.
+        may have been updated *after* starting your PR.
     Your PR **MUST** be updated to reflect changes to `master` in order to be merged.
-    Use the following procedure for updating your branch:
+    Use the following procedure for updating your branch and are ready to commit your changes:
 
     <!--- language: lang-shell -->
 
-        # While on your PR branch <issue-branch>
-        git commit
+        ## While on your PR branch <issue-branch> hosted at <your-remote> repository:
+		git commit
         git checkout master
-        git pull origin --rebase
+        git pull <illixr-remote> --rebase                           ## Line A
         git checkout <issue-branch>
-        git rebase master
-        git push origin <issue-branch> --force-with-lease
+        git rebase master                                           ## Line B
+        git push <your-remote> <issue-branch> --force-with-lease    ## Line C
+
+	If you want to update your branch while you are still making changes to your branch:
+
+    <!--- language: lang-shell -->
+
+        ## While on your PR branch <issue-branch> hosted at <your-remote> repository:
+		git stash
+        git checkout master
+        git pull <illixr-remote> --rebase                           ## Line A
+        git checkout <issue-branch>
+        git rebase master                                           ## Line B
+		git stash apply <stash-number>
+		## Finish your changes
+		git commit
+        git push <your-remote> <issue-branch> --force-with-lease    ## Line C
+
+    For ILLIXR team members:
+
+	-	In the examples above, `<illixr-remote>` and `<your-remote>` are the same.
+
+    -   When collaborating on branches in our repository, `Line A` may pull in changes that overwrite
+            the git commit history when performing `Line B`.
+		Subsequently, performing `Line C` will rewrite the history in the public branch.
+		To preserve branch commit histories in the case that a rollback is needed, we will employ
+			a checkpointing process for force updated branches.
+		This process will be manually performed, but may be automated in the future.
+
+		If `Line A` shows an update to master, the following example illustates your local repository
+			just after performing `Line A`:
+
+			A -- B -- C	-- P -- Q -- R					## master
+					   \
+						D -- E -- F 					## issue-123-fixing-bug
+
+		To checkpoint the `issue-123-fixing-bug` while it is checked out:
+
+        <!--- language: lang-shell -->
+
+		    git branch issue-123.0-fixing-bug                                   ## Make alias for old issue-123-fixing-bug
+		    git checkout -b issue-123.1-fixing-bug                              ## Make new branch to rebase with master
+		    git rebase master                                                   ## Replay issue-123-fixing-bug onto master
+		    git branch -D issue-123-fixing-bug                                  ## Remove old issue-123-fixing-bug
+		    git branch issue-123-fixing-bug                                     ## Make issue-123-fixing-bug an alias of new branch
+		    git push <illixr-remote> issue-123.{0,1}-fixing-bug                 ## Push new checkpointed branches to remote
+		    git push <illixr-remote> issue-123-fixing-bug --force-with-lease    ## Force update issue-123-fixing-bug
+
+		After checkpointing, your local repository should look as follows:
+
+									   D' -- E' -- F'   ## issue-123.1-fixing-bug, issue-123-fixing-bug
+									  /
+			A -- B -- C	-- P -- Q -- R					## master
+					   \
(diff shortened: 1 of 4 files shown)


CONTRIBUTING.md
```diff
@@ -24,19 +24,79 @@ Please follow these steps when making pull requests (PRs):
...
+        git stash
+        git checkout master
+        git pull <illixr-remote> --rebase                           ## Line A
+        git checkout <issue-branch>
+        git rebase master                                           ## Line B
+        git stash apply <stash-number>
+        ## Finish your changes
+        git commit
+        git push <your-remote> <issue-branch> --force-with-lease    ## Line C
+
```
e3m3, line 55: This process may be onerous for new contributors or non-team members. Suggestions for facilitating collaboration are welcome.

CONTRIBUTING.md
```diff
@@ -24,19 +24,79 @@ Please follow these steps when making pull requests (PRs):
...
+            A -- B -- C -- P -- Q -- R                  ## master
+                       \
+                        D -- E -- F                     ## issue-123-fixing-bug
+
+        To checkpoint the `issue-123-fixing-bug` while it is checked out:
+
+        <!--- language: lang-shell -->
+
+            git branch issue-123.0-fixing-bug                                   ## Make alias for old issue-123-fixing-bug
+            git checkout -b issue-123.1-fixing-bug                              ## Make new branch to rebase with master
```
charmoniumQ, line 79: Shouldn't `issue-123.1-fixing-bug` be created after the rebase?

CONTRIBUTING.md
```diff
@@ -24,19 +24,79 @@ Please follow these steps when making pull requests (PRs):
...
+
+    <!--- language: lang-shell -->
+
+        ## While on your PR branch <issue-branch> hosted at <your-remote> repository:
+        git stash
+        git checkout master
+        git pull <illixr-remote> --rebase                           ## Line A
+        git checkout <issue-branch>
+        git rebase master                                           ## Line B
+        git stash apply <stash-number>
```
charmoniumQ, line 51: Use `git stash pop` instead of git stash apply to also keep your stash clean. If popping fails for some reason, git will not drop your work.

CONTRIBUTING.md
```diff
@@ -24,19 +24,79 @@ Please follow these steps when making pull requests (PRs):
...
         git checkout master
-        git pull origin --rebase
+        git pull <illixr-remote> --rebase                           ## Line A
         git checkout <issue-branch>
-        git rebase master
-        git push origin <issue-branch> --force-with-lease
+        git rebase master                                           ## Line B
+        git push <your-remote> <issue-branch> --force-with-lease    ## Line C
+
+    If you want to update your branch while you are still making changes to your branch:
```
charmoniumQ, line 41: One could merge these two sections with:

```
git commit # or git stash
...
# if you did git stash, then do: git stash pop && git commit
...
```

IMO easier to read and understand the differences/commonalities between the two occasions.

CONTRIBUTING.md
```diff
@@ -24,19 +24,79 @@ Please follow these steps when making pull requests (PRs):
...
+    Use the following procedure for updating your branch and are ready to commit your changes:
 
     <!--- language: lang-shell -->
 
-        # While on your PR branch <issue-branch>
+        ## While on your PR branch <issue-branch> hosted at <your-remote> repository:
         git commit
         git checkout master
-        git pull origin --rebase
+        git pull <illixr-remote> --rebase                           ## Line A
```
charmoniumQ, line 36: Why pull into master instead of pulling directly into your feature branch?

Pulling into master should be done just before starting a new feature branch, which might entail adding a new section here called "starting a new feature branch" or something.

CONTRIBUTING.md
charmoniumQ: Re "this is complicated for new devs": should we recommend git-aliases or scripts for: new feature branch, rebase && incrment checkpoint, and rebase feature branch onto updated feature branch?

The latter is a workflow operation you have not yet described. Suppose Alice and Bob are collaborating on the same feature. Alice writes `alice/issue-123.3`, Bob writes a different `bob/issue-123.3`. Alice updates the GitHub repo first. What does Bob do? He should rebase his work onto the new `origin/issue-123.3` and increment the branch checkpoint to `issue-123.4`.

CONTRIBUTING.md
e3m3: The branch is created and checked out here as an alias of `issue-123.0-fixing-bug`. The following rebase will then modify `issue-123.1-fixing-bug` in-place, assuming the rebase completes successfully (including handling conflicts). After this process, `issue-123.0-fixing-bug` remains at the old history. If we rebase first, we need to make sure that the resulting branch is at a new, unlabeled detach head for which we then apply an alias/rename. We also need to make sure to preserve the old history.

CONTRIBUTING.md
e3m3: I used `git stash apply` here to not assume that changes from this branch/feature are at the top of the stash queue. `git stash drop <stash-number>` can then be used to clean up.

CONTRIBUTING.md
e3m3: > As you make progress on your PR, keep your branch up-to-date with the `master` branch which may have been updated *after* starting your PR.

The idea here is that pulling and rebasing into`master` should be done frequently, and for all ongoing work on existing and new branches. Using this scheme, all updates to `master` are fast-forwards, and `master` is always up-to-date. The process for continuing work on a branch and starting work on a new branch off of `master` become the same, except for the case of true rebase conflicts from bringing an existing feature branch in line with `master`. I think this becomes more important as there are more concurrent feature branches in progress; the state of one's local repository can be assumed to be clean and up-to-date as branches are context switched.

charmoniumQ (approved):

CONTRIBUTING.md
```diff
@@ -24,19 +24,153 @@ Please follow these steps when making pull requests (PRs):
...
+
+            A -- B -- C -- P -- Q -- R                                          ## master
+                       \
+                        D -- E -- F                                             ## issue-123-fixing-bug
+
+        In this example, commits `P`, `Q`, and `R` have been merged to `master`
+            (from feature branches not shown) after feature branch `issue-123-fixing-bug` was
+            forked from `master`.
+
+        To checkpoint the `issue-123-fixing-bug` while it is checked out:
```
mhuzai, line 71: Most of the "complexity" in this document is coming from checkpointing. Why is checkpointing necessary, does every developer need to do it, and can we live without it?

mkdocs.yml
```diff
@@ -18,6 +18,7 @@ nav:
         - 'ILLIXR Glossary (Version 2)': glossary.md
         - 'ILLIXR API (Version 2)': api/html/annotated.html
         - 'ILLIXR License (Version 2)': LICENSE.md
+        - 'Contributing Guidelines (Version 2)': CONTRIBUTING.md
```
mhuzai, line 21: Unrelated to this PR, but making a note: I don't like the `(Version 2)` after every heading. It doesn't look clean at all, and I'm not sure if it's required given that v2 is the default (plus, most people won't even know the difference between v1 and v2). My recommendation would be to remove all the version 2s. For version 1, we can say `Legacy Documentation (Version 1)`, and then remove `(Version 1)` from the individual lines.

mhuzai (commented):
This looks really thorough and really well done! Left a few questions.

mkdocs.yml
e3m3: Agreed. New issue: #282.

CONTRIBUTING.md
e3m3: > Why is checkpointing necessary,

The need to checkpoint arises from using rebases instead of merges. Since rebases rewrite history, we have no choice but to do force-updates when updating a feature branch with commits pushed to `master`. Force-updates can cause lossy git fetches if care is not taken (as described in the document). Checkpoints are only relevant for public branches that are updated via rebases (currently proposed to be all feature branches), _and_ are being contributed to my more than one developer.

> does every developer need to do it

Checkpoints can only be done for branches/repos for which we have write permissions. Thus, there is no need to consider checkpointing for external repos from which a PR is generated. For ILLIXR-owned repos, branches that are being worked on (i.e. actively pushing commits) by multiple developers will require checkpointing. This means that checkpointing is primarily an ILLIXR-team developer concern.

> can we live without it

We can go back to the usual merge based flow. However, we lose some of the benefits afforded by rebased linear branch histories. This is also a perfectly fine solution, but maintenance up-front will be traded off for maintenance on-demand for the maintainer navigating merged histories. Although checkpointing as described is quite heavyweight for a manual process, I think this system would work quite well with automation. Checkpointing is needed to resolve collaboration conflicts, but conflicts do not frequently occur in our workflow currently. Hopefully, this means that we can enforce this manually without the overhead of handling conflicts often. Once automated, we won't have to worry about this any longer.

charmoniumQ (approved):

charmoniumQ (approved):

mhuzai (approved):

PR #249, opened 2021-03-23
Add stoplight to synchronize thread starts/stops

### LINKED ISSUE #217: Synchronize thread starts

charmoniumQ:
This is less critical than #32. Only review if the path to #32 is blocked up.

Closes #217.

Reviewer should
- See how it works (note the `std::condition_variable` code is lifted from [C++ docs](https://en.cppreference.com/w/cpp/thread/condition_variable)).
- Verify that API of `Stoplight` is sane.
- Ensure it works on your machine.

### DIFF
--- a/common/plugin.hpp
+++ b/common/plugin.hpp
@@ -27,8 +27,9 @@ namespace ILLIXR {
 		/**
 		 * @brief A method which Spindle calls when it starts the component.
 		 *
-		 * This is necessary because a constructor can't call derived virtual methods (due to
-		 * structure of C++).
+		 * This is necessary because a constructor can't call derived virtual
+		 * methods (due to structure of C++). See `threadloop` for an example of
+		 * this use-case.
 		 */
 		virtual void start() {
 			record_logger_->log(record{__plugin_start_header, {
--- a/common/plugin.hpp
+++ b/common/plugin.hpp
@@ -26,6 +26,9 @@ namespace ILLIXR {
 
 		/**
 		 * @brief A method which Spindle calls when it starts the component.
+		 *
+		 * This is necessary because a constructor can't call derived virtual methods (due to
+		 * structure of C++).
 		 */
 		virtual void start() {
 			record_logger_->log(record{__plugin_start_header, {
@@ -35,7 +38,7 @@ namespace ILLIXR {
 		}
 
 		/**
-		 * @brief A method which Spindle calls when it starts the component.
+		 * @brief A method which Spindle calls when it stops the component.
 		 *
 		 * This is necessary because the parent class might define some actions that need to be
 		 * taken prior to destructing the derived class. For example, threadloop must halt and join
--- a/common/runtime.hpp
+++ b/common/runtime.hpp
@@ -15,8 +15,18 @@ namespace ILLIXR {
 		virtual void load_so(const std::vector<std::string>& so) = 0;
 		virtual void load_so(const std::string_view so) = 0;
 		virtual void load_plugin_factory(plugin_factory plugin) = 0;
+
+		/**
+		 * Returns when the runtime is completely stopped.
+		 */
 		virtual void wait() = 0;
+
+		/**
+		 * Requests that the runtime is completely stopped.
+		 * Clients must call this before deleting the runtime.
+		 */
 		virtual void stop() = 0;
+
 		virtual ~runtime() {}
 	};
 
--- a/common/runtime.hpp
+++ b/common/runtime.hpp
@@ -27,7 +27,8 @@ namespace ILLIXR {
 		 */
 		virtual void stop() = 0;
 
-		virtual ~runtime() {}
+		virtual ~runtime() = default;
+
 	};
 
 	extern "C" runtime* runtime_factory(GLXContext appGLCtx);
--- a/common/stoplight.hpp
+++ b/common/stoplight.hpp
@@ -1,0 +1,114 @@
+#pragma once
+
+#include <mutex>
+#include <chrono>
+#include <condition_variable>
+#include "phonebook.hpp"
+
+namespace ILLIXR {
+
+/**
+ * @brief A boolean condition-variable.
+ *
+ * Inspired by https://docs.python.org/3/library/threading.html#event-objects
+ */
+class Event {
+private:
+	mutable std::mutex _m_mutex;
+	mutable std::condition_variable _m_cv;
+	bool _m_value = false;
+
+public:
+
+	/**
+	 * @brief Sets the condition-variable to new_value.
+	 *
+	 * Defaults to true, so that set() sets the bool.
+	 */
+	void set(bool new_value = true) {
+		{
+			std::lock_guard lock {_m_mutex};
+			_m_value = new_value;
+		}
+		if (new_value) {
+			_m_cv.notify_all();
+		}
+	}
+
+	/**
+	 * @brief Clears the condition-variable.
+	 */
+	void clear() { set(false); }
+
+	/**
+	 * @brief Test if is set without blocking.
+	 */
+	bool is_set() const {
+		std::unique_lock<std::mutex> lock {_m_mutex};
+		return _m_value;
+	}
+
+	/**
+	 * @brief Wait indefinitely for the event to be set.
+	 */
+	void wait() const {
+		std::unique_lock<std::mutex> lock {_m_mutex};
+		// Check if we even need to wait
+		if (_m_value) {
+			return;
+		}
+		_m_cv.wait(lock, [this] { return _m_value; });
+	}
+
+	/**
+	 * @brief Wait for the event to be set with a timeout.
+	 *
+	 * Returns whether the event was actually set.
+	 */
+	template <class Clock, class Rep, class Period>
+	bool wait_timeout(const std::chrono::duration<Rep, Period>& duration) const {
+		auto timeout_time = Clock::now() + duration;
+		std::unique_lock<std::mutex> lock {_m_mutex};
+		if (_m_value) {
+			return true;
+		}
+		while (_m_cv.wait_until(lock, timeout_time) != std::cv_status::timeout) {
+			if (_m_value) {
+				return true;
+			}
+		}
+		return false;
+	}
+};
+
+class Stoplight : public phonebook::service {
+public:
+	void wait_for_ready() const {
+		_m_ready.wait();
+	}
+
+	void ready() {
+		_m_ready.set();
+	}
+
+	bool should_stop() const {
+		return _m_stop.load();
+	}
+
+	void stop() {
+		_m_stop.store(true);
+	}
+private:
+	Event _m_ready;
+
+	/*
+	  I use an atomic instead of an event for this one, since it is being "checked" not "waited on."
+
+	  Event uses std::condition_variable, which does not support
+	  std::shared_mutex, so readers would need a std::unique_lock,
+	  which would exclude other readers.
+	*/
+	std::atomic<bool> _m_stop {false};
+};
+
+} // namespace ILLIXR
--- a/common/stoplight.hpp
+++ b/common/stoplight.hpp
@@ -98,8 +98,17 @@ public:
 	void stop() {
 		_m_stop.store(true);
 	}
+
+	void wait_for_shutdown() const {
+		_m_shutdown.wait();
+	}
+
+	void shutdown_done() {
+		_m_shutdown.set();
+	}
 private:
 	Event _m_ready;
+	Event _m_shutdown;
 
 	/*
 	  I use an atomic instead of an event for this one, since it is being "checked" not "waited on."
--- a/common/stoplight.hpp
+++ b/common/stoplight.hpp
@@ -44,7 +44,7 @@ public:
 	 * @brief Test if is set without blocking.
 	 */
 	bool is_set() const {
-		std::unique_lock<std::mutex> lock {_m_mutex};
+		std::lock_guard<std::mutex> lock {_m_mutex};
 		return _m_value;
 	}
 
--- a/common/stoplight.hpp
+++ b/common/stoplight.hpp
@@ -101,14 +101,33 @@ public:
  */
 class Stoplight : public phonebook::service {
 public:
-	const Event& ready() const { return _m_ready; }
-	Event& ready() { return _m_ready; }
+	void wait_for_ready() const {
+		_m_ready.wait();
+	}
 
-	const Event& should_stop() const { return _m_should_stop; }
-	Event& should_stop() { return _m_should_stop; }
+	void signal_ready() {
+		_m_ready.set();
+	}
 
-	const Event& shutdown_complete() const { return _m_shutdown_complete; }
-	Event& shutdown_complete() { return _m_shutdown_complete; }
+	bool check_should_stop() const {
+		return _m_should_stop.is_set();
+	}
+
+	void signal_should_stop() {
+		_m_should_stop.set();
+	}
+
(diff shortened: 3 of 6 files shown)


runtime/runtime_impl.hpp
```diff
@@ -49,6 +51,7 @@ class runtime_impl : public runtime {
 		std::for_each(plugins.cbegin(), plugins.cend(), [](const auto& plugin) {
 			plugin->start();
 		});
+		pb.lookup_impl<Stoplight>()->ready();
```
mhuzai, line 54: Add comment saying this unblocks/starts/kicks off all threadloops?

runtime/runtime_impl.hpp
```diff
@@ -64,21 +67,21 @@ class runtime_impl : public runtime {
...
 		for (const std::unique_ptr<plugin>& plugin : plugins) {
 			plugin->stop();
 		}
-		terminate.store(true);
 	}
 
 	virtual ~runtime_impl() override {
-		if (!terminate.load()) {
+		if (!pb.lookup_impl<Stoplight>()->should_stop()) {
             ILLIXR::abort("You didn't call stop() before destructing this plugin.");
```
mhuzai, line 85: Not a plugin.

runtime/runtime_impl.hpp
```diff
@@ -64,21 +67,21 @@ class runtime_impl : public runtime {
...
 
 	virtual void wait() override {
-		while (!terminate.load()) {
+		while (!pb.lookup_impl<Stoplight>()->should_stop()) {
 			std::this_thread::sleep_for(std::chrono::milliseconds{10});
 		}
 	}
 
 	virtual void stop() override {
+		pb.lookup_impl<Stoplight>()->stop();
```
mhuzai, line 76: Setting this before doing the actual stop would unblock `wait()` while the stop is in progress. Move to end of function.

common/stoplight.hpp
```diff
@@ -0,0 +1,114 @@
...
+	/**
+	 * @brief Clears the condition-variable.
+	 */
+	void clear() { set(false); }
+
+	/**
+	 * @brief Test if is set without blocking.
+	 */
+	bool is_set() const {
+		std::unique_lock<std::mutex> lock {_m_mutex};
```
mhuzai, line 47: Is there a reason to not use `std::lock_guard` here?

common/stoplight.hpp
```diff
@@ -0,0 +1,114 @@
...
+		_m_ready.set();
+	}
+
+	bool should_stop() const {
+		return _m_stop.load();
+	}
+
+	void stop() {
+		_m_stop.store(true);
+	}
```
mhuzai, line 100: Based on this API, is my understanding correct here that stoplight is meant to be only used for one start and one stop, and cannot be used again after it has been stopped?

common/stoplight.hpp
```diff
@@ -0,0 +1,114 @@
...
+		while (_m_cv.wait_until(lock, timeout_time) != std::cv_status::timeout) {
+			if (_m_value) {
+				return true;
+			}
+		}
+		return false;
+	}
+};
+
+class Stoplight : public phonebook::service {
```
mhuzai, line 84: Comments on usage would be good.

mhuzai (commented):
**Higher level decision question:** should we prohibit non-managed threads (not-threadloops) from now on because stoplight won't apply to them and they'll start working immediately? Alternatively, mandate the use of stoplight on non-managed threads? The former might be simpler.

Also, SB threads haven't been changed to use stoplight. That's fine if all the publisher threads use stoplight, but could bypass this start/stop synchronization if the publishers are non-managed threads that don't use stoplight.

common/threadloop.hpp
```diff
@@ -102,10 +102,11 @@ class threadloop : public plugin {
 				break;
 			}
 			case skip_option::stop:
-				stop();
-				break;
+				goto break_loop;
```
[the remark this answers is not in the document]
charmoniumQ: `break` would break out of the `switch`. I want to break out of the loop which encloses the `switch` (like a `break` and then `break`). `goto` is the best way to do that in C/C++ according to [StackOverflow](https://stackoverflow.com/questions/27788326/breaking-out-of-nested-loop-c).

common/plugin.hpp
```diff
@@ -24,6 +24,9 @@ namespace ILLIXR {
 
 		/**
 		 * @brief A method which Spindle calls when it starts the component.
+		 *
+		 * This is necessary because a constructor can't call derived virtual methods (due to
```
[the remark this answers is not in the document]
charmoniumQ: This does not pertain to the logging call; It pertains to `threadloop`s, since the base class wants to call `_p_iteration`, which is virtual in the child class.

runtime/runtime_impl.hpp
charmoniumQ: `Stoplight::stop()` needs to happen first, and then each `threadloop::stop()`.

`Stoplight::stop()` tells the plugin threads to stop after `_p_one_iteration()`; `threadloop::stop()` joins on the plugin threads (waits for them to terminate).

common/stoplight.hpp
charmoniumQ: Yeah, this is because once threads get stopped, there is no way to "unpause" them.

common/stoplight.hpp
charmoniumQ: The CV wait requires a `unique_lock` so it can be re-locked, but I just changed this method to a `lock_guard` as you suggested.

charmoniumQ:
> should we prohibit non-managed threads (not-threadloops) from now on because stoplight won't apply to them and they'll start working immediately?

Yes, they should be prohibited. There are a few places we use non-managed threads (ZED and audio_pipeline, I believe). I am switching these over in [a different PR](https://github.com/ILLIXR/ILLIXR/tree/issue-220-managed-thread-2).

> Also, SB threads haven't been changed to use stoplight. That's fine if all the publisher threads use stoplight, but could bypass this start/stop synchronization if the publishers are non-managed threads that don't use stoplight.

Yes. I wanted to keep SB independent of the rest of ILLIXR including Phonebook (although maybe I should manage that with `#ifdef`s instead).

common/stoplight.hpp
mhuzai: Yeah, that makes sense. I was just wondering whether there is utility in having a reusable stoplight for, say, arbitrarily pausing the entire system and then resuming again. Could be useful for some of the simulation scenarios that we've discussed as well as for debugging/snapshots, etc. Just food for thought.

mhuzai:
> 
> 
> > should we prohibit non-managed threads (not-threadloops) from now on because stoplight won't apply to them and they'll start working immediately?
> 
> Yes, they should be prohibited. There are a few places we use non-managed threads (ZED and audio_pipeline, I believe). I am switching these over in [a different PR](https://github.com/ILLIXR/ILLIXR/tree/issue-220-managed-thread-2).

Great. We should make this clear in the developing a plugin documentation.

> 
> > Also, SB threads haven't been changed to use stoplight. That's fine if all the publisher threads use stoplight, but could bypass this start/stop synchronization if the publishers are non-managed threads that don't use stoplight.
> 
> Yes. I wanted to keep SB independent of the rest of ILLIXR including Phonebook (although maybe I should manage that with `#ifdef`s instead).

Right, "SB-core" can be independent but perhaps "SB-ILLIXR" should use stoplight like the rest of the system. But maybe this is not necessary given that if all the plugins use stoplight, then there is no need for SB to.

runtime/runtime_impl.hpp
mhuzai: Right, that makes sense. But I was talking about `runtime_impl->wait()`. As soon as you call `pb.lookup_impl<Stoplight>()->stop()`, `runtime_impl->wait()` will return, meaning that [this](https://github.com/ILLIXR/ILLIXR/blob/master/runtime/main.cpp#L98) will unblock and cause an exit possibly before the plugins have actually stopped. Or am I missing something?

common/plugin.hpp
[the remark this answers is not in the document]
mhuzai: I'm confused. Can you explain a bit more? What's the relation between `start()` and `_p_one_iteration()`?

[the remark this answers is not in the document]
charmoniumQ: One might think of moving that the constructor, but `plugin::plugin()` calls `threadloop::threadloop()` which calls `the_plugin::the_plugin()` at the end. So creating a thread and calling `the_plugin::_p_one_iteration()` in `threadloop::threadloop()` is not safe; `the_plugin` is not constructed until the end of `threadloop::threadloop()`.

This is why we need `plugin::start()` and `threadloop::start()`. `plugin::start()` calls `threadloop::start()` which starts a thread to call `the_plugin::_p_one_iteration()`, after `the_plugin` is constructed. This is recommended in the [ISO C++ FAQ](https://isocpp.org/wiki/faq/strange-inheritance#calling-virtuals-from-ctor-idiom) when you would need to call a virtual method from the constructor.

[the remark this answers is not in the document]
mhuzai: Thanks, that makes sense. But this is not clear from the current comment and needs to be documented better, so would be good to expand the comment.

common/threadloop.hpp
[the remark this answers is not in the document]
e3m3: If we don't anticipate post-loop code blocks, we could also return directly from the switch case. Neither solution is pretty.

common/stoplight.hpp
charmoniumQ: If we wanted that, I think it should go inside `Stoplight`, but in a different PR.

charmoniumQ: That may be useful. If we wanted that, we should put it in this `Stoplight` class.

charmoniumQ:
I will add some synchronization to Switchboard threads once #32 is merged. All else should be good. I added comments and documentation to address reviewer feedback.

common/runtime.hpp
```diff
@@ -15,8 +15,18 @@ namespace ILLIXR {
...
+		 */
 		virtual void wait() = 0;
+
+		/**
+		 * Requests that the runtime is completely stopped.
+		 * Clients must call this before deleting the runtime.
+		 */
 		virtual void stop() = 0;
+
 		virtual ~runtime() {}
```
e3m3, line 30:
```suggestion
		virtual ~runtime() = default;
```
I think this is slightly different semantically than an empty body (`{ }`), but I'm still not sure which should be [preferred](https://stackoverflow.com/questions/13576055/how-is-default-different-from-for-default-constructor-and-destructor). I've been switching to `= default` as I edit.

docs/writing_your_plugin.md
```diff
@@ -65,6 +65,14 @@ You can also replace any existing functionality this way.
     -   If you need custom concurrency (more complicated than a loop),
             triggered concurrency (by events fired in other plugins),
             or no concurrency then your plugin class should extend [`plugin`][13].
+			
+        - If you spin your own threads, they **must** wait for
+          `pb->lookup_impl<Stoplight>()->wait_for_ready()` the first time they
+          run. This allows the start of all threads in ILLIXR to be
+          synchronized.
```
e3m3, line 72: Should we enforce this with a new derived class for non-managed types?

common/stoplight.hpp
```diff
@@ -0,0 +1,123 @@
...
+ */
+class Event {
+private:
+	mutable std::mutex _m_mutex;
+	mutable std::condition_variable _m_cv;
+	bool _m_value = false;
+
+public:
+
+	/**
```
e3m3, line 23: Thoughts on using [noexcept](https://stackoverflow.com/questions/10787766/when-should-i-really-use-noexcept) for our classes?

common/stoplight.hpp
```diff
@@ -0,0 +1,123 @@
...
+	 * @brief Clears the condition-variable.
+	 */
+	void clear() { set(false); }
+
+	/**
+	 * @brief Test if is set without blocking.
+	 */
+	bool is_set() const {
```
e3m3, line 46: Technically this [blocks](https://en.cppreference.com/w/cpp/thread/mutex/lock) if the lock is held, but I think you are saying that this is unlikely right?

common/stoplight.hpp
```diff
@@ -0,0 +1,123 @@
...
+	Event _m_shutdown;
+
+	/*
+	  I use an atomic instead of an event for this one, since it is being "checked" not "waited on."
+
+	  Event uses std::condition_variable, which does not support
+	  std::shared_mutex, so readers would need a std::unique_lock,
+	  which would exclude other readers.
+	*/
+	std::atomic<bool> _m_stop {false};
```
e3m3, line 120: :+1:

runtime/runtime_impl.hpp
e3m3: Confirming: `wait_for_shutdown()` and `shutdown_done()` were added to address this right?

e3m3:
> Ensure it works on your machine.

Clean launch and drain :+1:.

common/stoplight.hpp
mhuzai: From the [condition variable documentation](https://en.cppreference.com/w/cpp/thread/condition_variable):

> Even if the shared variable is atomic, it must be modified under the mutex in order to correctly publish the modification to the waiting thread.

But this only applies to `set`, right? We should be able to make `is_set()` lock free if we can make `_m_value` atomic?

runtime/runtime_impl.hpp
```diff
@@ -64,22 +70,35 @@ class runtime_impl : public runtime {
 	}
 
 	virtual void wait() override {
-		while (!terminate.load()) {
+		while (!pb.lookup_impl<Stoplight>()->should_stop()) {
```
mhuzai, line 73: This loop is no longer required because of the blocking wait below. `wait_for_shutdown()` can only ever return if `should_stop()` is true, so this check is redundant.
The comment below might need a slight tweak without the presence of this loop.

common/runtime.hpp
charmoniumQ: I think `default` is preferred so that the destructor is considered "trivial." It should be trivial unless we have a reason to make it non-trivial.

docs/writing_your_plugin.md
charmoniumQ: Non-managed threads are quite a special case, so I don't know if a solution would generalize well.

common/stoplight.hpp
charmoniumQ: Agree, will change

runtime/runtime_impl.hpp
charmoniumQ: Agree, will change

common/stoplight.hpp
charmoniumQ: If we later add an exception, then we would have to change the interface. I think its best to leave the option open unless there is a performance or semantic reason to do so.

e3m3: Agreed about changing the interface. Thinking about this more, any implementation change which adds the possibility of an exception being thrown is a change to the interface, so we could make an argument that forcing these functions to `std::terminate` is more stable (i.e. introduced exceptions must be handled).

runtime/runtime_impl.hpp
charmoniumQ: Yes.@mhuzai, if you are satisfied by this solution, mark conversation as resolved.

e3m3 (approved):

mhuzai (approved):
Looks great!

Issue #256, opened 2021-03-27
install_deps.sh defaults to one thread


mhuzai:
Building with one thread is extremely slow. The user is unaware that the number of threads can be specified unless they look at the source code. The documentation doesn't mention specifying threads either. And it is unlikely someone would run `install_deps.sh --help`.

Minimally, the documentation should say something about this. And maybe the script should ask the user the number of threads.

PR #261, opened 2021-04-05
Hotfix: Added Git tags for `monado_integration` and fixed `clean` for `configs/clean.yaml`

e3m3:
Closes #260.
- Git tags are now used in `ILLIXR/configs/monado.yaml'.

PR #264, opened 2021-04-05
Canonicalize scripts

### LINKED ISSUE #263: CI and install scripts should be standardized and track dependencies
Currently, the install scripts and CI workflow are not standardized, making it difficult to maintain.
Moreover, without minimal dependency tracking, some installations can be broken on client machines.

e3m3:
Closes #263. Depends on #261.

Changes:
- 'install_deps.sh' now shows the help message when run with no arguments
- Updated help message in 'install_deps.sh'
- New 'deps.sh' file tracking dependency directory configurations
- 'Dockerfile' now defaults to 'Release' for cmake installs
- 'Dockerfile' now accepts an 'ACTION' argument for Runner
- 'Dockerfile' now defaults to Ubuntu 20.04
- Manual or interactive ('install_deps.sh') script runs default to 'RelWithDebInfo' for cmake installs
- Removed 'temp_dir' from 'Dockerfile'
- Updated 'GTEST_LOC' to generate from 'deps.sh' in 'common/common.mk'
- New 'install_prompt' utility in 'install_deps.sh' for clearing old installs
- Canonicalized install script structure (fetch -> build -> install -> log)
- Updated CUDA installation detection in 'scripts/install_apt_deps.sh'
- Added dependency info logging and configuration loading (default to '.deps/')
- Added '.deps' to '.gitignore'


e3m3:
Pardon the bash scripts. These prototypes should be moved to Python (or similar) once the build/package interfaces are more mature, and should reflect the results of #104's discussions.

e3m3:
Closes #256.

charmoniumQ (dismissed):
Nix is an obvious priority. This will serve us well until then.

install_deps.sh
```diff
@@ -132,40 +289,88 @@ then
...
+        echo "Found conda installation log '${dep_log_path_conda}' : dir <- '${src_dir_conda}'"
+    else
+        dep_log_path_conda="${dep_log_path}"
+        src_dir_conda="${parent_dir_conda}/${dep_name_conda}"
+        print_warning "Installation log for conda not found at '${dep_log_path_conda}'."
+        echo "Assuming : dir <- '${src_dir_conda}'"
+    fi
+
+    ## Prompt for path to conda environment configuration
+    prompt_value env_config_path "${env_config_path}"
```
charmoniumQ, line 361: I do think we can safely assume that paths within our repo are owned/managed by us. No need to prompt for the env.

install_deps.sh
```diff
@@ -132,40 +289,88 @@ then
...
+    ## Check for a previous conda installation
+    detect_dependency "${dep_name_conda}" "${deps_log_dir}"
+    if [ "$?" -eq 0 ]; then
+        dep_log_path_conda="${dep_log_path}"
+        src_dir_conda="${src_dir}"
+        echo "Found conda installation log '${dep_log_path_conda}' : dir <- '${src_dir_conda}'"
+    else
+        dep_log_path_conda="${dep_log_path}"
+        src_dir_conda="${parent_dir_conda}/${dep_name_conda}"
+        print_warning "Installation log for conda not found at '${dep_log_path_conda}'."
```
charmoniumQ, line 356: Does conda not found imply we should quit?

e3m3: Yeah I think it would be safe to abort here. The script is written to try for conda anyway below as a stop-gap (`detect_dependency` is only looking for a log entry in `.deps/`, but conda can still have been installed via a previous version of the script, or via the system).

install_deps.sh
e3m3: The prompts are just in case the user wants to override our default (like in the old script version case mentioned above). If this will add confusion then I agree we can remove it.

e3m3: The target path was never directly configurable so this is probably not a concern for this variable.

charmoniumQ (approved):

charmoniumQ (dismissed):

charmoniumQ (dismissed):

charmoniumQ (approved):

deps.sh
```diff
@@ -0,0 +1,94 @@
...
+export parent_dir_opencv="${opt_dir}"
+export dep_prompt_opencv="Install OpenCV from source"
+export dep_ver_opencv="3.4.6-instrumented"
+
+## Eigen
+export dep_name_eigen="eigen"
+export script_path_eigen="scripts/install_eigen.sh"
+export parent_dir_eigen="${opt_dir}"
+export dep_prompt_eigen="Install eigen (Eigen3) from source"
+export dep_ver_eigen="3.3.7"
```
mhuzai, line 31: Not 3.3.9?

scripts/install_vulkan_headers.sh
```diff
@@ -1,8 +1,54 @@
...
+### Fetch, build and install ###
+
+## Fetch
+git clone --depth 1 --branch "${dep_ver}" "${repo_url}" "${src_dir}"
+
+## Build
 cmake \
-	-S "${temp_dir}/Vulkan-Headers" \
-	-B "${temp_dir}/Vulkan-Headers/build" \
-	-D CMAKE_INSTALL_PREFIX=install
```
mhuzai, line 7: Wait. Where was this being installed previously?

scripts/install_vulkan_headers.sh
```diff
@@ -1,8 +1,54 @@
...
+    src_dir="${parent_dir_vulkan}/${dep_name_vulkan}"
+fi
+
+if [ -z "${dep_ver}" ]; then
+    dep_ver="${dep_ver_vulkan}"
+fi
+
+repo_url="https://github.com/KhronosGroup/Vulkan-Headers.git"
+build_dir="${src_dir}/build"
+#prefix_dir="install" # Override default
```
mhuzai, line 34: Required?

scripts/install_opencv.sh
```diff
@@ -1,32 +1,87 @@
...
+fi
+
+dep_ver_extra="3.4.6"
+repo_url="https://github.com/ILLIXR/opencv"
+repo_url_extra="https://github.com/opencv/opencv_contrib"
+opencv_dir_extra="${opt_dir}/opencv_contrib"
+build_dir="${src_dir}/build"
+file_cleanup_list="a.out cmake_hdf5_test.o"
+
+flags_cuda="-allow-unsupported-compiler"
```
mhuzai, line 37: Sorry, forgot to update you earlier. We no longer require CUDA when installing OpenCV (good riddance). All of the CUDA related stuff in the OpenCV install can be safely removed.

mhuzai (commented):
What a reorg!

scripts/install_vulkan_headers.sh
e3m3: To the `install` directory at the root of the ILLIXR project. Not sure why this was chosen, but hopefully changing it to the usual `/usr/local` is safe.

deps.sh
e3m3: Version `3.3.7` matches the current version installed by Ubuntu 20.04 LTS. Should it be `3.3.9` for everyone? I forgot which plugin requires it. KinectFusion?

Dockerfile
```diff
@@ -1,61 +1,63 @@
...
 ENV DEBIAN_FRONTEND=noninteractive
 ENV TZ=Europe/Moscow
 ENV CC=clang-10
 ENV CXX=clang++-10
-ENV temp_dir /tmp/ILLIXR
-ENV opt_dir /opt/ILLIXR
-ENV illixr_nproc ${JOBS}
+ENV opt_dir=/opt/ILLIXR
+ENV illixr_nproc=${JOBS}
+ENV build_type=Release
```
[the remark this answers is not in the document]
e3m3: `Release` is needed for CI due to storage provisioning. Some of the deps get very large with `RelWithDebInfo`.

scripts/install_vulkan_headers.sh
e3m3: Nope, but left it in because of the uncertainty about comment above:
> Wait. Where was this being installed previously?

charmoniumQ (approved):

charmoniumQ (dismissed):

deps.sh
mhuzai: Kimera needs 3.3.9 otherwise we have to use Kimera's built-in Eigen. Isn't that the entire fix that you pushed a few weeks ago?

Dockerfile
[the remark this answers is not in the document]
mhuzai: Fair enough. But out of curiosity, I thought only GTSAM needed to be in Release, and the other dependencies could be in RelWithDebInfo.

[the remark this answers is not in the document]
e3m3: Setting `Release` for GTSAM and `RelWithDebInfo` for (some) of the others did work/fit in CI. However, this solution is ad-hoc, as we need to manually configure each plugin's build type so that we get an image that is small enough to fit. I thought using `Release` for all CI for the time being would be the easiest to maintain.

charmoniumQ (approved):

mhuzai:
Why are boost and clang being built from source now? And before I review this again, can you summarize the changes from last time?

e3m3:
> Why are boost and clang being built from source now? And before I review this again, can you summarize the changes from last time?

Scripts for compiling clang and boost have been added, but they are not enabled. These will be helpful as both of these dependencies change between Ubuntu versions. 3rd party binaries could be used here, but binaries are not available for all supported platforms (since [clang-10](https://releases.llvm.org/download.html#10.0.1) is relatively older). If someone needs clang-10 or boost, these scripts will get them started. Boost in particular is disabled because of GTSAM (see file header `scripts/install_boost.sh`.

For changes, it's helpful to work backwards from the bottom of the commit filter in the `Files changed` tab. You can also choose to see changes since your last review (I think as long as there hasn't been a forced update).

Functionally, the rest of the new changes are for the correctness of the script output, comments, dependency detection bugs, and bugs finding conda environment. In terms of what's actually installed, no major changes I think since your last review (Cuda 11, realsense packages, etc). See the commit logs for more info.

charmoniumQ (dismissed):

charmoniumQ (dismissed):

charmoniumQ (dismissed):

mhuzai (approved):

charmoniumQ (approved):

PR #266, opened 2021-04-08
Feature: Offloaded SLAM

K-Wu:
Commit changes before mid term progress report.
please squash the commits to make it single commit to the branch.


K-Wu:
@JeffreyZh4ng Can you please kindly review this?

JeffreyZh4ng:
The commits in this branch enable ILLIXR to communicate with MXRE. MXRE will return the output of Kimera and ILLIXR will use that pose. Currently the IMU integrator and pose prediction are disabled as of the latest commit because the rate at which we are receiving poses back from MXRE is slower than the frequency of camera frames being read in. This causes the IMU integrator to be backed up and accumulate IMU values. 

The next steps for this branch:
- Drop stale camera frames
- Debug negative round trip time between ILLIXR and MXRE

Issue #259, opened 2021-04-05
Expand `realsense` camera support to T26x


ltstein:
As I understand it, ILLIXR currently supports the D435i camera. I would like to target a broader range of the Realsense hardware, starting with the T26x variants although they do not support as high a rate of IMU data. Primary motivation is most Project North Star headsets have this as the default tracking solution so this would enable those users to leverage the ILLIXR framework for future development.   First look, it appears it would mostly be allowing alternate values for the `constexpr`s in `plugin.cpp` but I don't know if a configurable plugin or completely separate plugin would be preferred considering the whole framework.

mhuzai:
Thanks for looking into this!

Configurable plugin would be the way to go because most of the code is the same, and it should hopefully just be a matter of detecting which camera is attached and then setting the parameters correctly. @Hyjale actually did run a T265 at some point, so he can shed some light on this.

Hyjale:
Hi, 

I actually didn't get the chance to run with the T265. I would recommend creating a separate plugin for the T26x sensors since those have pose data built-in. 

For reference: https://github.com/IntelRealSense/librealsense/blob/master/examples/pose/rs-pose.cpp

mhuzai:
This can be easily and cleanly done within one plugin. We can query the attached camera by using [query_devices()](https://intelrealsense.github.io/librealsense/doxygen/classrs2_1_1context.html#a0e0bfd3451cab5a32ab78f8037a4d793), and then configure it appropriately. [This page](https://github.com/IntelRealSense/librealsense/tree/master/examples/sensor-control) provides good examples of querying device information.

Also, ILLIXR's camera/IMU plugins do **not** publish poses. They only publish RGB/grayscale images and IMU values, and optionally depth images. So we shouldn't be worrying about pose in this plugin.

ltstein:
> This can be easily and cleanly done within one plugin. We can query the attached camera by using [query_devices()](https://intelrealsense.github.io/librealsense/doxygen/classrs2_1_1context.html#a0e0bfd3451cab5a32ab78f8037a4d793), and then configure it appropriately. [This page](https://github.com/IntelRealSense/librealsense/tree/master/examples/sensor-control) provides good examples of querying device information.
> 
> Also, ILLIXR's camera/IMU plugins do **not** publish poses. They only publish RGB/grayscale images and IMU values, and optionally depth images. So we shouldn't be worrying about pose in this plugin.

[...]

PR #269, opened 2021-04-11
Expand `realsense` supported cameras

ltstein:
Resolve issue #259 by enabling configuration of which realsense device to use.

### DIFF
--- a/realsense/plugin.cpp
+++ b/realsense/plugin.cpp
@@ -42,13 +42,47 @@ public:
         , sb{pb->lookup_impl<switchboard>()}
         , _m_imu_cam{sb->get_writer<imu_cam_type>("imu_cam")}
         , _m_rgb_depth{sb->get_writer<rgb_depth_type>("rgb_depth")}
-        , realsense_cam{ILLIXR::getenv_or("REALSENSE_CAM", "D4XX")}
+        , realsense_cam{ILLIXR::getenv_or("REALSENSE_CAM", "auto")}
         {      
             cfg.disable_all_streams();
             configure_camera();
         }
+    void auto_select_cam(){
+        rs2::context ctx;
+        rs2::device_list devices = ctx.query_devices();
+        for (rs2::device device : devices) {
+            std::string name = "Unknown Device";
+            if (device.supports(RS2_CAMERA_INFO_NAME)) {
+                name = device.get_info(RS2_CAMERA_INFO_NAME);
+                if (name.find("D4") != std::string::npos){
+                    std::cout << "Found D4XX Device, Checking for supported sensors" << std::endl;
+                    std::vector<rs2::sensor> sensors = device.query_sensors();
+                    for (rs2::sensor sensor : sensors){
+                        if (sensor.supports(RS2_CAMERA_INFO_NAME)){
+                            std::string sensor_name = sensor.get_info(RS2_CAMERA_INFO_NAME);
+                            if (sensor_name.find("Motion Module") != std::string::npos){
+                                std::cout << "Setting realsense_cam: D4XX" << std::endl;
+                                realsense_cam = "D4XX";
+                                break;
+                            }
+                        }
+                    }
+                    if (realsense_cam.compare("D4XX") == 0){
+                        break;
+                    }
+                } 
+                else if (name.find("T26") != std::string::npos){
+                    realsense_cam = "T26X";
+                    std::cout << "Setting realsense_cam: T26X" << std::endl;
+                }
+            }
+        }
+    }
     void configure_camera()
     {
+        if (realsense_cam.compare("auto") == 0) {
+            auto_select_cam();
+        }
         if (realsense_cam.compare("T26X") == 0){
             cfg.enable_stream(RS2_STREAM_ACCEL, RS2_FORMAT_MOTION_XYZ32F); // adjustable to 0, 63 (default), 250 hz
             cfg.enable_stream(RS2_STREAM_GYRO, RS2_FORMAT_MOTION_XYZ32F); // adjustable set to 0, 200 (default), 400 hz
--- a/realsense/plugin.cpp
+++ b/realsense/plugin.cpp
@@ -17,24 +17,6 @@ static constexpr int ACCEL_RATE_D4XX = 250; // 63 or 250
 
 static constexpr int IMAGE_WIDTH_T26X = 848;
 static constexpr int IMAGE_HEIGHT_T26X = 800;
-typedef struct {
-    cv::Mat img0;
-	cv::Mat img1;
-	cv::Mat rgb;
-	cv::Mat depth;
-    int iteration;
-} cam_type;
-
-typedef enum {
-    UNSUPPORTED,
-    D4XXI,
-    T26X
-} cam_enum;
-
-typedef struct {
-	rs2_vector* accel_data;
-	int iteration;
-} accel_type;
 
 class realsense : public plugin {
 public:
@@ -48,106 +30,6 @@ public:
             cfg.disable_all_streams();
             configure_camera();
         }
-    
-    void find_supported_devices(rs2::device_list devices){
-        for (rs2::device device : devices) {
-            if (device.supports(RS2_CAMERA_INFO_PRODUCT_LINE)){
-                std::string product_line = device.get_info(RS2_CAMERA_INFO_PRODUCT_LINE); 
-                #ifndef NDEBUG
-                    std::cout << "Found Product Line: " << product_line << std::endl;
-                #endif
-                if (product_line == "D400"){
-                    #ifndef NDEBUG
-                        std::cout << "Checking for supported streams" << std::endl;
-                    #endif
-                    std::vector<rs2::sensor> sensors = device.query_sensors();
-                    for (rs2::sensor sensor : sensors){
-                        std::vector<rs2::stream_profile> stream_profiles = sensor.get_stream_profiles();
-                        //Currently, all D4XX cameras provide infrared, RGB, and depth, so we only need to check for accel and gyro
-                        for (auto&& sp : stream_profiles)
-                        {
-                            if (sp.stream_type() == RS2_STREAM_GYRO){
-                                gyro_found = true;
-                            };
-                            if (sp.stream_type() == RS2_STREAM_ACCEL){
-                                accel_found = true;
-                            }
-                        }
-                    }
-                    if (accel_found && gyro_found){
-                        D4XXI_found = true;
-                        #ifndef NDEBUG
-                            std::cout << "Supported D4XX found!" << std::endl;
-                        #endif
-                    }
-                } 
-                else if (product_line == "T200"){
-                    T26X_found = true;
-                    #ifndef NDEBUG
-                        std::cout << "T26X found! " << std::endl;
-                    #endif
-                }
-            }
-            
-        }
-        if (!T26X_found && !D4XXI_found){
-            std::cout << "No supported Realsense device detected!" << std::endl;
-        }
-    }
-
-    void configure_camera()
-    {
-        rs2::context ctx;
-        rs2::device_list devices = ctx.query_devices();
-        find_supported_devices(devices);
-        if (realsense_cam.compare("auto") == 0) {
-            if (D4XXI_found){
-                cam_select = D4XXI;
-                #ifndef NDEBUG
-                    std::cout << "Setting cam_select: D4XX" << std::endl;
-                #endif
-            }
-            else if (T26X_found){
-                cam_select = T26X;
-                #ifndef NDEBUG
-                    std::cout << "Setting cam_select: T26X" << std::endl;
-                #endif
-            }
-        }
-        else if ((realsense_cam.compare("D4XX") == 0) && D4XXI_found){
-            cam_select = D4XXI;
-            #ifndef NDEBUG
-                std::cout << "Setting cam_select: D4XX" << std::endl;
-            #endif
-        }
-        else if ((realsense_cam.compare("T26X") == 0) && T26X_found){
-            cam_select = T26X;
-            #ifndef NDEBUG
-                std::cout << "Setting cam_select: T26X" << std::endl;
-            #endif
-        }
-        if (cam_select == UNSUPPORTED){
-            ILLIXR::abort("Supported Realsense device NOT found!");
-        }
-        if (cam_select == T26X){
-            //T26X series has fixed options for accel rate, gyro rate, fisheye resolution, and FPS
-            cfg.enable_stream(RS2_STREAM_ACCEL, RS2_FORMAT_MOTION_XYZ32F); // 62 Hz
-            cfg.enable_stream(RS2_STREAM_GYRO, RS2_FORMAT_MOTION_XYZ32F); // 200 Hz
-            cfg.enable_stream(RS2_STREAM_FISHEYE, 1, RS2_FORMAT_Y8); //848x800, 30 FPS
-            cfg.enable_stream(RS2_STREAM_FISHEYE, 2, RS2_FORMAT_Y8); //848x800, 30 FPS
-            profiles = pipe.start(cfg, [&](const rs2::frame& frame) { this->callback(frame); });
-        }
-        else if (cam_select == D4XXI){
-            cfg.enable_stream(RS2_STREAM_ACCEL, RS2_FORMAT_MOTION_XYZ32F, ACCEL_RATE_D4XX); // adjustable to 0, 63 (default), 250 hz
-            cfg.enable_stream(RS2_STREAM_GYRO, RS2_FORMAT_MOTION_XYZ32F, GYRO_RATE_D4XX); // adjustable set to 0, 200 (default), 400 hz
-            cfg.enable_stream(RS2_STREAM_INFRARED, 1, IMAGE_WIDTH_D4XX, IMAGE_HEIGHT_D4XX, RS2_FORMAT_Y8, FPS_D4XX);
-            cfg.enable_stream(RS2_STREAM_INFRARED, 2, IMAGE_WIDTH_D4XX, IMAGE_HEIGHT_D4XX, RS2_FORMAT_Y8, FPS_D4XX);
-            cfg.enable_stream(RS2_STREAM_COLOR, IMAGE_WIDTH_D4XX, IMAGE_HEIGHT_D4XX, RS2_FORMAT_BGR8, FPS_D4XX);
-            cfg.enable_stream(RS2_STREAM_DEPTH, IMAGE_WIDTH_D4XX, IMAGE_HEIGHT_D4XX, RS2_FORMAT_Z16, FPS_D4XX);
-            profiles = pipe.start(cfg, [&](const rs2::frame& frame) { this->callback(frame); });
-            profiles.get_device().first<rs2::depth_sensor>().set_option(RS2_OPTION_EMITTER_ENABLED, 0.f); // disables IR emitter
-        }
-    }
 
 	void callback(const rs2::frame& frame)
         {
@@ -246,7 +128,7 @@ public:
                     
                     // Submit to switchboard
                     _m_imu_cam.put(_m_imu_cam.allocate<imu_cam_type>(
-                        imu_cam_type{
+                        {
                             imu_time_point,
                             av,
                             la,
@@ -259,7 +141,7 @@ public:
                     if (rgb && depth)
                     {
                         _m_rgb_depth.put(_m_rgb_depth.allocate<rgb_depth_type>(
-                            rgb_depth_type{
+                            {
                                 rgb,
                                 depth,
                                 imu_time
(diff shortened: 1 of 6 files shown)


realsense/plugin.cpp
```diff
@@ -9,20 +9,42 @@
 
 using namespace ILLIXR;
 
+#define T26X
```
ltstein, line 12: Is there a preferred method for getting configuration information from config yaml to the plugin code? What I have found so far seems to use env vars set by the runner script, such as 
config.yaml
`enable_alignment: False`
main.py
`enable_offload_flag = config["enable_offload"]`
`enable_offload{ILLIXR::str_to_bool(ILLIXR::getenv_or("ILLIXR_OFFLOAD_ENABLE", "False"))}` and
`if (enable_offload) ...`
plugin.cpp

ltstein:
This initial commit is just to test. I noticed that the axes seemed to be off by 90 degrees compared to the rotations I would expect when using a realsense device, assuming it is mounted facing the users front, is there configuration somewhere that I missed?

charmoniumQ:
It's great to have you contributing to ILLIXR :)

Just a heads up -- We are going to merge #32 some day (hopefully next week?). This will likely cause a lot of conflicts because we changed the Switchboard interface; No major semantic changes though, so most of the code will be easy to adapt.

realsense/plugin.cpp
e3m3: Currently, the best way is to follow the `offload_data` plugin example as you've found. This requires a few steps:
- [ ] Add your variable to `runner/config_schema.yaml` (like [`enable_offload`](https://github.com/ILLIXR/ILLIXR/blob/508fd83e9361e8b011c598787892ffe1e982592b/runner/config_schema.yaml#L128)).
- [ ] In each action ([`load_native`](https://github.com/ILLIXR/ILLIXR/blob/508fd83e9361e8b011c598787892ffe1e982592b/runner/runner/main.py#L84), [`load_tests`](https://github.com/ILLIXR/ILLIXR/blob/508fd83e9361e8b011c598787892ffe1e982592b/runner/runner/main.py#L140), [`load_monado`](https://github.com/ILLIXR/ILLIXR/blob/508fd83e9361e8b011c598787892ffe1e982592b/runner/runner/main.py#L167)):
    - [ ] Load the value from the config (like [`enable_offload`](https://github.com/ILLIXR/ILLIXR/blob/508fd83e9361e8b011c598787892ffe1e982592b/runner/runner/main.py#L144))
    - [ ] Optionally process the value
    - [ ] Set the value in the environment (like [`ILLIXR_ENABLE_OFFLOAD`](https://github.com/ILLIXR/ILLIXR/blob/508fd83e9361e8b011c598787892ffe1e982592b/runner/runner/main.py#L158))
    - [ ] Read, process, and set your value within your client code (like [`ILLIXR_ENABLE_OFFLOAD`](https://github.com/ILLIXR/ILLIXR/blob/508fd83e9361e8b011c598787892ffe1e982592b/offload_data/plugin.cpp#L33))

For your use case, it looks like you would like to do a few things:
1. Set the camera type,
1. Set some parameters based on the camera type, and
1. Do some logic based on the camera type.

Since are passing in parameters to the program using environment variables, we have to be careful with statically determined variables and compilation:
1. The camera type will probably have to be string environment variable that we can set in the relevant files in `configs`. This value will have to be loaded in the constructor block of the plugin because the environment variables are set _after_ compilation.
2. Either in the constructor, or in the body in the context where each variable is used, the camera property should be read to determine the values.
3. In the body where logic is muxed based on the camera property, the logic needs to be handled dynamically (without static compilation guards).

This is a tedious process we are hoping to streamline with the #198 patch, but we have some other patches on the critical path first.

ltstein: @e3m3 Thanks for the detailed reply, I'll probably get this updated later today.

mhuzai: I will not have time to review this properly before next week, but why do we need to set configs/parameters/defines at all for this? Can we not automatically detect which camera has been plugged in using [query_devices](https://github.com/IntelRealSense/librealsense/blob/master/examples/sensor-control/api_how_to.h#L29) and then follow its code path?

ltstein: > I will not have time to review this properly before next week, but why do we need to set configs/parameters/defines at all for this? Can we not automatically detect which camera has been plugged in using [query_devices](https://github.com/IntelRealSense/librealsense/blob/master/examples/sensor-control/api_how_to.h#L29) and then follow its code path?

We certainly could do that but my reasoning was that the user may have multiple cameras attached to the system and would want to specify which model to use. Once #198 is added, it would be easier to refactor and allow the user to specify a serial number, or configure image size/FPS, accel/gyro rates, etc. for their use case. Right now, the goal would be to make it relatively simple to switch between the T26X and the currently supported RGB-D series streams. 
Open to suggestions  or other ideas on this.

realsense/plugin.cpp
```diff
@@ -110,43 +182,72 @@ class realsense : public plugin {
...
-                        img0 = cam_type_.img0;
-                        img1 = cam_type_.img1;
-                        rgb = cam_type_.rgb;
-                        depth = cam_type_.depth;
+                     // Images
+                    if (realsense_cam.compare("D4XX") == 0){
+                        std::optional<cv::Mat *> img0 = std::nullopt;
+                        std::optional<cv::Mat *> img1 = std::nullopt;
+                        std::optional<cv::Mat *> rgb = std::nullopt;
+                        std::optional<cv::Mat *> depth = std::nullopt;
```
e3m3, line 190: Having a pointer type inside of a `std::optional` is generally redundant, but the current interface uses `cv::Mat*` for the Switchboard data type. As a heads up, this changes in the upcoming [#32 patch](https://github.com/ILLIXR/ILLIXR/blob/865edb7cfe2e13dce2f0972cd3840c61889afeae/common/data_format.hpp#L33).

realsense/plugin.cpp
```diff
@@ -9,20 +9,27 @@
...
+    cv::Mat* rgb;
+    cv::Mat* depth;
+    int iteration;
+} cam_type_D4XX;
 
+typedef struct {
+    cv::Mat* img0;
+    cv::Mat* img1;
+    int iteration;
+} cam_type_T26X;
```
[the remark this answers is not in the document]
e3m3: Note that this suggestion has a lot of parallels to what's already going on in [`common/data_format.hpp`](https://github.com/ltstein/ILLIXR/blob/2f89db2595b68e6da53d48ac9db4fc47e7ed3741/common/data_format.hpp#L29), so it could be worth constructing these data types directly without the auxiliary types.

realsense/plugin.cpp
```diff
@@ -110,43 +182,72 @@ class realsense : public plugin {
...
+                                la,
+                                img0,
+                                img1,
                                 imu_time,
                             });
+            
+                        auto imu_integrator_params = new imu_integrator_seq{
+                            .seq = static_cast<int>(++_imu_integrator_seq),
+                        };
+                        _m_imu_integrator->put(imu_integrator_params);
```
e3m3, line 249: Another heads up: The `imu_integrator_seq` topic is removed from the plugin in [#32](https://github.com/ILLIXR/ILLIXR/blob/865edb7cfe2e13dce2f0972cd3840c61889afeae/realsense/plugin.cpp#L154).

e3m3 (commented):
This is shaping up well! I like the idea of having an `auto` option in addition to being able to manually specify the camera in the config. Beyond mulling over refactoring the design based on the other comments, the main thing still missing is setting the `realsense_cam` values in the `configs/` files.

Also, I appreciate the nicely organized commits! It does help with filtering the files when reviewing. No need to strictly enforce this as we squash commits when merging to `master`, but thought I would point it out :smile:.

ltstein:
> This is shaping up well! I like the idea of having an `auto` option in addition to being able to manually specify the camera in the config. Beyond mulling over refactoring the design based on the other comments, the main thing still missing is setting the `realsense_cam` values in the `configs/` files.
> 
> Also, I appreciate the nicely organized commits! It does help with filtering the files when reviewing. No need to strictly enforce this as we squash commits when merging to `master`, but thought I would point it out smile.

Planning on refactoring several of the things you mentioned here so thanks for your input, I  will be watching for #32 to get things updated as well. 
Regarding setting `realsense_cam` in config, I set a default value within the config_schema. Is it preferred to have a default value set in each separate config even if this plugin is not being used?

e3m3:
> Regarding setting realsense_cam in config, I set a default value within the config_schema. Is it preferred to have a default value set in each separate config even if this plugin is not being used?
 
Oh I missed the default `auto` value there. Yep, that works!

realsense/plugin.cpp
```diff
@@ -36,17 +43,65 @@ class realsense : public plugin {
...
+            cfg.enable_stream(RS2_STREAM_FISHEYE, 2, RS2_FORMAT_Y8);
+            profiles = pipe.start(cfg, [&](const rs2::frame& frame) { this->callback(frame); });
+        }
+        else if (realsense_cam.compare("D4XX") == 0){
+            cfg.enable_stream(RS2_STREAM_ACCEL, RS2_FORMAT_MOTION_XYZ32F, ACCEL_RATE_D4XX); // adjustable to 0, 63 (default), 250 hz
+            cfg.enable_stream(RS2_STREAM_GYRO, RS2_FORMAT_MOTION_XYZ32F, GYRO_RATE_D4XX); // adjustable set to 0, 200 (default), 400 hz
+            cfg.enable_stream(RS2_STREAM_INFRARED, 1, IMAGE_WIDTH_D4XX, IMAGE_HEIGHT_D4XX, RS2_FORMAT_Y8, FPS_D4XX);
+            cfg.enable_stream(RS2_STREAM_INFRARED, 2, IMAGE_WIDTH_D4XX, IMAGE_HEIGHT_D4XX, RS2_FORMAT_Y8, FPS_D4XX);
+            cfg.enable_stream(RS2_STREAM_COLOR, IMAGE_WIDTH_D4XX, IMAGE_HEIGHT_D4XX, RS2_FORMAT_BGR8, FPS_D4XX);
+            cfg.enable_stream(RS2_STREAM_DEPTH, IMAGE_WIDTH_D4XX, IMAGE_HEIGHT_D4XX, RS2_FORMAT_Z16, FPS_D4XX);
```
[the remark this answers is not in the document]
e3m3: Also, if it helps to simplify the logic, we can introduce a generic intermediate value for shared options:
```cpp
switch_case (realsense_cam) {
    case RS_CAM_D4XX: 
        image_width = IMAGE_WIDTH_DX44;
        break;
    case RS_CAM_T26X:
        image_width = IMAGE_WIDTH_T26X:
        break;
    default:
        throw std::runtime_error("Unexpected camera value for 'realsense_cam'");
}
```
If there area a lot of shared camera properties, we can use a properties `struct` to organize them.

e3m3:
Another idea I forgot to mention: In the case where the user has specified a non-`auto` option for the camera, we can use `device_query` to verify the selected camera is indeed present.

realsense/plugin.cpp
```diff
@@ -55,26 +110,43 @@ class realsense : public plugin {
...
-                depth.convertTo(converted_depth, CV_32FC1, depth_scale * 1000.f);
-                cam_type_ = cam_type{
-                    new cv::Mat{ir_left},
-                    new cv::Mat{ir_right},
-                    new cv::Mat{rgb},
-                    new cv::Mat{converted_depth},
-                    iteration_cam,
-                };
-                iteration_cam++;
+            if (realsense_cam.compare("D4XX") == 0){
```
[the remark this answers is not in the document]
ltstein: Switched to using an enum for this

realsense/plugin.cpp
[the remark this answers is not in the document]
ltstein: I switched to detecting supported hardware using the streams they provide. It would be possible to search for an array of requested streams or data types if that would be useful in the future.

realsense/plugin.cpp
```diff
@@ -35,17 +49,95 @@ class realsense : public plugin {
...
+        else if ((realsense_cam.compare("D4XX") == 0) && D4XXI_found){
+            cam_select = D4XXI;
+            std::cout << "Setting cam_select: D4XX" << std::endl;
+        }
+        else if ((realsense_cam.compare("T26X") == 0) && T26X_found){
+            cam_select = T26X;
+            std::cout << "Setting cam_select: T26X" << std::endl;
+        }
+        if (cam_select == UNSUPPORTED){
+            std::cout << "Supported device NOT found!" << std::endl;
```
charmoniumQ, line 120: We should `abort()` here.

realsense/plugin.cpp
```diff
@@ -109,36 +218,66 @@ class realsense : public plugin {
...
-                        imu_time,
-                    ));
-
-                    if (rgb && depth)
-                    {
-                        _m_rgb_depth.put(_m_rgb_depth.allocate<rgb_depth_type>(
-                            rgb,
-                            depth,
-                            imu_time,
+                    else if (cam_select == T26X){
```
charmoniumQ, line 260: I think the code can be simplified a bit by declaring all four `std::optional<cv::Mat>` regardless of underlying camera type, but we can check `rgb && depth` and know if there is nothing there. This removes this conditional block and the need for `cam_type_${hardware}`. The earlier in code we "unify" the behavior for underlying hardware the better.

realsense/plugin.cpp
ltstein: Thanks, I wasn't sure what the preference was so I had this basically as a placeholder.

ltstein: Also, do we want to keep these behind a debug flag or is there a logger setup I should pass them to instead of the console?

charmoniumQ: @e3m3, will `abort()` resolve to `std::abort()` or `ILLIXR::abort()`? I forgot...

If a device is found is found, hide it behind `#ifndef NDEBUG`. If no device is found, that should be printed unconditionally.

e3m3: > @e3m3, will abort() resolve to std::abort() or ILLIXR::abort()? I forgot...

I think neither, and the compiler should complain (ambiguous). We can stick to `ILLIXR::abort(...)`.

charmoniumQ (dismissed):

e3m3 (approved):

realsense/plugin.cpp
```diff
@@ -35,17 +43,111 @@ class realsense : public plugin {
...
+                    #endif
+                    std::vector<rs2::sensor> sensors = device.query_sensors();
+                    for (rs2::sensor sensor : sensors){
+                        std::vector<rs2::stream_profile> stream_profiles = sensor.get_stream_profiles();
+                        //Currently, all D4XX cameras provide infrared, RGB, and depth, so we only need to check for accel and gyro
+                        for (auto&& sp : stream_profiles)
+                        {
+                            if (sp.stream_type() == RS2_STREAM_GYRO){
+                                gyro_found = true;
+                            };
```
mhuzai, line 71: Trailing `;`

realsense/plugin.cpp
```diff
@@ -35,17 +43,111 @@ class realsense : public plugin {
...
-            cfg.enable_stream(RS2_STREAM_ACCEL, RS2_FORMAT_MOTION_XYZ32F, ACCEL_RATE); // adjustable to 0, 63 (default), 250 hz
-            cfg.enable_stream(RS2_STREAM_GYRO, RS2_FORMAT_MOTION_XYZ32F, GYRO_RATE); // adjustable set to 0, 200 (default), 400 hz
-            cfg.enable_stream(RS2_STREAM_INFRARED, 1, IMAGE_WIDTH, IMAGE_HEIGHT, RS2_FORMAT_Y8, FPS);
-            cfg.enable_stream(RS2_STREAM_INFRARED, 2, IMAGE_WIDTH, IMAGE_HEIGHT, RS2_FORMAT_Y8, FPS);
-            cfg.enable_stream(RS2_STREAM_COLOR, IMAGE_WIDTH, IMAGE_HEIGHT, RS2_FORMAT_BGR8, FPS);
-            cfg.enable_stream(RS2_STREAM_DEPTH, IMAGE_WIDTH, IMAGE_HEIGHT, RS2_FORMAT_Z16, FPS);
+            configure_camera();
+        }
+    
+    void find_supported_devices(rs2::device_list devices){
```
mhuzai, line 52: Should be private?

realsense/plugin.cpp
```diff
@@ -35,17 +43,111 @@ class realsense : public plugin {
...
+                std::cout << "Setting cam_select: D4XX" << std::endl;
+            #endif
+        }
+        else if ((realsense_cam.compare("T26X") == 0) && T26X_found){
+            cam_select = T26X;
+            #ifndef NDEBUG
+                std::cout << "Setting cam_select: T26X" << std::endl;
+            #endif
+        }
+        if (cam_select == UNSUPPORTED){
```
mhuzai, line 129: This should really just be an `else`.

realsense/plugin.cpp
```diff
@@ -35,17 +43,111 @@ class realsense : public plugin {
...
+                    #ifndef NDEBUG
+                        std::cout << "T26X found! " << std::endl;
+                    #endif
+                }
+            }
+            
+        }
+        if (!T26X_found && !D4XXI_found){
+            std::cout << "No supported Realsense device detected!" << std::endl;
+        }
```
mhuzai, line 95: Only required during debug.

realsense/plugin.cpp
```diff
@@ -35,17 +43,111 @@ class realsense : public plugin {
...
+        else if (cam_select == D4XXI){
+            cfg.enable_stream(RS2_STREAM_ACCEL, RS2_FORMAT_MOTION_XYZ32F, ACCEL_RATE_D4XX); // adjustable to 0, 63 (default), 250 hz
+            cfg.enable_stream(RS2_STREAM_GYRO, RS2_FORMAT_MOTION_XYZ32F, GYRO_RATE_D4XX); // adjustable set to 0, 200 (default), 400 hz
+            cfg.enable_stream(RS2_STREAM_INFRARED, 1, IMAGE_WIDTH_D4XX, IMAGE_HEIGHT_D4XX, RS2_FORMAT_Y8, FPS_D4XX);
+            cfg.enable_stream(RS2_STREAM_INFRARED, 2, IMAGE_WIDTH_D4XX, IMAGE_HEIGHT_D4XX, RS2_FORMAT_Y8, FPS_D4XX);
+            cfg.enable_stream(RS2_STREAM_COLOR, IMAGE_WIDTH_D4XX, IMAGE_HEIGHT_D4XX, RS2_FORMAT_BGR8, FPS_D4XX);
+            cfg.enable_stream(RS2_STREAM_DEPTH, IMAGE_WIDTH_D4XX, IMAGE_HEIGHT_D4XX, RS2_FORMAT_Z16, FPS_D4XX);
             profiles = pipe.start(cfg, [&](const rs2::frame& frame) { this->callback(frame); });
             profiles.get_device().first<rs2::depth_sensor>().set_option(RS2_OPTION_EMITTER_ENABLED, 0.f); // disables IR emitter
```
mhuzai, line 148: I had never noticed this before. Please add a comment saying that disabling infrared allows us to use the stereo images for SLAM, but negatively affects depth quality, especially in low light conditions.

realsense/plugin.cpp
```diff
@@ -35,17 +43,111 @@ class realsense : public plugin {
...
+        }
+    }
+
+    void configure_camera()
+    {
+        rs2::context ctx;
+        rs2::device_list devices = ctx.query_devices();
+        find_supported_devices(devices);
+        if (realsense_cam.compare("auto") == 0) {
+            if (D4XXI_found){
```
mhuzai, line 104: I'm assuming this is written with the assumption that only one camera will be connected to the system at any given time, and if two are connected, the D4XX will be given priority over the T26X. I agree with the design choice -- just add a comment making it clear.

realsense/plugin.cpp
```diff
@@ -160,12 +285,21 @@ class realsense : public plugin {
 	rs2_vector gyro_data;
 	rs2_vector accel_data;
 
+    
 	cam_type cam_type_;
+    cam_enum cam_select{UNSUPPORTED};
+    bool gyro_found{false};
+    bool accel_found{false};
```
mhuzai, line 292: Should be local to `find_supported_devices`.

realsense/plugin.cpp
```diff
@@ -9,20 +9,28 @@
...
+	cv::Mat depth;
+    int iteration;
 } cam_type;
 
+typedef enum {
+    UNSUPPORTED,
+    D4XXI,
+    T26X
+} cam_enum;
+
```
mhuzai, line 33: Should these be inside the class?

realsense/plugin.cpp
```diff
@@ -35,17 +43,111 @@ class realsense : public plugin {
...
+                }
+            }
+            
+        }
+        if (!T26X_found && !D4XXI_found){
+            std::cout << "No supported Realsense device detected!" << std::endl;
+        }
+    }
+
+    void configure_camera()
```
mhuzai, line 98: Private?

realsense/plugin.cpp
```diff
@@ -122,23 +243,27 @@ class realsense : public plugin {
...
+                    
                     // Submit to switchboard
                     _m_imu_cam.put(_m_imu_cam.allocate<imu_cam_type>(
-                        imu_time_point,
-                        av,
-                        la,
-                        img0,
-                        img1,
-                        imu_time,
+                        imu_cam_type{
```
mhuzai, line 249: We can just forward args directly.

realsense/plugin.cpp
```diff
@@ -122,23 +243,27 @@ class realsense : public plugin {
...
                     ));
-
+                    
                     if (rgb && depth)
                     {
                         _m_rgb_depth.put(_m_rgb_depth.allocate<rgb_depth_type>(
-                            rgb,
-                            depth,
-                            imu_time,
+                            rgb_depth_type{
```
mhuzai, line 262: We can forward args directly.

mhuzai (commented):
Thanks again for working on this, especially the auto detection feature. I left a few comments.

realsense/plugin.cpp
charmoniumQ: I think this print should be there always because it signifies a fatal error.

mhuzai: That was my initial reaction too, but this function only finds devices. `configure_camera` does the error handling, and aborts on line 130 if no device is found. I think it's a clean solution, but open to suggestions.

realsense/plugin.cpp
ltstein: @mhuzai  So,I had tried this but removing the type I end up with compilation errors

> /home/lincoln/workspace/ILLIXR/realsense/plugin.cpp:130:47: error: no matching member function for call to 'allocate'
                      _m_imu_cam.put(_m_imu_cam.allocate<imu_cam_type>(
                                     ~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~
  /home/lincoln/workspace/ILLIXR/realsense/common/switchboard.hpp:471:29: note: candidate function template not viable: no known conversion from 'ILLIXR::time_type' (aka 'time_point<std::chrono::_V2::system_clock, duration<long, ratio<1, 1000000000> > >') to 'ILLIXR::imu_cam_type' for 1st argument
          ptr<specific_event> allocate(Args&&... args) {

e3m3 (approved):

PR #274, opened 2021-04-20
Implemented multi-CI and refactored configurations

### LINKED ISSUE #270: CI should build and run all plugins
Currently, we only receive feedback for runs on CI using `configs/ci.yaml`. Some existing flows cannot run under CI yet (e.g. `configs/native.yaml`). Some existing flows cannot build under CI yet (e.g. OpenCV with CUDA enabled). For flows that build but cannot run in CI, we should make a build-only config that includes the modules and compiles the project.

The following plugins are currently built and run:
- `configs/ci.yaml`:
    - `timewarp_gl`
    - `audio`
    - `offline_imu_cam`
    - `pose_prediction`
    - `ground_truth_slam`
    - `gldemo`
    - `Kimera-VIO`
    - `gtsam_integrator`

Minimally, we need the following flows to run in CI:
- `configs/ci/`:
    - `run-solo.yaml`:
        - `offline_imu_cam.yaml`
        - `pose_prediction.yaml`
        - `ground_truth_slam.yaml`
        - `kimera-vio.yaml`
        - `openvins.yaml`
        - `rk4_integrator.yaml`
        - `gtsam_integrator.yaml`
        - `offload_data.yaml`
    - `build-only.yaml`:
        - `audio.yaml` # May move to `run-solo` with hotfix
        - `debugview.yaml`
        - `gldemo.yaml`
        - `timewarp_gl.yaml`
    - `no-build.yaml`:
        - `hotlab.yaml`
        - `zed.yaml` # May move to `build-only`
        - `realsense.yaml` # May move to `build-only`
        - `opencv-with-cuda.yaml` # Not a plugin
        - `monado_integration.yaml` # May move to `build-only`
        - `monado_openxr_app.yaml` # May move to `build-only`
        - `pose_lookup.yaml` # Will move to `run-solo` with hotfix (#32)
    - `flows/`:
        - `kimera-gtsam.yaml`:
            - `timewarp_gl`
            - `audio`
            - `offline_imu_cam`
            - `pose_prediction`
            - `ground_truth_slam`
            - `gldemo`
            - `Kimera-VIO`
            - `gtsam_integrator`
            - `offload_data`
        - `kimera-rk4.yaml`:
            - `timewarp_gl`
            - `audio`
            - `offline_imu_cam`
            - `pose_prediction`
            - `ground_truth_slam`
            - `gldemo`
            - `Kimera-VIO`
            - `rk4_integrator`
            - `offload_data`
        - `openvins-gtsam.yaml`:
            - `timewarp_gl`
            - `audio`
            - `offline_imu_cam`
            - `pose_prediction`
            - `ground_truth_slam`
            - `gldemo`
            - `OpenVINS`
            - `gtsam_integrator`
            - `offload_data`
        - `openvins-rk4.yaml`:
            - `timewarp_gl`
            - `audio`
            - `offline_imu_cam`
            - `pose_prediction`
            - `ground_truth_slam`
            - `gldemo`
            - `OpenVINS`
            - `rk4_integrator`
            - `offload_data`
        - `lookup.yaml`:
            - `timewarp_gl`
            - `audio`
            - `offline_imu_cam`
            - `pose_prediction`
            - `gldemo`
            - `pose_lookup`
            - `rk4_integrator`
            - `offload_data`

e3m3:
Closes #270.

Pushing draft early in case this is useful for ongoing experiments. Needs documentation updates, but hopefully the given structure is easy enough to follow by example.

Todo:
- [x] Documentation
- [x] Updated CI classes for failing plugins (dependent on merges/hotfixes)


docs/glossary.md
```diff
@@ -24,45 +24,66 @@ A modular component that can be detected and enabled for use by an ILLIXR applic
...
+ILLIXR also implements a [_Monado_][58] runtime [_translation plugin_][75].
 
 For a list of supported plugins and their details, see the [ILLIXR Plugins][68] page.
 For instructions for how to modify or write your own plugins, see the [Modifying a Plugin][60]
     and [Writing Your Plugin][61] pages.
 
-See the [_Plugin_ API documentation][53].
+-   **Plugin Group**:
+    A _plugin group_ is a collection of plugins to be loaded and executed by a [_flow_][50]
+        launched by from a [_Runner_][55] configuration.
```
mhuzai, line 35: Delete 'from'.

docs/writing_your_plugin.md
```diff
@@ -98,34 +98,61 @@ You can also replace any existing functionality this way.
...
-        remove that component from the config before running the new component.
+    Move to your copy of ILLIXR repo and update `ILLIXR/configs/native.yaml`.
+    If the new plugin is the same type (i.e., publishes the same topics)
+        as one of the other components you will need to remove that component
+        from the config before running the new component.
     For example, if the new component is a SLAM then the old SLAM needs to be removed from
         the config.
-    See [Building ILLIXR][10] for more details on the config file.
+    This can be accomplished by editing the flows/plugin groups included for the `native`
+    The plugin flow structure for `native` shown here:
```
mhuzai, line 108: `native.yaml` is shown here:

runner/config_schema.yaml
```diff
@@ -42,80 +42,102 @@ definitions:
...
-                description: Vars that get passed to the build system
-            required:
-              - path
+        flow:
+          $ref: '#/definitions/flow'
+  append:
+    type: object
+    default: { plugin_group: [] }
+    additionalProperties: false
+    description: Configuration group to append to all included flows
```
mhuzai, line 95: Just to make sure I understand this correctly, this really does mean _all_? So if you had four flows and two appended plugins, the two plugins will get appended to each flow?

configs/data/euroc-default.yaml
```diff
@@ -0,0 +1 @@
+euroc-vicon-default.yaml
```
mhuzai, line 1: Per-dataset yamls make sense, but why have `euroc-vicon-default` and `euroc-default`? The nesting seems to be too much. Why not just include the dataset directly in `data-default`?

configs/ci/run-solo.yaml
```diff
@@ -0,0 +1,9 @@
+plugin_group:
```
mhuzai, line 1: I am not sure I understand the intent of this file. With the exception of `audio` and `offline_imu_cam`, all the plugins here are dependent on other plugins, and running them solo will not really test anything. And given that both `audio` and `offline_imu_cam` are going to be a part of a flow, why is this file required at all?

configs/ci/build-only.yaml
```diff
@@ -0,0 +1,5 @@
+plugin_group:
+  - !include "plugins/debugview.yaml"
+  - !include "plugins/gldemo.yaml"
+  - !include "plugins/timewarp_gl.yaml"
+  - !include "plugins/pose_lookup.yaml"
```
mhuzai, line 5: These three will get compiled and run as part of the CI flows. Shouldn't `build-only.yaml` only be for plugins that can't be run at all in CI?

configs/plugins/groups/shared.yaml
```diff
@@ -0,0 +1,3 @@
+plugin_group:
```
mhuzai, line 1: Why is this required, given the `common` and `runtime` fields in `action`?

runner/config_schema.yaml
e3m3: Yes, both plugins with be appended to each flow. Would "each" be better?

configs/ci/run-solo.yaml
e3m3: I would argue that our plugin set and our Switchboard implementation should be well behaved enough that a plugin running in isolation should not affect the stability of the system. While I don't know semantically what this should mean for pure Phonebook services yet, I think plugins that are producers and consumers over Switchboard should launch and drain without crashing at a minimum.

There are also edge cases about the system design that might be covered by running plugins in isolation. For example, have we ever seen a case in the wild where Switchboard's queues fill up because there are no available consumers? We can extend our `run-solo` Ii to generate new test case scenarios.

We can also [see `run-solo` at work](https://github.com/ILLIXR/ILLIXR/runs/2564890077?check_suite_focus=true#step:10:2633) now that we have both Ubuntu 18 and 20 in our CI:
- Ubuntu 18 crashes on the `offline_imu_cam` plugin, with a curious mention to a VTK missing reference on drain (`dlclose`).
- Ubuntu 20 makes it past the `run-solo` section, and crashes during the [`pose_lookup` flow](https://github.com/ILLIXR/ILLIXR/runs/2564889969?check_suite_focus=true#step:10:2840).

configs/ci/build-only.yaml
e3m3: The interface for specifying plugins (across this entire PR refactor) is designed to be verbose, but explicit. This makes it clearer how things connect. The file structure is designed so that adding and removing new plugins, or editing properties about specific data/plugins generates the fewest conflicts for existing configs/plugins

In terms of dependencies, again, the interface is explicit here. We indicate everything that should be built. The CI doesn't assume anything about the dependence structure, or properties of the application that might _change_ over time. Instead, it allows for a smarter implementation within Runner or a new build system to detect the dependencies and alter the compilation strategy accordingly.

Following this idea, the verbosity of the plumbing can be reduced later by augmenting Runner to do more automatic detection of plugins.

e3m3: > In terms of dependencies, again, the interface is explicit here. We indicate everything that should be built. The CI doesn't assume anything about the dependence structure, or properties of the application that might change over time. Instead, it allows for a smarter implementation within Runner or a new build system to detect the dependencies and alter the compilation strategy accordingly.

I add a note here that this design is also motivated by maintenance overhead. If we handle the strategies in the Runner/scripts, its much easier for an ILLIXR dev, new user, or plugin writer to understand what the plumbing changes are (similar to reducing conflicts above).

configs/plugins/groups/shared.yaml
e3m3: Yeah I agree that the current design makes it better to use `config["runtime"]` and `config["common"]`, and have been moving some of my Runner stuff to this, but it's not everywhere yet.

I have been thinking that we are still in-between design decisions for `common` and `runtime`. If we do move to the multi-repo design for both, it might make sense to continue treating them like normal Runner `Path`s / plugins (for build related actions). With that said, this file isn't strictly necessary.

configs/plugins/groups/rt_slam.yaml
```diff
@@ -0,0 +1,3 @@
+plugin_group:
+  - !include "plugins/offline_imu_cam.yaml"
+  - !include "plugins/pose_prediction.yaml"
```
e3m3, line 3: Refactoring for the plugin groups (like this one) seems necessary to make things simpler. The existing group yaml files with matching names to the old files (e.g. `plugins/groups/rt_slam.yaml` and `rt_slam_plugins.yaml`) are the result of factoring out the common plugins into new yaml files based on their usage across the action yaml files. However, the plugins left behind don't seem to match the purpose of their original plugin group's name.

configs/plugins/groups/misc-native.yaml
```diff
@@ -0,0 +1,3 @@
+plugin_group:
+  - !include "plugins/debugview.yaml"
+  - !include "plugins/offload_data.yaml"
```
e3m3, line 3: This may use a better name (see `plugins/groups/rt_slam.yaml` comment).

configs/plugins/groups/misc.yaml
```diff
@@ -0,0 +1,3 @@
+plugin_group:
+  - !include "plugins/ground_truth_slam.yaml"
+  - !include "plugins/gldemo.yaml"
```
e3m3, line 3: This may use a better name (see `plugins/groups/rt_slam.yaml` comment).

Dockerfile
```diff
@@ -1,32 +1,55 @@
...
-RUN apt-get update && apt-get install -y sudo
+RUN apt-get update && apt-get install -y sudo curl unzip
+
+## Create illixr_dir, cache_path, data_dir_euroc, ${sub_path_euroc} together
+## Prevents Runner from fetching
+RUN mkdir -p ${data_dir_euroc}/${sub_path_euroc}
+
+## Download (and cache) the euroc dataset early in the build
+RUN curl ${url_euroc} -o ${zip_path_euroc}
+RUN unzip -d ${data_dir_euroc} ${zip_path_euroc} ${sub_path_euroc}/ # slash required
```
e3m3, line 42: EuRoC is now cached. Closes #279.

e3m3: A more robust solution would install conda as early as possible, change into the work directory, and run `./runner.sh configs/fetch.yaml` (a new action for data and possibly also external plugins; `pathify`'d objects).

runner/config_schema.yaml
mhuzai: No, the writing is fine, I was just thinking that given that it's _all_, we could theoretically list all the common plugins here, and only list the flow-specific ones under `flows`. That would have been nice, but then I remembered that, among other things, plugin order matters.

e3m3: Right that make sense. I was thinking about also adding a similar `prepend` option, but that only alleviates the problem slightly.

configs/ci/run-solo.yaml
mhuzai: Gotcha. Just so we're on the same page: the idea here is to make sure that the plugin actually doesn't do anything, like it's supposed to (if there are no pubs)?

e3m3: > Gotcha. Just so we're on the same page: the idea here is to make sure that the plugin actually doesn't do anything, like it's supposed to (if there are no pubs)?

Right! For now I think that's sufficient. If the plugin interface changes of course, we can reconsider what the behavior should be.

e3m3: In terms of execution time, I think we can play with the `RUN_DURATION` here if we find CI takes too long in the `run-solo` section. The `run-solo` section is also parallelizable if needed, since ILLIXR with just one plugin will not need many resources.

Parallelizing flows is trickier, since each flow would be a full ILLIXR app running on the CI machine. Ideally, we would launch each flow on a dedicated CI machine, but this requires exposing more functionality via Docker+Runner to GitHub Actions I think.

e3m3: > Gotcha. Just so we're on the same page: the idea here is to make sure that the plugin actually doesn't do anything, like it's supposed to (if there are no pubs)?

Plugins that depend on HW can perform their detection too, if we are able to passthrough the right resources via virtualization.

e3m3:
Depends on #303 / #304 .

astro-friedel:
Thisis superceded by the upcoming testing framework

PR #287, opened 2021-05-06
Changed errno assertions to RAC calls

### LINKED ISSUE #284: Remove errno asserts
Monado does not have the type of errno checking that we rely on in ILLIXR, and therefore the errno asserts fail all the time, unless we build Monado in `Release`. This is not ideal in general, and definitely not ideal for profiling (no symbols). I'm thinking that we should only keep the RACs and remove all the asserts. The RACs are sufficient as a debugging aid.

e3m3:
Closes #284.

### DIFF
--- a/common/cpu_timer.hpp
+++ b/common/cpu_timer.hpp
@@ -29,7 +29,7 @@ cpp_clock_gettime(clockid_t clock_id) {
 				  : "memory" /* Clobbers */);
     struct timespec ts;
 
-    assert(errno == 0 && "Errno should not be set before clock_gettime");
+    RAC_ERRNO_MSG("cpu_timer before clock_gettime");
 
     if (clock_gettime(clock_id, &ts)) {
         throw std::runtime_error{std::string{"clock_gettime returned "} + strerror(errno)};
--- a/common/dynamic_lib.hpp
+++ b/common/dynamic_lib.hpp
@@ -47,7 +47,7 @@ public:
 		char* error;
 
 		// dlopen man page says that it can set errno sp
-		assert(errno == 0 && "Errno should not be set before dlopen");
+		RAC_ERRNO_MSG("dynamic_lib before dlopen");
 		void* handle = dlopen(path.data(), RTLD_LAZY | RTLD_LOCAL);
 		RAC_ERRNO_MSG("dynamic_lib after dlopen");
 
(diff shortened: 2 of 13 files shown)


mhuzai (approved):
I had forgotten how many RACs we have in the code. Do we really need that many?

e3m3:
> I had forgotten how many RACs we have in the code. Do we really need that many?

Short answer: No. The assertions replaced by RACs in this PR occasionally did catch new errno's from time to time. We can remove more of the RACs, understanding that the log produced by the remaining triggered RACs will describe errors at a coarser granularity.

charmoniumQ (approved):

mhuzai:
> 
> 
> > I had forgotten how many RACs we have in the code. Do we really need that many?
> 
> Short answer: No. The assertions replaced by RACs in this PR occasionally did catch new errno's from time to time. We can remove more of the RACs, understanding that the log produced by the remaining triggered RACs will describe errors at a coarser granularity.

I think we discussed this before, but I'm forgetting the exact discussion: how many errors does errno actually catch? Outside of that one audio bug where we weren't checking whether the file was opened properly, has errno helped us at all? Why don't other projects have similar errno checking mechanisms?

charmoniumQ:
If a function calls another function that uses `errno` to indicate errors, it must clear before and test afterwards in order to know if an error happened. All projects do this for the sake of correctness.

But in ILLIXR we even stricter. We basically push the clear up to the last operation that could have set errno. This way, if `errno` is set in the middle of inoccuous code, it always indicates someone forgot to clear it, which might make it easier to identify code that doesn't check for errors, of which there was a lot in ILLIXR. When I set out on this, I didn't know how many functions would be able to set `errno` spuriously (e.g. `errno` of `someFunction()` is only meaningful if it returns non-zero). I think this is why most projects don't do this.

I think we have covered much of the code that doesn't check for errors now, so I am less concerned by that.

My recommendation: If little/no maintenance is required, keep the errno checks and clears. give us the errno "always means something is wrong" property. If it requires effort to maintain or if it spews spurious output, remove them. Unfortunately, this must be a globally consistent choice. While it is possible to have different disciplines in different domains, I do not recommend this.

mhuzai (approved):

charmoniumQ (approved):

PR #288, opened 2021-05-06
Implemented formatter and formatted `ILLIXR/`

### LINKED ISSUE #45: Run clang-format

e3m3:
Closes #45.
- Formatter and delinter code implemented as `delint` action (`configs/delint.yaml`).
- Delinter disabled (clang compilation errors for included libraries).
- Code at the root of `ILLIXR/` now formatted.


charmoniumQ (approved):

common/threadloop.hpp
```diff
@@ -27,151 +31,153 @@ const record_header __threadloop_iteration_header {"threadloop_iteration", {
...
+    {
+        assert(_m_stoplight->check_should_stop());
+        assert(!_m_thread.joinable());
+    }
 
 protected:
-	std::size_t iteration_no = 0;
-	std::size_t skip_no = 0;
+    std::size_t iteration_no = 0;
+    std::size_t skip_no      = 0;
```
e3m3, line 83: `common/threadloop.hpp` Sample 1: `threadloop`

mhuzai:
The code looks _so_ good now! I like the format overall. The only thing I don't like is the spacing of the braced lists -- there is too much of it. So if we could do `Cpp11BracedListStyle: false` --> `Cpp11BracedListStyle: true` that would be great.

Let the holy wars begin.

mhuzai (commented):
Looks good overall. A few things:
- These components haven't been formatted: `realsense`, `offload_data`.
- Why is it necessary to specify the plugins in `delint.yaml`? Shouldn't we just go through the entire repo and format everything? Or is this required because of the linter?
- What about all the non-C++ files -- configs, install scripts, runner, etc.?
- What's going to be the overhead of duplicating this in the external component repos (OV, audio, etc.)?
- We should add instructions on how to keep the code formatted to `CONTRIBUTING.md`.
- Tangential: there are a lot of unused properties in `delint.yaml`, and looks like it's because most of them don't have default values. Should we add defaults to the schema to clean things up a bit? **This is minor.**

e3m3:
> Looks good overall. A few things:

>   These components haven't been formatted: realsense, offload_data.

:+1: 

>    Why is it necessary to specify the plugins in delint.yaml? Shouldn't we just go through the entire repo and format everything? Or is this required because of the linter?

All actions in the current design needs a list of plugins. In #270 this will be very clean to specify, albeit verbose. However, the verbose interface is explicit, and will be easy to replace with automated scripts later.

Linter does require "domain" knowledge about what is a plugin, since all headers are needed for compilation. Its easier to handle a plugin as a unit, and explode its contents as a list of sources and headers for processing.

>    What about all the non-C++ files -- configs, install scripts, runner, etc.?

These are not handled. We would need a `yaml`, `python`, and `bash` formatter. These may be integrated later, but it won't be as seamless, since the current delint action is plugin-based (via `pyyaml-include` and the action config schema).

>    What's going to be the overhead of duplicating this in the external component repos (OV, audio, etc.)?

Nil. Once #270 merges, it will be seamless to run the delinter on external plugins. The only trick needed is that one would need to `cd` into the `.cache/path` to commit the delinted code from there. This may require a little manual work for managing remotes (since these are cloned by Runner).

Another alternative that is just as easy: when delinting external plugins, point the yaml to a repo not in `.cache/path` (e.g., `path: ../audio_pipeline`), delint, and then commit from that externally set up directory.

>    We should add instructions on how to keep the code formatted to CONTRIBUTING.md.

:+1: 

>    Tangential: there are a lot of unused properties in delint.yaml, and looks like it's because most of them don't have default values. Should we add defaults to the schema to clean things up a bit? This is minor.

This is/will be taken care of between #270 and #198.

e3m3:
> What's going to be the overhead of duplicating this in the external component repos (OV, audio, etc.)?

>> Nil. Once #270 merges, it will be seamless to run the delinter on external plugins. The only trick needed is that one would need to cd into the .cache/path to commit the delinted code from there. This may require a little manual work for managing remotes (since these are cloned by Runner).

>> Another alternative that is just as easy: when delinting external plugins, point the yaml to a repo not in .cache/path (e.g., path: ../audio_pipeline), delint, and then commit from that externally set up directory.

Follow up point: A downside to delinting external plugins is that fast-forwarding them is tricky. We will want to keep rebasing to replay the formatting on top of the upstream changes, otherwise, tons of conflicts will make updates impossible.

mhuzai:
Closed via #343.

PR #296, opened 2021-05-13
Applied memory leak fixes

### LINKED ISSUE #295: Memory leaks in `common/extended_window.hpp`
From this [discussion](https://github.com/ILLIXR/ILLIXR/pull/112#issuecomment-837308821).

Changes found [here](https://github.com/ILLIXR/ILLIXR/pull/112/files#diff-c1f222918af8ba25d523451d3d976b833e649b8567b75a2d79c82bf159058802).

e3m3:
Closes #295.

### DIFF
--- a/common/extended_window.hpp
+++ b/common/extended_window.hpp
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <cerrno>
+#include <cassert>
 #include <X11/X.h>
 #include <X11/Xlib.h>
 #include <GL/glx.h>
@@ -19,10 +20,10 @@ namespace ILLIXR {
     public:
         int        width;
         int        height;
-        Display    *dpy;
+        Display*   dpy;
         Window     win;
         GLXContext glc;
-        xlib_gl_extended_window(int _width, int _height, GLXContext _shared_gl_context){
+        xlib_gl_extended_window(int _width, int _height, GLXContext _shared_gl_context) {
             width = _width;
             height = _height;
 
@@ -32,7 +33,7 @@ namespace ILLIXR {
             RAC_ERRNO_MSG("extended_window at start of xlib_gl_extended_window constructor");
 
             dpy = XOpenDisplay(nullptr);
-            if (!dpy) {
+            if (dpy == nullptr) {
                 ILLIXR::abort("Cannot connect to X server");
             } else {
 				// Apparently, XOpenDisplay's _true_ error indication is whether dpy is nullptr.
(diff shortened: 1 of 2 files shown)


charmoniumQ (approved):

charmoniumQ (approved):

common/extended_window.hpp
```diff
@@ -176,5 +182,13 @@ namespace ILLIXR {
             */
 #endif
         }
+
+        ~xlib_gl_extended_window() {
+            XDestroyWindow(dpy, win);
+            Window root = DefaultRootWindow(dpy);
+            XDestroyWindow(dpy, root);
+            //glXDestroyContext(dpy, glc);
```
mhuzai, line 190: Comment this in, and also bring the [_entire_ destructor](https://github.com/ILLIXR/ILLIXR/blob/master/timewarp_gl/plugin.cpp#L783) of timewarp here -- timewarp shouldn't be cleaning up resources allocated by this service.
The only thing I'm unsure of is whether the color map needs to be freed explicitly or not, given that the [XCloseDisplay](https://tronche.com/gui/x/xlib/display/XCloseDisplay.html) documentation says that it destroys all the resources attached to the display.

mhuzai (changes requested):

common/extended_window.hpp
e3m3: Ah that makes sense. I actually commented out `glxDestroyContext` because I wasn't able to get it to run without segfaulting. Maybe moving the `timewarp_gl` destructor here will help?

mhuzai: Yup, timewarp was the culprit. It already freed the context, resulting in the segfault here. Also remember to make the context current here before doing anything with it (it's already taken care of in the timewarp desctructor).

common/extended_window.hpp
```diff
@@ -176,5 +181,27 @@ namespace ILLIXR {
...
+            [[maybe_unused]] const bool gl_result = static_cast<bool>(glXMakeCurrent(dpy, None, nullptr));
+            assert(gl_result && "glXMakeCurrent should not fail");
+
+            /// Destroyed by XCloseDisplay
+            /// See [documentation](https://tronche.com/gui/x/xlib/display/XCloseDisplay.html)
+            // XDestroyWindow(dpy, win);
+            // Window root = DefaultRootWindow(dpy);
+            // XDestroyWindow(dpy, root);
+            // glXDestroyContext(dpy, glc);
+            // XFreeColormap(dpy, _m_cmap);
```
e3m3, line 197: Leave with comment, or remove?

charmoniumQ: Leave a comment listing the destroyed resources. No need to have the actual destructor functions.

mhuzai: I am not sure that `XCloseDisplay` actually cleans up [_everything_](https://community.khronos.org/t/properly-destroying-a-window/47476). Even [Khronos' own example](https://www.khronos.org/opengl/wiki/Programming_OpenGL_in_Linux:_GLX_and_Xlib) explicitly cleans up the attached resources before closing the display. So I think we should too. I'm finding conflicting information regarding the order of the frees, but I would imagine freeing in LIFO order should be the way to go: context, window, color map, display.

mhuzai (approved):

PR #297, opened 2021-05-15
Add support for OAK-D Camera

### LINKED ISSUE #280: Support for OAK-D type hardware
Recently, Luxonis has developed and shipped the OAK-D, a stereo camera + RGB + IMU as well as on-device vision processing (Myriad X) https://docs.luxonis.com/en/latest/pages/products/bw1098obc/. This same hardware also is available in customizable configurations, making it an interesting platform for future research. 

It would be beneficial to add initial support for this hardware family, following the same interface as the current Realsense and Zed cameras.

ltstein:
Creating a plugin to allow using the [OAK-D](https://docs.luxonis.com/en/latest/pages/products/bw1098obc/) within ILLIXR. It is intended as a base integration as it does not fully leverage the onboard compute available but provides similar output as the existing Realsense and Zed camera plugins. 

Closes #280 

### Working To Do:
 - [x] Integrate DepthAI SDK dependencies into build / dependency installation setup


ltstein:
First pass on getting DepthAI devices integrated. This is still dependent on a branch in DepthAI that is in review.

scripts/install_depthai.sh
```diff
@@ -0,0 +1,56 @@
+#!/bin/bash
+
+#--- Script for installing Kimera (RPGO) ---#
```
e3m3, line 3: Old comment

scripts/install_depthai.sh
```diff
@@ -0,0 +1,56 @@
...
+    dep_name="${dep_name_depthai}"
+fi
+
+if [ -z "${src_dir}" ]; then
+    src_dir="${parent_dir_depthai}/${dep_name_depthai}"
+fi
+
+if [ -z "${dep_ver}" ]; then
+    dep_ver="${dep_ver_depthai}"
+fi
```
e3m3, line 30: We are moving to this form now (for readability, starting in #270):
`dep_name="${dep_name:=dep_name_depthai}"`

depthai/plugin.cpp
```diff
@@ -0,0 +1,344 @@
+
+#include <cstdio>
+#include <iostream>
+
+#include "opencv2/opencv.hpp"
+
+// Inludes common necessary includes for development using depthai library
+#include "depthai/depthai.hpp"
```
e3m3, line 8: These should be searched from the environment I think:
```cpp
#include <opencv2/opencv.hpp>
#include <depthai/depthai.hpp>
```

depthai/plugin.cpp
```diff
@@ -0,0 +1,344 @@
...
+class depthai : public plugin {
+public:
+    depthai(std::string name_, phonebook* pb_)
+        : plugin{name_, pb_}
+        , sb{pb->lookup_impl<switchboard>()}
+        , _m_imu_cam{sb->get_writer<imu_cam_type>("imu_cam")}
+        , _m_rgb_depth{sb->get_writer<rgb_depth_type>("rgb_depth")}
+        //Initialize DepthAI pipeline and device 
+        , p{createCameraPipeline()}
+        , d{p}
```
e3m3, line 26: Maybe we can use a variables that a a little more descriptive but still short (e.g., `cp`, or `dev`)

depthai/plugin.cpp
```diff
@@ -0,0 +1,344 @@
...
+            auto rectifR = rectifRightQueue->tryGet<dai::ImgFrame>();
+            
+            cv::Mat color = cv::Mat(colorFrame->getHeight(), colorFrame->getWidth(), CV_8UC3, colorFrame->getData().data());
+            cv::Mat rgb_out{color.clone()};
+            cv::Mat rectifiedLeftFrame = cv::Mat(rectifL->getHeight(), rectifL->getWidth(), CV_8UC1, rectifL->getData().data());
+            cv::Mat LeftOut{rectifiedLeftFrame.clone()};
+            cv::flip(LeftOut, LeftOut, 1);
+            cv::Mat rectifiedRightFrame = cv::Mat(rectifR->getHeight(), rectifR->getWidth(), CV_8UC1, rectifR->getData().data());
+            cv::Mat RightOut{rectifiedRightFrame.clone()};
+            cv::flip(RightOut, RightOut, 1);
```
e3m3, line 81: How come these are cloned before use?

depthai/plugin.cpp
```diff
@@ -0,0 +1,344 @@
...
+        
+        std::chrono::time_point<std::chrono::steady_clock, std::chrono::steady_clock::duration> gyroTs;
+        Eigen::Vector3f la;
+        Eigen::Vector3f av;
+        
+        if(imuPacket_go) {
+            auto imuPacket = imuQueue->tryGet<dai::IMUData>();
+            #ifndef NDEBUG
+                if(imu_packet == 0){
+                    first_packet = std::chrono::steady_clock::now();
```
e3m3, line 119:
```suggestion
                    first_packet_time = std::chrono::steady_clock::now();
```
A little clearer

depthai/plugin.cpp
```diff
@@ -0,0 +1,344 @@
...
+        int rgbd_pub{0};
+        int rgb_count{0};
+        int left_count{0};
+        int right_count{0};
+        int depth_count{0};
+        int all_count{0};
+    #endif
+    std::chrono::time_point<std::chrono::steady_clock> first_packet;
+    std::chrono::time_point<std::chrono::steady_clock, std::chrono::steady_clock::duration> test_time_point;
+    bool useRaw = true;
```
e3m3, line 224: Can configure later (#198)

depthai/plugin.cpp
```diff
@@ -0,0 +1,344 @@
...
+                }
+                imu_packet++;
+            #endif
+            
+            auto imuDatas = imuPacket->imuDatas;
+            for(auto& imuData : imuDatas) {
+                if(useRaw){
+                    gyroTs = std::chrono::time_point_cast<std::chrono::nanoseconds>(imuData.rawGyroscope.timestamp.getTimestamp());
+                    if (gyroTs <= test_time_point){return;}
+                    test_time_point = gyroTs;
```
e3m3, line 129: What is the initial value of `test_time_point`?

depthai/plugin.cpp
```diff
@@ -0,0 +1,344 @@
...
+                #endif
+                _m_rgb_depth.put(_m_rgb_depth.allocate<rgb_depth_type>(
+                    {
+                        rgb,
+                        depth,
+                        imu_time
+                    }
+                ));
+            }
+        }
```
e3m3, line 178: In the current implementation, we always have `rgb` and `depth` available (i.e. not `nullopt`) since we always copy them together with `img0` and `img1` (we wait for all for `_go`'s). If this is the desired behavior, then we can short-circuit this `callback` earlier if we haven't received all the data yet.

depthai/plugin.cpp
```diff
@@ -0,0 +1,344 @@
...
+    bool useRaw = true;
+
+    dai::Pipeline p;
+    dai::Device d;
+
+    std::shared_ptr<dai::DataOutputQueue> color;
+    std::shared_ptr<dai::DataOutputQueue> depthQueue;
+    std::shared_ptr<dai::DataOutputQueue> rectifLeftQueue;
+    std::shared_ptr<dai::DataOutputQueue> rectifRightQueue;
+    std::shared_ptr<dai::DataOutputQueue> imuQueue;
```
e3m3, line 233: Change `color` -> `colorQueue` for consistency.

depthai/plugin.cpp
```diff
@@ -0,0 +1,344 @@
...
+
+class depthai : public plugin {
+public:
+    depthai(std::string name_, phonebook* pb_)
+        : plugin{name_, pb_}
+        , sb{pb->lookup_impl<switchboard>()}
+        , _m_imu_cam{sb->get_writer<imu_cam_type>("imu_cam")}
+        , _m_rgb_depth{sb->get_writer<rgb_depth_type>("rgb_depth")}
+        //Initialize DepthAI pipeline and device 
+        , p{createCameraPipeline()}
```
charmoniumQ, line 25: If we don't need `p`, let's not store it; just construct `d` directly.

depthai/plugin.cpp
```diff
@@ -0,0 +1,344 @@
...
+            cv::flip(LeftOut, LeftOut, 1);
+            cv::Mat rectifiedRightFrame = cv::Mat(rectifR->getHeight(), rectifR->getWidth(), CV_8UC1, rectifR->getData().data());
+            cv::Mat RightOut{rectifiedRightFrame.clone()};
+            cv::flip(RightOut, RightOut, 1);
+
+            cv::Mat depth = cv::Mat(depthFrame->getHeight(), depthFrame->getWidth(), CV_16UC1, depthFrame->getData().data());
+            cv::Mat converted_depth;
+            depth.convertTo(converted_depth, CV_32FC1, 1000.f);
+            cam_type_ = cam_type {
+                .img0 = cv::Mat{LeftOut},
```
charmoniumQ, line 87: We should be able to set `.img0 = LeftOut`.

depthai/plugin.cpp
charmoniumQ: We'll wait for @ltstein to respond, but I think it's because the library wants to reuse the same `cv::Mat` allocation.

depthai/plugin.cpp
```diff
@@ -0,0 +1,344 @@
...
+        if(rectifR_go && rectifL_go && depth_go && color_go) {
+            #ifndef NDEBUG
+                all_count++;
+            #endif
+            auto colorFrame = color->tryGet<dai::ImgFrame>();
+            auto depthFrame = depthQueue->tryGet<dai::ImgFrame>();
+            auto rectifL = rectifLeftQueue->tryGet<dai::ImgFrame>();
+            auto rectifR = rectifRightQueue->tryGet<dai::ImgFrame>();
+            
+            cv::Mat color = cv::Mat(colorFrame->getHeight(), colorFrame->getWidth(), CV_8UC3, colorFrame->getData().data());
```
charmoniumQ, line 74: Is this not equivalent to `.clone()`?

depthai/plugin.cpp
charmoniumQ: If #250 gets merged first, change this to be based on the relative clock.

depthai/plugin.cpp
```diff
@@ -0,0 +1,344 @@
...
+            cv::Mat rectifiedLeftFrame = cv::Mat(rectifL->getHeight(), rectifL->getWidth(), CV_8UC1, rectifL->getData().data());
+            cv::Mat LeftOut{rectifiedLeftFrame.clone()};
+            cv::flip(LeftOut, LeftOut, 1);
+            cv::Mat rectifiedRightFrame = cv::Mat(rectifR->getHeight(), rectifR->getWidth(), CV_8UC1, rectifR->getData().data());
+            cv::Mat RightOut{rectifiedRightFrame.clone()};
+            cv::flip(RightOut, RightOut, 1);
+
+            cv::Mat depth = cv::Mat(depthFrame->getHeight(), depthFrame->getWidth(), CV_16UC1, depthFrame->getData().data());
+            cv::Mat converted_depth;
+            depth.convertTo(converted_depth, CV_32FC1, 1000.f);
```
charmoniumQ, line 85: I think `.convertTo` makes a clone, so need to `.clone` depth. This is important because cloning is probably expensive.

depthai/plugin.cpp
```diff
@@ -0,0 +1,344 @@
...
+            cam_type_ = cam_type {
+                .img0 = cv::Mat{LeftOut},
+                .img1 = cv::Mat{RightOut},
+                .rgb = cv::Mat{rgb_out},
+                .depth = cv::Mat{converted_depth},
+                .iteration = iteration_cam,
+            };
+            iteration_cam++;
+        }
+        // Images
```
charmoniumQ, line 95: If you move these lines to the start of the method, then you can set them directly (instead of `cam_type_`.

depthai/plugin.cpp
```diff
@@ -0,0 +1,344 @@
...
+            depthQueue = d.getOutputQueue("depth", 1, false);
+            rectifLeftQueue = d.getOutputQueue("rectified_left", 1, false);
+            rectifRightQueue = d.getOutputQueue("rectified_right", 1, false);
+            imuQueue = d.getOutputQueue("imu", 1, false);
+            std::function<void(void)> imu_callback = [&](){callback();};
+            imuQueue->addCallback(imu_callback);
+        }
+
+    void callback(){  
+        std::lock_guard<std::mutex> lock(mutex);
```
charmoniumQ, line 42: Is the `callback` going to be called from multiple threads? If not, we should remove this lock and mutex.

depthai/plugin.cpp
ltstein: That makes sense I think, since they are installed vs found within the project

depthai/plugin.cpp
ltstein: This is a good point, will do.

depthai/plugin.cpp
ltstein: I am looking into this more (along with the constructor vs cloning above) but the working hypothesis is that the DepthAI library is reusing the same memory for the queue, causing an issue when ILLIXR is trying to utilize it in the pipeline.

depthai/plugin.cpp
ltstein: Does not appear to be. 


data | Pointer  to the user data. Matrix constructors that take data and step  parameters do not allocate matrix data. Instead, they just initialize  the matrix header that points to the specified data, which means that _no  data is copied._
-- | --



[Doc Ref](https://docs.opencv.org/master/d3/d63/classcv_1_1Mat.html#a51615ebf17a64c968df0bf49b4de6a3a)

depthai/plugin.cpp
ltstein: Yes, it appears this use of `.convertTo` would trigger reallocation so no need for _another_ clone. From the docs


m | output matrix; if it does not have a proper size or type before the operation, it is reallocated.
-- | --




[Doc Ref](https://docs.opencv.org/master/d3/d63/classcv_1_1Mat.html#adf88c60c5b4980e05bb556080916978b)

depthai/plugin.cpp
ltstein: I'm thinking this will need some conversion then since the DepthAI library uses steady_clock, but I'm not sure

depthai/plugin.cpp
ltstein: It shouldn't be, but I need to verify. Also, I wanted to investigate this comment regarding that
https://github.com/ILLIXR/ILLIXR/blob/287f6c698b2a479545d60f409cb2fcfa70a96f2f/realsense/plugin.cpp#L37

> // This lock guarantees that concurrent invocations of `callback` are serialized.
            // Without this lock, prior invocations of `callback` are not necessarily "happens-before" ordered, so accessing persistent variables constitutes a data-race, which is undefined behavior in the C++ memory model.

ltstein:
Going to be refactoring based on the IMU dependencies now released in the main branch along with a few other changes to update.

charmoniumQ (commented):
I've been away from this project (only working in my free time), so sorry for the delayed response.

The code all looks good pending pending @e3m3's comments.

scripts/install_depthai.sh
ltstein: Is the format backwards compatible or do I need to wait for that to be implemented?

depthai/plugin.cpp
ltstein: I'm confused on what we are short circuiting and the gain. We always publish the ```imu_cam_type``` based on IMU data regardless of the images and if we don't have RGB/depth data we jump to the end of the callback.

scripts/install_depthai.sh
e3m3: Yup it's backwards compatible! We can change it here or wait until #270 with no issue

e3m3 (approved):

PR #304, opened 2021-06-03
Fixed `pose_lookup` and `dynamic_lib`

### LINKED ISSUE #303: `pose_lookup` initialization is buggy
Fix the constructor initialization list in `ILLIXR/pose_lookup/plugin.cpp` to use the appropriate default values (e.g. `Eigen::Matrix3f::Zero()`).

e3m3:
Closes #303 .
- Fixed `pose_lookup` initialization in `pose_lookup/plugin.cpp`
- Added debugging prints to `common/dynamic_lib.hpp`
- New `path_basename` for debugging in `common/dynamic_lib.hpp`
- Failed calls to `dlclose` non-fatal with `dbg` in `common/dynamic_lib.hpp`

### DIFF
--- a/common/dynamic_lib.hpp
+++ b/common/dynamic_lib.hpp
@@ -25,28 +25,28 @@ Usage:
 
 class dynamic_lib {
 private:
-	dynamic_lib(
-	    void_ptr&& handle
+    dynamic_lib(
+        void_ptr&& handle
         , const std::string& lib_name = ""
-	) : _m_handle{std::move(handle)}
+    ) : _m_handle{std::move(handle)}
       , _m_lib_name{std::move(lib_name)}
-	{ }
+    { }
 
 public:
-	dynamic_lib(dynamic_lib&& other)
-		: _m_handle{std::move(other._m_handle)}
+    dynamic_lib(dynamic_lib&& other)
+        : _m_handle{std::move(other._m_handle)}
         , _m_lib_name{std::move(other._m_lib_name)}
-	{ }
+    { }
 
-	dynamic_lib& operator=(dynamic_lib&& other) {
-		if (this != &other) {
-			_m_handle   = std::move(other._m_handle);
-			_m_lib_name = std::move(other._m_lib_name);
-		}
-		return *this;
-	}
+    dynamic_lib& operator=(dynamic_lib&& other) {
+        if (this != &other) {
+            _m_handle   = std::move(other._m_handle);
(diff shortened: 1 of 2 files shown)


charmoniumQ (commented):
Imo, the `-DNDEBUG` and normal builds have deviated too far in `dynamic_lib`. One has a destructor, the other doesn't.

Can we merge the paths and always propagate debug information?

e3m3:
Partially addresses #305 for `dbg`. Does not resolve root issue with missing symbol reference to VTK in `debugview`.

charmoniumQ (commented):
I intend `dynamic_lib` to be used outside of ILLIXR, and in that case, the basename of the dirname is not so meaningful. Let's set the name to the while pathname.

Then lgtm.

common/dynamic_lib.hpp
```diff
@@ -23,70 +25,97 @@ using void_ptr = std::unique_ptr<void, std::function<void(void*)>>;
...
+        return dynamic_lib{
+            void_ptr{handle, [](void* handle) {
+                RAC_ERRNO();
+
+                char* error;
+                int ret = dlclose(handle);
+                if ((error = dlerror()) || ret) {
+                    const std::string msg_error {"dlclose(): " + (error == nullptr ? "NULL" : std::string{error})};
+#ifndef NDEBUG
+                    /// If debugging, only report the dlclose error (non-fatal, can leak memory)
```
charmoniumQ, line 84: Why not always throw the error? The message should be pretty-printed if the exception bubbles up to the user (i.e. if it isn't caught along the way).

mhuzai (approved):

common/dynamic_lib.hpp
```diff
@@ -23,70 +25,92 @@ using void_ptr = std::unique_ptr<void, std::function<void(void*)>>;
...
+                "dlsym(\"" + symbol_name + "\"): " + (error == nullptr ? "NULL" : std::string{error})
+            };
+        }
+        return symbol;
+    }
+
+    template <typename T>
+    const T get(const std::string& symbol_name) const {
+        const void* obj = (*this)[symbol_name];
+        // return reinterpret_cast<const T>((*this)[symbol_name]);
```
charmoniumQ, line 107: I believe we should use a C++-style cast here. It's more explicit because it says to reinterpret, not to dynamic or static cast.

charmoniumQ (approved):

PR #311, opened 2021-10-11
Fixed ZED CMake

### LINKED ISSUE #95: Fix ZED Cmake
ZED's CMakeLists.txt currently has an additional linkage step which requires us to manually add C++ flags and a fake main(). Need to clean this up.

hungdche:
Closes #95 

- Deleted the fake `main` in `zed/src/main.cpp`
- Added `SET(CMAKE_CXX_STANDARD 17)` and '-lpthread' at the end of `target_link_libraries` in `zed/CMakeLists.txt` to remove the C++ flag
- Removed all Windows preferences

### DIFF
--- a/deps.sh
+++ b/deps.sh
@@ -25,12 +25,11 @@ export script_path_apt="scripts/install_apt_deps.sh"
 #export dep_ver_clang="llvmorg-10.0.1"
 
 ## Boost
-## Locally built boost not in use yet
-#export dep_name_boost="boost"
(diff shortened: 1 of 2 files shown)


charmoniumQ:
@mhuzai it has merges because I hit the "update to master" button in GitHub xD

In feature-branch history, I don't care that much because it gets squashed anyway on its way into master, but I'll resist the temptation to press that particular button next time.

PR #318, opened 2021-12-17
Issue 317 added --force for git checkout 

qinjunj:
Added --force to make sure local changes in .caches are overwritten.

### DIFF
--- a/configs/rt_slam_plugins.yaml
+++ b/configs/rt_slam_plugins.yaml
@@ -3,14 +3,14 @@ plugin_group:
   # - path: zed/
   # - path: realsense/
   # - path: depthai/
-  # - name: Kimera-VIO/
-  #   path:
-  #     git_repo: https://github.com/ILLIXR/Kimera-VIO.git
-  #     version: "3.1"
-  - name: OpenVINS
+  - name: Kimera-VIO/
     path:
-      git_repo: https://github.com/ILLIXR/open_vins.git
-      version: "3.0"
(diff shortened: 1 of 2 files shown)


configs/rt_slam_plugins.yaml
```diff
@@ -3,14 +3,14 @@ plugin_group:
   # - path: zed/
   # - path: realsense/
   # - path: depthai/
-  - name: Kimera-VIO/
```
charmoniumQ, line 6: Let's keep Kimera instead of OpenVINS in this PR. We will need to make both work, but still leave Kimera as the default.

charmoniumQ: Other than that, looks good!

qinjunj: Added Kimera back and commented OpenVINS/

charmoniumQ (approved):

charmoniumQ:
@jianxiapyh can you review?

The point of adding `--force` is to make sure that the version of the plugin located inside `.cache` is the same as the one defined in the config file (it should be 'declarative' style). `git checkout` is not strong enough by itself to enforce that consistency.

If one _does_ need to edit the plugin, the recommended workflow is to clone the plugin locally and change the `config/*.yaml` to refer to a local path rather than a git repository.

jianxiapyh (approved):

jianxiapyh (approved):

PR #321, opened 2022-01-24
Enable boost, without extraneous line

charmoniumQ:


mhuzai:
@shg8 has determined that installing boost from the apt repo will work just fine, so we'll revert the custom install and keep the apt install in another PR.

PR #323, opened 2022-02-16
Revert boost install script

mhuzai:
This reverts commit 8590c7422c9e0e72d11df0717cee201a12cb00d0.
We do not need to build boost ourselves, the one from apt works just fine.

### DIFF
--- a/deps.sh
+++ b/deps.sh
@@ -25,11 +25,12 @@ export script_path_apt="scripts/install_apt_deps.sh"
 #export dep_ver_clang="llvmorg-10.0.1"
 
 ## Boost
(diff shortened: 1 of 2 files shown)


JeffreyZh4ng (approved):

shg8 (changes requested):
I think we also need to comment out the prompt_install commands in install_deps.sh

mhuzai:
> I think we also need to comment out the prompt_install commands in install_deps.sh

Good catch! How on earth was this even working before the commit that I reverted got pushed? :confused:

charmoniumQ (approved):

PR #324, opened 2022-02-25
Separate IMU and Cam

qinjunj:


common/plugin.hpp
```diff
@@ -49,6 +49,7 @@ namespace ILLIXR {
 		 * freed memory. Instead, we call plugin->stop manually before destrying anything.
 		 */
 		virtual void stop() { }
+		virtual void internal_stop() {}
```
charmoniumQ, line 52: This should be in `threadloop` but not `plugin`. `plugin` doesn't really have a notion of "internal stop" because it isn't a thread. Classes that inherit `plugin` (like `threadloop`) can add this individually if they want to.

charmoniumQ: Let's also put a "documentation comment" before it saying this is what a `threadloop` should call if it wants to stop itself.

common/threadloop.hpp
```diff
@@ -81,9 +85,10 @@ class threadloop : public plugin {
 		record_coalescer it_log {record_logger_};
 		std::cout << "thread," << std::this_thread::get_id() << ",threadloop," << name << std::endl;
 
+		_m_stoplight->wait_for_ready();
 		_p_thread_setup();
 
-		_m_stoplight->wait_for_ready();
+		// _m_stoplight->wait_for_ready();
 		while (!_m_stoplight->check_should_stop()) {
```
charmoniumQ, line 92:
```suggestion
		while (!_m_stoplight->check_should_stop() || should_terminate()) {
```

It seems the class already has a variable, `_m_should_terminate`, that we can use instead of `_m_internal_stop`. Fewer variables/methods is preferable. I don't _think_ it's used anywhere else.

common/threadloop.hpp
```diff
@@ -169,8 +174,13 @@ class threadloop : public plugin {
 		return _m_terminate.load();
 	}
 
+	bool should_internal_stop() {
+		return _m_internal_stop.load();
+	}
+
 private:
 	std::atomic<bool> _m_terminate {false};
+	std::atomic<bool> _m_internal_stop {false};
```
charmoniumQ, line 183:
```suggestion
```

charmoniumQ (approved):

common/threadloop.hpp
qinjunj: `_m_should_terminate` is used in [runtime/main.cpp] (https://github.com/ILLIXR/ILLIXR/blob/823e3a6161de588468244e4f9a24c86df0ebad30/runtime/main.cpp#L58)

charmoniumQ: That is a different `_m_should_terminate` defined [here](https://github.com/ILLIXR/ILLIXR/blob/823e3a6161de588468244e4f9a24c86df0ebad30/runtime/main.cpp#L64).

qinjunj: Yeah you are right

qinjunj: Can I stick with `_m_internal_stop`? The name makes more sense to express what it is used for.

PR #326, opened 2022-03-08
Adding rk4_integrator to monado-mainline.yaml

### LINKED ISSUE #325: Adding rk4_integrator to monado-mainline.yaml
Head poses are incorrect when running Unreal application with ILLIXR because IMU integrator is not listed as a plugin in `monado-mainline.yaml`. Adding `rk4_integrator` helps resolve the issue.

hungdche:
Closes #325 

Head poses are incorrect when running Unreal application with ILLIXR because IMU integrator is not listed as a plugin in `monado-mainline.yaml`. Adding `rk4_integrator` helps resolve the issue.

### DIFF
--- a/configs/monado-mainline.yaml
+++ b/configs/monado-mainline.yaml
@@ -3,6 +3,7 @@ plugin_groups:
     - path: offline_imu_cam
     - path: gtsam_integrator
     - path: pose_prediction
+    - path: rk4_integrator
     - name: OpenVINS
       path:
         git_repo : https://github.com/ILLIXR/open_vins.git
--- a/configs/monado-mainline.yaml
+++ b/configs/monado-mainline.yaml
@@ -1,7 +1,7 @@
 plugin_groups:
   - plugin_group:
     - path: offline_imu_cam
-    - path: gtsam_integrator
+    #- path: gtsam_integrator
     - path: pose_prediction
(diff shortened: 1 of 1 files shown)


charmoniumQ:
I think we want to prefer `gtsam_integrator` (if it works) since that is used by `configs/native.yaml`.

hungdche:
For some reason `gtsam_integrator` does not work for me. I had to use `rk4_integrator` for both `native.yaml` and `monado-mainline.yaml`

charmoniumQ:
Interesting. I think we should switch both to `rk4_integrator` or fix `gtsam_integrator`.

What do you think @mhuzai?

mhuzai:
This is interesting. I thought you had meant that there was _no_ integrator in the monado YAML file, @hungdche, not that `gtsam_integrator` was there but wasn't working. So the issue here isn't that we need to use `rk4_integrator`, but that `gtsam_integrator` should work out of the box -- if it had, Unreal would have worked fine for you. I understand the build difficulties we've had with gtsam in the past, but can you give latest ILLIXR a shot (you would have to install dependencies again to fix the paths)?

hungdche:
Woo. Reinstalling gtsam dependencies resolves this issue. I will mark the issue as closed.

PR #334, opened 2022-04-13
Remove warnings and install realsense for ubuntu20.04 directly 

qinjunj:
Intel RealSense now supports binary package installation for Ubuntu 20.04. So I removed the warnings and directly installed the package in the script. I am not sure it supports 21.04. Should I add a message for that?


mhuzai (dismissed):

charmoniumQ:
Should we leave the warning there? The condition `!(ver == 18.04 || ver == 21.04)` is still scary.

mhuzai:
> Should we leave the warning there? The condition `!(ver == 18.04 || ver == 20.04)` is still scary.

Yeah, let's keep it.

mhuzai:
> > Should we leave the warning there? The condition `!(ver == 18.04 || ver == 20.04)` is still scary.
> 
> Yeah, let's keep it.

@qinjunj, if you could add the warning back like Sam suggested, this PR is good to go.

mhuzai:
Absorbed into #250.

PR #339, opened 2022-05-20
Separate IMU and Cam

### LINKED ISSUE #50: Separate IMU/camera Switchboard topics
Or explain why we shouldn't.
This currently complicates the ZED code (where the Camera thread has to send frames to the IMU thread, which pairs them and sends them on a Switchboard topic).
This currently simplifies OpenVINS (where we need to process the camera after the IMU).

qinjunj:
Fix rebase.

Closes #50.


mhuzai:
Closes #50.

configs/offload-server.yaml
```diff
@@ -15,7 +15,7 @@ plugin_groups:
       - name: OpenVINS
         path:
            git_repo: https://github.com/ILLIXR/open_vins.git
-           version: "3.1"
```
charmoniumQ, line 18: Let's merge this in OpenVINS to 3.2 before accepting this PR.

offline_cam/data_loading.hpp
```diff
@@ -0,0 +1,82 @@
...
+#include <opencv2/core/mat.hpp>
+#include <opencv2/imgcodecs.hpp>
+#include <opencv2/imgproc.hpp>
+#include <eigen3/Eigen/Dense>
+
+#include "csv_iterator.hpp"
+
+typedef unsigned long long ullong;
+
+//#define LAZY
```
charmoniumQ, line 15: Let's document this:

Uncommenting this preprocessor macro makes the offline_cam load each data from the disk as it is needed. Otherwise, we load all of them at the beginning, hold them in memory, and drop them in the queue as needed. Lazy loading has an artificial negative impact on performance which is absent from an online-sensor system. Eager loading deteriorates the startup time and uses more memory. I prefer eager loading.

realsense/plugin.cpp
```diff
@@ -109,65 +70,76 @@ class realsense : public plugin {
...
+                    cv::Mat ir_right = cv::Mat(cv::Size(IMAGE_WIDTH_D4XX, IMAGE_HEIGHT_D4XX), CV_8UC1, (void *)ir_frame_right.get_data());
+                    cv::Mat rgb = cv::Mat(cv::Size(IMAGE_WIDTH_D4XX, IMAGE_HEIGHT_D4XX), CV_8UC3, (void *)rgb_frame.get_data());
+                    cv::Mat depth = cv::Mat(cv::Size(IMAGE_WIDTH_D4XX, IMAGE_HEIGHT_D4XX), CV_16UC1, (void *)depth_frame.get_data());
+                    cv::Mat converted_depth;
+                    float depth_scale = pipe.get_active_profile().get_device().first<rs2::depth_sensor>().get_depth_scale(); // for converting measurements into millimeters
+                    depth.convertTo(converted_depth, CV_32FC1, depth_scale * 1000.f);
+                    _m_cam.put(_m_cam.allocate<cam_type>(
+                        {
+                            cam_time_point,
+                            cv::Mat{ir_left},
```
charmoniumQ, line 112: Is the copy constructor necessary here?

configs/offload-server.yaml
qinjunj: Yeah sure. Please have the [OpenVINS PR](https://github.com/ILLIXR/open_vins/pull/15) and the [Kimera-VIO PR](https://github.com/ILLIXR/Kimera-VIO/pull/9) reviewed together with this one.

realsense/plugin.cpp
qinjunj: I think no. Updated.

offline_cam/data_loading.hpp
qinjunj: updated.

mhuzai:
Please rebase on top of latest master. It'll be a bit painful, unfortunately.

PR #341, opened 2022-07-06
VulkanGL Interop

Jebbly:
### What
This pull request implements support for Vulkan-based OpenXR applications (it is linked to the monado_integration pull request). However, there are still some UE4-specific issues that are not yet resolved.

### How
`timewarp_gl` has been modified so that it expects a set of image handles during swapchain creation. For OpenGL-based applications, this "image handle" is just the texture that's been linked to the image. For Vulkan-based applications, this is the necessary file descriptor information to link an OpenGL to the Vulkan object.


common/data_format.hpp
```diff
@@ -174,17 +174,68 @@ namespace ILLIXR {
...
+	enum class graphics_api {
+		OPENGL,
+		VULKAN
+	};
+
+	typedef struct vk_image_handle {
+		int file_descriptor;
+		GLuint64 format;
+		GLuint64 allocation_size;
+		GLsizei width, height;
```
mhuzai, line 187: Declare one per line.

common/data_format.hpp
```diff
@@ -174,17 +174,68 @@ namespace ILLIXR {
...
+	// This is used to share swapchain images between ILLIXR and Monado.
+	// When Monado uses its GL pipeline, it's enough to just share a context during creation.
+	// Otherwise, file descriptors are needed to share the images.
+	struct image_handle : public switchboard::event {
+		graphics_api type;
+		union {
+			GLuint gl_handle;
+			vk_image_handle vk_handle;
+		};
+		// is there a better way to do this? All it does is store
```
mhuzai, line 207: In the future, perhaps we can have Monado publish all the swapchain information on a different topic. That way, we won't have to publish static information each time. But I don't see the current implementation causing any performance issues, so this is good for now!

common/data_format.hpp
```diff
@@ -174,17 +174,68 @@ namespace ILLIXR {
...
+			GLuint gl_handle;
+			vk_image_handle vk_handle;
+		};
+		// is there a better way to do this? All it does is store 
+		// the total number of swapchain images (i.e. every event has the same number)
+		// and it's only used to check if timewarp_gl has completed creating the swapchain.
+		int num_images;
+		int swapchain_index; // decides whether it's the left or right
+
+		image_handle(unsigned int gl_handle_, int num_images_, int swapchain_index_)
```
mhuzai, line 213: Why not `GLuint` for `gl_handle_`?

common/data_format.hpp
```diff
@@ -174,17 +174,68 @@ namespace ILLIXR {
...
+		graphics_api type;
+		union {
+			GLuint gl_handle;
+			vk_image_handle vk_handle;
+		};
+		// is there a better way to do this? All it does is store 
+		// the total number of swapchain images (i.e. every event has the same number)
+		// and it's only used to check if timewarp_gl has completed creating the swapchain.
+		int num_images;
+		int swapchain_index; // decides whether it's the left or right
```
mhuzai, line 211: *left or right swapchain

common/data_format.hpp
```diff
@@ -174,17 +174,68 @@ namespace ILLIXR {
...
+			, swapchain_index{swapchain_index_}
+		{}
+	};
+
 	// Using arrays as a swapchain
 	// Array of left eyes, array of right eyes
 	// This more closely matches the format used by Monado
 	struct rendered_frame : public switchboard::event {
-		std::array<GLuint, 2> texture_handles; // Does not change between swaps in swapchain
+		std::array<int, 2> texture_handles; // Does not change between swaps in swapchain
```
mhuzai, line 232: This is because we no longer publish the handles themselves but rather indices into the swapchain, right? If so, rename the variable to something more appropriate.

timewarp_gl/plugin.cpp
```diff
@@ -64,7 +65,27 @@ class timewarp_gl : public threadloop {
...
-	{ }
+	{ 
+		image_handles_ready = false;
+		swapchain_ready = false;
+		sb->schedule<image_handle>(id, "image_handle", [this](switchboard::ptr<const image_handle> handle, std::size_t) {
+			if (handle->swapchain_index > 1) {
+				return;
+			}
+			
+			assert(handle->swapchain_index == 0 || handle->swapchain_index == 1);
```
mhuzai, line 76: More than two swapchains are actually possible. For example, Varjo headsets have two displays per eye -- an inner one and an outer one -- and each gets its own swapchain. Therefore, there are a total of four swapchains. Of course, supporting such a display would probably require some other changes too, so leave a comment here saying we only support two swapchains for now.

timewarp_gl/plugin.cpp
```diff
@@ -64,7 +65,27 @@ class timewarp_gl : public threadloop {
...
+
+
+			// If the graphics backend is OpenGL, Monado's OpenGL context should be
+			// shared with ILLIXR, so the buffers are the same.
```
mhuzai, line 80: Could you explain a little bit more what this comment is trying to say?

timewarp_gl/plugin.cpp
```diff
@@ -64,7 +65,27 @@ class timewarp_gl : public threadloop {
...
+			assert(handle->swapchain_index == 0 || handle->swapchain_index == 1);
+
+
+			// If the graphics backend is OpenGL, Monado's OpenGL context should be
+			// shared with ILLIXR, so the buffers are the same. 
+			this->_m_image_handles[handle->swapchain_index].push_back(*handle);
+
+			if (this->_m_image_handles[0].size() == (size_t) handle->num_images &&
+				this->_m_image_handles[1].size() == (size_t) handle->num_images) {
+				this->image_handles_ready = true;
```
mhuzai, line 85: There is a data race here. `image_handles_ready` is being written here and being read on line 428. Since the two are happening in two different threads, and the two threads can be running in parallel, and there is no other synchronization between the two threads, `image_handles_ready` needs to be an atomic bool.

timewarp_gl/plugin.cpp
```diff
@@ -85,6 +106,12 @@ class timewarp_gl : public threadloop {
 
 	static constexpr std::chrono::nanoseconds vsync_period {freq2period(DISPLAY_REFRESH_RATE)};
 
+	// Shared image handles between ILLIXR and the application
+	graphics_api client_backend;
+	std::array<std::vector<image_handle>, 2> _m_image_handles;
+	std::array<std::vector<GLuint>, 2> _m_swapchain; 
+	bool image_handles_ready, swapchain_ready;
```
mhuzai, line 113: Declare one per line.

timewarp_gl/plugin.cpp
```diff
@@ -223,6 +254,45 @@ class timewarp_gl : public threadloop {
 		return pixels;
 	}
 
+	// this is the inverse of what they do in Monado
+	// is there a better way to handle this?
```
mhuzai, line 258: Looks good to me.

timewarp_gl/plugin.cpp
```diff
@@ -499,6 +571,23 @@ class timewarp_gl : public threadloop {
         [[maybe_unused]] const bool gl_result = static_cast<bool>(glXMakeCurrent(xwin->dpy, xwin->win, xwin->glc));
 		assert(gl_result && "glXMakeCurrent should not fail");
 
+		if (!swapchain_ready) {
+			assert(image_handles_ready);
+
+			client_backend = _m_image_handles[0][0].type;
+			for (int i = 0; i < 2; i++) {
```
mhuzai, line 578: Rename to something more expressive; e.g., `eye` or something.

timewarp_gl/plugin.cpp
```diff
@@ -499,6 +571,23 @@ class timewarp_gl : public threadloop {
...
+				int num_images = _m_image_handles[i][0].num_images;
+				for (int j = 0; j < num_images; j++) {
```
mhuzai, line 580: Rename to something more expressive.

timewarp_gl/plugin.cpp
```diff
@@ -223,6 +254,45 @@ class timewarp_gl : public threadloop {
...
+			case VK_FORMAT_B8G8R8A8_SRGB: {
+				swizzle_mask[0] = GL_BLUE;
+				swizzle_mask[2] = GL_RED;
+			}
+			case VK_FORMAT_R8G8B8A8_SRGB:  return GL_SRGB8_ALPHA8;
+			default: return 0;
+		}
+	}
+
+	void VulkanGLInterop(const vk_image_handle& vk_handle, int swapchain_index) {
```
mhuzai, line 271: Nicely done!

One minor comment is that both `glGenTextures()` and `glCreateMemoryObjectsEXT()` can create multiple handles in a single call, so we can potentially reduce the number of GL calls if we refactor the code. But this is super minor, and mostly just a general tip.

mhuzai (commented):
Left a few minor comments. The overall structure looks great!

timewarp_gl/plugin.cpp
Jebbly: Sorry, that was an old comment - I meant to delete it earlier.

mhuzai:
Just rebase on top of latest master and this is good to go!

runner/runner/main.py
```diff
@@ -284,44 +289,61 @@ def process_plugin(plugin_config: Mapping[str, Any]) -> Path:
...
-            ## Clean up leftover socket. It can only either be in $XDG_RUNTIME_DIR or /tmp
-            Path(env_monado_service['XDG_RUNTIME_DIR'] + "/monado_comp_ipc").unlink(missing_ok=True)
-            Path("/tmp/monado_comp_ipc").unlink(missing_ok=True)
-
-        print("\nstdout:\n")
-        sys.stdout.buffer.write(outs)
-
-        print("\nstderr:\n")
-        sys.stderr.buffer.write(errs)
+    env_list = [f"{shlex.quote(var)}={shlex.quote(val)}" for var, val in env_override.items()]
```
qinjunj, line 311: As we discussed, change the runner to start the list of applications that the user specifies in the openxr_app.yaml. Add instructions about how to manually start multiple applications. Finally, clean the code that allows gdb and add instructions about that.

qinjunj (commented):
Tested with native and openxr_simple_example, but would like more applications to test.

Jebbly:
The initial script to add GDB to monado-service still requires some examination. Right now, ``subprocess_run`` is a blocking call, so it blocks the script from being able to automatically run any other OpenXR applications. A possible hacky solution would be to launch a thread for each OpenXR app, and sleep for some arbitrary time (trusting that the service will be running by then), before running the service. Unfortunately, ``subprocess.Popen`` doesn't seem viable to use along with GDB.

Also, at the moment, running ``monado.yaml`` in ``dbg`` mode also throws the following exception (needs further debugging/):
```
{monado_integration}/src/xrt/compositor/main/comp_target_swapchain.c:945: void comp_target_swapchain_mark_timing_point(struct comp_target *, enum comp_target_timing_point, int64_t, uint64_t): Assertion `frame_id == cts->current_frame_id' failed.
```

Jebbly:
Here's an updated ZIP of the packaged Unreal demo: https://uofi.box.com/s/ssgbcaaemrlbsu61vyv7z7jhkrivzje0.

Jebbly:
Closing with a more updated version in #373 .

PR #346, opened 2022-07-24
Issue 345 fix realsense

jianxiapyh:


realsense/plugin.cpp
```diff
@@ -39,7 +38,10 @@ class realsense : public plugin {
         // Even if the API does not invoke `callback` in parallel, this is still important for the memory-model.
         // Without this lock, prior invocations of `callback` are not necessarily "happens-before" ordered, so accessing
         // persistent variables constitutes a data-race, which is undefined behavior in the C++ memory model.
-
+        if (!_m_clock->has_started()) {
+            // if the data arrived before the relative clock started, ignore
+            return;
+        }
```
mhuzai, line 44: Now that you've moved `configure_camera()` to `start()`, the clock is guaranteed to be ready before the first callback occurs. Or am I missing something?

common/relative_clock.hpp
```diff
@@ -116,6 +116,13 @@ class RelativeClock : public phonebook::service {
             std::chrono::nanoseconds{relative.time_since_epoch()}.count();
     }
 
+    /**
+     * @brief Starts Check if the clock has started.
+     */
+    bool has_started() const {
+        return _m_start > std::chrono::steady_clock::time_point{};
+    }
+
```
mhuzai, line 125: Remove if comment in `plugin.cpp` is correct.

realsense/plugin.cpp
```diff
@@ -109,8 +111,8 @@ class realsense : public plugin {
                 rs2_vector        gyro_data = gyro.get_motion_data();
 
                 // IMU data
-                Eigen::Vector3f la = {accel.x, accel.y, accel.z};
-                Eigen::Vector3f av = {gyro_data.x, gyro_data.y, gyro_data.z};
+                Eigen::Vector3f la = {accel.z, accel.x, accel.y};
+                Eigen::Vector3f av = {gyro_data.z, gyro_data.x, gyro_data.y};
```
mhuzai, line 115: Add a comment explaining why the order is (z, x, y) and not (x, y, z).

pose_prediction/plugin.cpp
```diff
@@ -178,7 +178,7 @@ class pose_prediction_impl : public pose_prediction {
         swapped_pose.position.z() = -pose.position.x();
 
         // Make any chanes to orientation of the output below
-        // For the dataset were currently using (EuRoC), the output orientation acts as though
+        // For the SLAM we are using (OpenVINS), the output orientation acts as though
```
mhuzai, line 181: Both OpenVINS and Kimera.

mhuzai (commented):
Nice fix!

realsense/plugin.cpp
qinjunj: Relative clcok is currently started after all plugins are [started](https://github.com/ILLIXR/ILLIXR/blob/024494818bd7834a0c70d08d58bd2746ffc512a7/runtime/runtime_impl.hpp#L64-L68). So moving that `configure_camera()` to `start()` won't fix the problem as the first callback might already happen before the clock starts. Instead, in all plugins inherited from threadloop, we are using [`Stoplight`](https://github.com/ILLIXR/ILLIXR/blob/024494818bd7834a0c70d08d58bd2746ffc512a7/common/threadloop.hpp#L89) to guarantee the start of Relative Clock, but nothing like that exists for non-threadloop plugins like RealSense. We can add a "if started" checking to Relative Clock for future use cases, which I doubt if there would be many (perhaps only for sensor plugins like RealSense that is neither evoked by any other plugins nor inherited from threadloop).

mhuzai: My bad. I was thinking of `threadloop->setup()` when I wrote the above. I see your point. The above solution will definitely work, but I'm not sure it's the cleanest way from a design perspective. In some ways, we've seen this issue before, and I'm wondering if this is a good time to revisit [that discussion](https://github.com/ILLIXR/ILLIXR/pull/250#discussion_r858174782).

charmoniumQ: I think two stoplight events is the best solution.

jianxiapyh:
superseded by #353

PR #350, opened 2022-08-24
Update CUDA key in install_deps, plus minor documentation fix

### LINKED ISSUE #349: CUDA Repo Key Changes
Nvidia updated their signing key for the cuda repository earlier this year (see [here](https://developer.nvidia.com/blog/updating-the-cuda-linux-gpg-repository-key/)). This is causing some issues when running the dependency installer, since we have hardcoded their old key.

chaseblock:
Nvidia changed their keys for cuda earlier this year, and I think this update should fix that for our dependency installation script.

Also, a minor typo in the documentation.

Closes #349

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -16,6 +16,7 @@
 
 Sarita V. Adve <sadve@illinois.edu>
 Dhairya Bahl <dhairyabahl5@gmail.com>
+Charles Block <coblock2@illinois.edu>
 Henry Che <hungdc2@illinois.edu>
 Rishi Desai <rishipd2@illinois.edu>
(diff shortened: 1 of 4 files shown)


charmoniumQ:
This seems fine, but try to keep the PRs centered on a single subject.

charmoniumQ (approved):

mhuzai:
> This seems fine, but try to keep the PRs centered on a single subject.

I agree that that is the right thing in general, but it's okay in this case because Chase found several minor bugs while setting up ILLIXR, and this is going to be a miscellaneous bug fix PR.

mhuzai (approved):

PR #354, opened 2022-09-20
RealSense Plugin Bug Fixes

### LINKED ISSUE #345: Fixing Realsense plugin bugs
Three issues 
1. IMU coordinate misalignment 
Currently, the imu coordinate system (x left, y down, z forward)  is different from the right-hand coordinate system used by OpenVINS (x forward, y left, z up).  This fix applies transformation at the reading data phase in the RealSense plugin so that the headset rotation reflects the corresponding RealSense rotation.
2. RealSense plugin starts a pipeline that streams camera images and imu data. However, this stream can arrive earlier before the RelativeClock->start() is called. I am incorporating Qinjun's fix from branch fix-realsense-timer in this fix. 
3. In the pose_prediction plugin's correct_pose function, I think the comment for Euroc is misleading. This conversion is strictly between OpenVINs coordinate system (x forward, y left, z up) and OpenGL's coordinate system (left-hand system, x right, y up, z backward). It has nothing to do with Euroc specifically. We need this transformation as long as we are using OpenVINS.

jianxiapyh:
closes #345


pose_prediction/plugin.cpp
```diff
@@ -33,6 +39,11 @@ class pose_prediction_impl : public pose_prediction {
         }
     }
 
+    ~pose_prediction_impl() {
+        // pose_output_post.close();
+        // pose_output_post_offset.close();
+    }
+
```
charmoniumQ, line 46:
```suggestion
```
Remove empty constructor if not needed.
In particular, I believe those files will be closed in the default destructor.

configs/native.yaml
```diff
@@ -1,10 +1,10 @@
 plugin_groups:
   - !include "rt_slam_plugins.yaml"
   - plugin_group:
-      - path: ground_truth_slam/
+      #- path: ground_truth_slam/
```
charmoniumQ, line 4: Make sure to restore the config files to their previous state before merging this PR.

jianxiapyh:
I am checking out pull request #379 , if that works, this can be closed

PR #365, opened 2022-12-02
Add support for OpenNI

hungdche:
Add `openni` plugin, which writes to `rgb_depth_type` topic.


openni/plugin.cpp
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+protected:
+    virtual skip_option _p_should_skip() override {
+        // read cam data
+        color.readFrame(&colorFrame);
+        depth.readFrame(&depthFrame);
+        // get timestamp
+        assert(colorFrame.getTimestamp() != depthFrame.getTimestamp());
+        cam_time = colorFrame.getTimestamp() * 1000;
```
charmoniumQ, line 41: Try to avoid doing work in `_p_should_skip` and do it in `_p_one_iteration` instead. I would try to use the C++ clock instead of the sensor timestamp to tell if it is time to take another picture, but there is some chance that won't work. In that case, look for a lightweight way to get the timestamp from the sensor other than taking a whole picture.

openni/plugin.cpp
```diff
@@ -0,0 +1,159 @@
...
+        if (cam_time > last_ts) {
+            std::this_thread::sleep_for(std::chrono::milliseconds{2});
+            return skip_option::run;
+        } else {
+            return skip_option::skip_and_yield;
+        }
+    }
+
+    virtual void _p_one_iteration() override {
+        RAC_ERRNO_MSG("zed at start of _p_one_iteration");
```
charmoniumQ, line 52: Change `zed` to `openni`.

hungdche:
This PR is moved to #370.

PR #370, opened 2023-01-20
Openni plugin revert commits

hungdche:
Revert overriding commits in #365 
Add `openni plugin`, which writes to `rgb_depth_type` topic.


openni/CMakeLists.txt
```diff
@@ -0,0 +1,20 @@
+# OpenCV
+find_package(OpenCV REQUIRED)
```
jianxiapyh, line 2: in order to be consistent with other plugins, you might want to specify the OpenCV version

common/data_format.hpp
```diff
@@ -41,12 +41,11 @@ struct imu_type : switchboard::event {
         , linear_a{linear_a_} { }
 };
 
-class rgb_depth_type : public switchboard::event {
+struct rgb_depth_type : public switchboard::event {
     [[maybe_unused]] time_point time;
     cv::Mat                     rgb;
```
jianxiapyh, line 46: maybe in the future, we might have to separate rgb and depth as well.

jianxiapyh (dismissed):
just change the OpenCV format for the CMake is sufficient.

openni/plugin.cpp
```diff
@@ -0,0 +1,160 @@
+#include "../common/data_format.hpp"
+#include "../common/phonebook.hpp"
+#include "../common/pose_prediction.hpp"
```
Boyuan-Tian, line 3: Not sure why is pose prediction needed here.

openni/plugin.cpp
```diff
@@ -0,0 +1,160 @@
...
+        colorMat.create(colorFrame.getHeight(), colorFrame.getWidth(), CV_8UC3);
+        const openni::RGB888Pixel* colorBuffer = (const openni::RGB888Pixel*) colorFrame.getData();
+        memcpy(colorMat.data, colorBuffer, 3 * colorFrame.getHeight() * colorFrame.getWidth() * sizeof(uint8_t));
+        cv::cvtColor(colorMat, colorMat, cv::COLOR_BGR2BGRA);
+        colorMat.convertTo(colorMat, CV_8UC4);
+
+        cv::Mat depthMat;
+        depthMat.create(depthFrame.getHeight(), depthFrame.getWidth(), CV_16UC1);
+        const openni::DepthPixel* depthBuffer = (const openni::DepthPixel*) depthFrame.getData();
+        memcpy(depthMat.data, depthBuffer, depthFrame.getHeight() * depthFrame.getWidth() * sizeof(uint16_t));
```
Boyuan-Tian, line 66: What is the unit of captured depth image?
The loaded depth pixel should be converted to meters or millimeters instead of the raw values.
Probably want to run that conversion before publish the depth image.

openni/plugin.cpp
```diff
@@ -0,0 +1,160 @@
...
+
+#ifndef NDEBUG
+        // get depth options
+        const openni::SensorInfo*               depthInfo  = device.getSensorInfo(openni::SENSOR_DEPTH);
+        const openni::Array<openni::VideoMode>& modesDepth = depthInfo->getSupportedVideoModes();
+        for (int i = 0; i < modesDepth.getSize(); i++) {
+            printf("Depth Mode %i: %ix%i, %i fps, %i format\n", i, modesDepth[i].getResolutionX(),
+                   modesDepth[i].getResolutionY(), modesDepth[i].getFps(), modesDepth[i].getPixelFormat());
+        }
+        rc = depth.setVideoMode(modesDepth[4]);
```
Boyuan-Tian, line 105: What are the modesDepth?
Is this the only mode we want to use for a particular OpenNI device?
Is this hardcoded No.4 applicable to all other OpenNI supported devices?

openni/plugin.cpp
```diff
@@ -0,0 +1,160 @@
...
+
+#ifndef NDEBUG
+        // get color options
+        const openni::SensorInfo*               colorInfo  = device.getSensorInfo(openni::SENSOR_COLOR);
+        const openni::Array<openni::VideoMode>& modesColor = colorInfo->getSupportedVideoModes();
+        for (int i = 0; i < modesColor.getSize(); i++) {
+            printf("Color Mode %i: %ix%i, %i fps, %i format\n", i, modesColor[i].getResolutionX(),
+                   modesColor[i].getResolutionY(), modesColor[i].getFps(), modesColor[i].getPixelFormat());
+        }
+        rc = color.setVideoMode(modesColor[9]);
```
Boyuan-Tian, line 128: Same as above.

debugview/plugin.cpp
```diff
@@ -288,6 +314,7 @@ class debugview : public threadloop {
         mouse_velocity = mouse_velocity * 0.95;
 
         load_camera_images();
+        load_rgb_depth();
```
Boyuan-Tian, line 317: What if both `cam_type` and `rgb_depth_type` are getting published at the same time?
Seems to me that sequentially running these two functions overrides whatever loaded from `cam_type`.

Boyuan-Tian:
I am not able run this branch on my machine with openni plugin enabled and offline_imu_cam disabled, mainly because OpenNI can not be found correctly.
Not sure if this is an issue with my machine or compiling script.
@jianxiapyh Can you run a test on your side?
@hungdche Also we will need to update installation script if OpenNI is not currently included.

openni/plugin.cpp
hungdche: Line 102-103 prints out all of the mode and resolution of Depth and the associate index, 4 is just the one that we use for ORB-SLAM + ITM. I will set it back to 0, and have a comment specifying what it does

debugview/plugin.cpp
hungdche: You're right. I'll make changes accordingly.

hungdche:
@Boyuan-Tian I had to specify the path that contains OpenNI [here](https://github.com/ILLIXR/ILLIXR/blob/5c424e64b79e1816c58c92e7d719e04506dc7c09/openni/CMakeLists.txt#L5).

openni/plugin.cpp
hungdche: The hardcoded No.4 specifies that we wants to run depth mode at 640x480 res and 30 fps. I'm not sure if they applies to all OpenNI supported device, hence I have [this line](https://github.com/ILLIXR/ILLIXR/blob/8f40cf21d34bac14cf125c0758329becab95718c/openni/plugin.cpp#L102) print all the available mode out, which must then be set manually by the user. I will set them back to default.

openni/README.md
```diff
@@ -0,0 +1,27 @@
...
+Device open failed: 	Could not open "1d27/0601@3/2": USB transfer timeout!
+```
+**Solution**: This is usually fixed by unplug and plug back in.
+
+```
+Device open failed: 	Could not open "1d27/0601@3/2": Failed to open the USB device!
+```
+**Solution**: Fix by running with sudo 
+
+### Note
```
jianxiapyh, line 21: might need to add a note if the user forget to use install boost

record_rgb_depth/Makefile
```diff
@@ -0,0 +1,3 @@
+LDFLAGS = $(shell pkg-config opencv --libs) -lboost_filesystem
```
jianxiapyh, line 1: Are you using the boost provided by ILLIXR?

openni/README.md
hungdche: I dont think OpenNI uses Boost though

openni/plugin.cpp
hungdche: Have verified that OpenNI publishes depth image in mm.

record_rgb_depth/Makefile
hungdche: ILLIXR installs all libraries of boost, including filesystem. I have never manually installed boost standalone, so I guess that one is the boost provided by ILLIXR.

debugview/plugin.cpp
```diff
@@ -195,16 +196,35 @@ class debugview : public threadloop {
...
-        ImGui::Begin("Onboard camera views");
-        auto windowSize     = ImGui::GetWindowSize();
-        auto verticalOffset = ImGui::GetCursorPos().y;
-        ImGui::Image((void*) (intptr_t) camera_textures[0], ImVec2(windowSize.x / 2, windowSize.y - verticalOffset * 2));
-        ImGui::SameLine();
-        ImGui::Image((void*) (intptr_t) camera_textures[1], ImVec2(windowSize.x / 2, windowSize.y - verticalOffset * 2));
-        ImGui::End();
+        if (use_cam) {
+            ImGui::SetNextWindowSize(ImVec2(700, 350), ImGuiCond_Once);
+            ImGui::SetNextWindowPos(ImVec2(ImGui::GetIO().DisplaySize.x, ImGui::GetIO().DisplaySize.y), ImGuiCond_Once,
```
jianxiapyh, line 201: It seems that there are a lot of redundant code between use_cam and use_rgbd

hungdche: use cam is mostly for stereo while use_rgb is for RGBD. I made these based on @Boyuan-Tian's comments on what if there are multiple streams of images (say both EuRoC and OpenNI). Those code would handle that.

jianxiapyh: I think you are missing my point. I was thinking about encapsulate settings into variables, something like use_cam -> display window (camera_texture) and use_rgbd->displaywindow(rgbd_texture)

Boyuan-Tian (approved):
LGTM.

jianxiapyh (dismissed):
Just perform the changes we discussed

jianxiapyh (approved):

Boyuan-Tian (approved):

astro-friedel:
This PR has been merged into PR #384.

PR #373, opened 2023-02-11
Monado Integration

Jebbly:
# What

This PR modifies timewarp_gl to render into two textures/framebuffers. This helps to generalize for the output that's expected from the Monado layer renderer. This is a more updated PR for #341.

This PR should be merged with [the corresponding monado_integration PR](https://github.com/ILLIXR/monado_integration/pull/18).

## Notes

There is a macro in timewarp_gl called ``USE_ALT_EYE_FORMAT``, which we may be able to cleanup with this more general approach. Could be done in this PR, or as a smaller issue.


common/data_format.hpp
```diff
@@ -147,21 +147,92 @@ typedef struct {
...
 } fast_pose_type;
 
+// Used to identify which graphics API is being used (for swapchain construction)
+enum class graphics_api { OPENGL, VULKAN, TBD };
+
+// Used to distinguish between different image handles
+enum class swapchain_usage { LEFT_SWAPCHAIN, RIGHT_SWAPCHAIN, LEFT_RENDER, RIGHT_RENDER, NA };
+
+typedef struct vk_image_handle {
+    int      file_descriptor;
```
shg8, line 157: Why is this a file descriptor?

shg8 (approved):

common/data_format.hpp
Jebbly: The naming is a little unfortunate because it's not a Vulkan descriptor, it's Linux's native resource handle. Maybe a typedef would be better here if we ever want to better support other platforms.

astro-friedel:
Closing this PR as it will be superseded by another one.

PR #363, opened 2022-11-25
Rebase issue 50 separate imu cam revert repeated commits

qinjunj:
Fix rebase in #339 and revert repeated commits.
Adapts offload-vio to separated imu and cam; update to the TCP implementation.

Close #339.

PR #375, opened 2023-03-02
Issue 371 configurable parameter

EliFeinberg:
Adds in features to runner and runtime to accommodate global variables defined from the yaml global_module_def


common/data_format.hpp
```diff
@@ -17,37 +17,46 @@
 #define USE_ALT_EYE_FORMAT
 
 namespace ILLIXR {
-using ullong = unsigned long long;
 
-struct cam_type : switchboard::event {
```
qinjunj, line 22: Why reverting commit #363 ?

common/data_format.hpp
```diff
@@ -20,23 +20,15 @@ namespace ILLIXR {
 
 using ullong = unsigned long long;
 
-// Data type that combines the IMU and camera data at a certain timestamp.
-// If there is only IMU data for a certain timestamp, img0 and img1 will be null
-// time is the current UNIX time where dataset_time is the time read from the csv
-struct imu_cam_type : public switchboard::event {
```
qinjunj: #363 is a large commit that contains a lot of changes. You may need to rebase your changes on top of the current master, and make sure the `revert #363` is not included.

common/error_util.hpp
```diff
@@ -38,12 +38,12 @@ inline void report_and_clear_errno([[maybe_unused]] const std::string& file, [[m
                                    [[maybe_unused]] const std::string& function, [[maybe_unused]] const std::string& msg = "") {
 #ifndef NDEBUG
     if (errno > 0) {
-        if (ILLIXR::ENABLE_VERBOSE_ERRORS) {
+        // if (ILLIXR::ENABLE_VERBOSE_ERRORS) {
```
qinjunj, line 41: `ENABLE_VERBOSE_ERRORS` is still defined above. We want to keep this config variable. Please change it according to the new way of defining configs.

offload_data/plugin.cpp
```diff
@@ -43,8 +47,8 @@ class offload_data : public plugin {
 
     virtual ~offload_data() override {
         // Write offloaded data from memory to disk
-        if (enable_offload) {
-            boost::filesystem::path p(obj_dir);
+        if (_m_enable_offload) {
+            boost::filesystem::path p(obj_dir); // maybe replace with _m_obj_dir
```
qinjunj, line 51: Yes, `OFFLOAD_PATH` is defined in `configs/global_module_defs.yaml` as also `metrics/offloaded_data/`, so it's replaceable.

qinjunj (commented):
For constant definitions in `common/const_decls.hpp`, how about we put their default values rather than `dev/null`?

astro-friedel:
This PR will be superceded by upcoming switchboard updates and changes

Issue #306, opened 2021-06-03
offline imu support for more datasets


jianxiapyh:
Currently offline imu is hardcoded for EuRoC  datasets, working towards an abstraction layer to support more datasets(TUM RGB-D for example)

charmoniumQ:
Consider using SQLite. You can import CSVs easily. However, SQLite can do much more than CSV, and it is much faster. You can have an arbitrary number of tables containing metadata (camera intrinsics, etc.).

jianxiapyh:
Thanks! I will give it a try

PR #379, opened 2023-04-15
Adding OpenVINS Internal Plugin

goldbattle:
This as a plugin that directly interfaces with OpenVINS. Not sure if this is what you guys desire, but I think the plugin could be split out to a small repository if needed. There are a couple key changes that I have made which make this new setup nice.

- Config files for sensors are local to the plugin, thus it is easy to simply add a ZED Kalibr calib as a yaml file and not change any code
- This is compiled against the main upstream OpenVINS, uses some pretty stable APIs, and thus should stay in sync if the main repo gets update (the `deps.sh` version would need to be updated and `./install_deps.sh` would need to be re-run to get the new code)
- This plugin ensures that we "drop" image messages if OpenVINS takes a long time to update, but the IMU will not be dropped and should always be appended. This is implemented via spawning a second thread to process the image.
- Support for VIO using a single camera has also be added, and OpenVINS can be configured by setting `max_cameras: 1` in the sensor yaml file to only take the left image even on stereo datasets like EuRoC.

I have also added directly publishing of poses at the rate of the IMU using the built in OpenVINS fast propagate. This can be disabled and the GTSAM / RK4 integrator could be re-added, but this seems redundant to me. See the cmake file for the define which can disable it.
https://github.com/goldbattle/ILLIXR/blob/ee710ca5b110e827303adae00fbc92d5493a8d4a/pose_openvins/CMakeLists.txt#L18-L20

I have also included the ability to download google drive links in the dataset which allows for running some example [datasets](https://github.com/rpng/ar_table_dataset) we have collected using a T265 sensor. This also required supporting reading of the [kalibr zip format](https://github.com/ethz-asl/kalibr/wiki/bag-format#bagcreater) in the camera, imu, and groundtruth publishers. This should help address https://github.com/ILLIXR/ILLIXR/issues/306.

Additionally, I have added the OpenVINS tracking image to the debug view to help users understand the status of VIO.
Screenshot from 2023-04-15 19-30-24

A ReadMe with some further details has been added to the plugin directory which I hope helps people run on their own sensors.
https://github.com/goldbattle/ILLIXR/blob/ee710ca5b110e827303adae00fbc92d5493a8d4a/pose_openvins/README.md


goldbattle:
Here is a demo of running on both datasets:
[](https://youtu.be/6C_aQMrWmqc)

And running on a live realsense sensor
[](https://youtu.be/1-cLoqwUrYY)

One can see there is some problems with the realsense plugin with very bad jitter in the rendered view.
The pose looks pretty good, but the rendered view has very bad jitter.
Not sure if a timestamp or the IMU to headset transformation.

qinjunj:
Merge to the branch in ILLIXR.

Issue #360, opened 2022-11-07
Memory leak in offline_imu_cam plugin


yavuz650:
I believe the code below leaks memory,

https://github.com/ILLIXR/ILLIXR/blob/38ef2007db0b3626900fd2a06371dc184c1315de/offline_imu_cam/plugin.cpp#L70
https://github.com/ILLIXR/ILLIXR/blob/38ef2007db0b3626900fd2a06371dc184c1315de/offline_imu_cam/plugin.cpp#L74

The unique pointer is released but is not reassigned, therefore the underlying object has no owner and it will not be destroyed.

charmoniumQ:
This seems credible. Perhaps `lazy_load_image::load()` should return a raw `cv::Mat`.

```diff
+    cv::Mat load() const {
+        return cv::imread(_m_path, cv::IMREAD_GRAYSCALE);
-    std::unique_ptr<cv::Mat> load() const {
-        auto img = std::unique_ptr<cv::Mat>{new cv::Mat{cv::imread(_m_path, cv::IMREAD_GRAYSCALE)}};
-        assert(!img->empty());
-        return img;
}
```

and 

```diff
        std::optional<cv::Mat> cam0 =
-            sensor_datum.cam0 ? std::make_optional<cv::Mat>(*(sensor_datum.cam0.value().load().release())) : std::nullopt;
+            sensor_datum.cam0 ? std::make_optional<cv::Mat>(sensor_datum.cam0.value().load()) : std::nullopt;

        std::optional<cv::Mat> cam1 =
-            sensor_datum.cam0 ? std::make_optional<cv::Mat>(*(sensor_datum.cam0.value().load().release())) : std::nullopt;
+            sensor_datum.cam1 ? std::make_optional<cv::Mat>(sensor_datum.cam1.value().load()) : std::nullopt;
```

yavuz650:
I guess that would solve the leak. An alternative could be just avoiding `release()`, e.g.

```diff
        std::optional<cv::Mat> cam0 =
-            sensor_datum.cam0 ? std::make_optional<cv::Mat>(*(sensor_datum.cam0.value().load().release())) : std::nullopt;
+            sensor_datum.cam0 ? std::make_optional<cv::Mat>(*(sensor_datum.cam0.value().load())) : std::nullopt;

        std::optional<cv::Mat> cam1 =
[...]

PR #372, opened 2023-01-25
Remove offline imu cam

qinjunj:
Remove the redundant `offline_imu_cam` directory and update documentations.

PR #382, opened 2023-05-04
Address memory leak in offline imu cam plugin

jianxiapyh:
addresses issue #360


jianxiapyh:
closed because this will be removed with issue #372

PR #381, opened 2023-05-03
Fixed spelling mistake in CMakeLists.txt

wermos:
It said `CMAKE_CXX_FLAGE_RELEASE`, which I fixed.

PR #384, opened 2023-05-12
Conversion of build system to CMake

astro-friedel:
The build system has been converted from a set of shell scripts to a CMake build system. The code should build on the following systems

- Fedora 37+
- Centos Stream9
- Ubuntu 20.04, 22.04 (18.04 is NOT supported because it is effectively EOL)

The following build system changes were implemented:

- OS repo based dependencies must be installed by the user prior to running `cmake`. The only exception to this is OpenCV in CentOS. The OS repo does not provide the full version (the viz module is missing), in this instance CMake will pull the appropriate version from the OpenCV git repo and build and use it.
- Non-OS repo based dependencies (e.g. DBoW2) are included as external packages in CMake. they are built and installed in the install prefix given for ILLIXR.
- External plugins (e.g. Kimera_VIO) are also treated as external packages and build in the same fashion as the non_OS based dependencies.
- All plugins rely on the same versions of OS-based dependencies (e.g. some plugins used a version of Eigen that was 15 years old while others used one that was 5 years old)
- Both gcc and clang should work for compilation

The following code changes were implemented:

- Due to changes in some dependency APIs some plugin code was updated to work with the newest versions, some of these changes are backward compatible with older but still supported API versions
- The "main.cpp" code was updated to be able to take command line arguments for which plugins to load, as well as arguments which were had been specified in environment variables (which are still supported), and a yaml file which contains any of these parameters.

As far as the parameters for the executable go, they will follow this priority:

1. If the parameter is specified on the command line it will have the highest priority
2. If the parameter has a value in the yaml file this value will only be used if it is not specified on the command line (second priority)
3. If the parameter has a value as an environment variable this value will only be used if it is not specified on the command line nor yaml file

Note that not everything may be working completely at this point, but I felt that the build system updates should be tested out as the details are finished up.

Also note that the cmake arguments for which plugins to build can be found [here](https://astrochem.web.illinois.edu/illixr.html) (temporary home)


charmoniumQ:
Don't forget to update documentation pages: Building ILLIXR, Modifying a plugin, Writing your own plugin. Write somewhere in the documentation the configuration places and precedence.

Maybe keep some of `install_apt_deps.sh`, since we are passing off installing OS-repo-based dependencies to the user.

CMakeLists.txt
```diff
@@ -0,0 +1,328 @@
...
+endif()
+
+if (POLICY CMP0077)
+    cmake_policy(SET CMP0077 NEW)
+endif()
+
+set(BUILDING_ILLIXR Yes)
+
+set(PACKAGE_NAME "ILLIXR")
+set(PACKAGE_STRING "ILLIXR 1.0.0")
```
shg8, line 45: We can set this to 3.2.0 since this would be our next major release

CMakeLists.txt
```diff
@@ -0,0 +1,328 @@
...
+endif()
+
+if(USE_OFFLOAD_DATA)
+    find_package(Boost REQUIRED COMPONENTS filesystem)
+elseif(USE_ORB_SLAM)
+    find_package(Boost REQUIRED COMPONENTS filesystem serialization)
+endif()
+
+if(USE_GTSAM_INTEGRATOR OR USE_ORB_SLAM)
+    find_package(Eigen3 REQUIRED)
```
shg8, line 255: I think more plugins use Eigen3 - for instance, gldemo and timewarp.

shg8 (commented):
For the offload plugin, it would also need gstreamer. Right now this is installed manually by the user. A lot of the distro version of gstreamer is outdated, so, preferably, we would want to pull a branch from their repo [here](https://gitlab.freedesktop.org/gstreamer/gstreamer) and link statically to it. Offloading currently use gstreamer's base and nvcodec plugins.

CMakeLists.txt
```diff
@@ -0,0 +1,328 @@
...
+string(TOLOWER "${CMAKE_BUILD_TYPE}" lower_type)
+if(lower_type MATCHES "debug")
+    set(ILLIXR_BUILD_SUFFIX ".dgb")
+elseif(lower_type MATCHES "release")
+    set(ILLIXR_BUILD_SUFFIX ".opt")
+elseif(lower_type MATCHES "relwithdebinfo")
+    set(ILLIXR_BUILD_SUFFIX ".optdbg")
+endif()
+set(CMAKE_CXX_FLAGS_DEBUG "-ggdb -Wall -Wextra -fPIC")
+set(CMAKE_CXX_FLAGE_RELEASE "-O3 -DNDEBUG -Wall -Wextra -fPIC")
```
shg8, line 90: Typo here

CMakeLists.txt
astro-friedel: Done

CMakeLists.txt
astro-friedel: Yes, there are several more. Some that don't include it directly, so I missed them initially. I think that since so many plugins require Eigen3 I will just make it required no matter what.

astro-friedel: Regarding gstreamer, is this a newer requirement as I do not see anything in the master branch that references gstreamer.

CMakeLists.txt
astro-friedel: Can you be more specific?

CMakeLists.txt
shg8: Yes, we previously asked folks to install it manually, but that proved to be a lot of trouble since different distros package different versions. It would be nice if this could be handled by the build system.

CMakeLists.txt
shg8: I believe `CMAKE_CXX_FLAGE_RELEASE` should be `CMAKE_CXX_FLAGS_RELEASE`. @wermos first spotted this in #381

CMakeLists.txt
astro-friedel: Which branch and plugin is it being used by? As I am making my changes off of the main branch and there is no mention of gstreamer anywhere in there.

CMakeLists.txt
astro-friedel: Fixed, thanks.

astro-friedel:
Regarding leaving parts of the `install_apt_deps.sh` script, I think it makes a cleaner and more universal to use the cmake and the web page interface that gives you the command line for the packages to install. I have also updated the documentation

CMakeLists.txt
shg8: It's used by the offload_vio plugins in offload-h265-clean, which we're hoping to merge in the next release. I don't know how much work it involves to compile gstreamer and link to it statically, but, if it's a lot, I believe we can hold off on that and provide instructions for manual installation.

charmoniumQ:
Ah, I see the command in the documentation now; that is quite sufficient.

I take back my objection to removing `install_apt_deps.sh`.

CMakeLists.txt
astro-friedel: Can I ask why gstreamer is statically linked?

astro-friedel: From looking at the code in that branch, the gstreamer libraries are all dynamically linked, not statically.

astro-friedel: Adding in the gstreamer dependency should be very straight forward.

charmoniumQ:
What are the pros/cons of vendoring dependencies like sqlite3pp, concurrentqueue, cxxopts vs pulling them in through CMake?

jianxiapyh:
Hi Doug, where is Sophus used? If we need it, I think we need to add libfmt-dev as well.

CMakeLists.txt
```diff
@@ -0,0 +1,329 @@
...
+
+
+foreach(ITEM IN LISTS EXT_PLUGIN_LIST)
+    string(TOUPPER "USE_${ITEM}" ITEM_UPPER)
+    if(${ITEM_UPPER})
+        include(cmake/Get${ITEM}.cmake)
+    endif()
+endforeach()
+
+file(DOWNLOAD ${DATA_FILE} data.zip SHOW_PROGRESS)
```
jianxiapyh, line 321: I am also running into issue here, if I use relative path, I am encountering error "file DOWNLOAD error: cannot create directory '' - Specify file by full path", I changed to ./data.zip and it works fine. My cmake version is 3.16.3.

astro-friedel:
@charmoniumQ 
> What are the pros/cons of vendoring dependencies like sqlite3pp, concurrentqueue, cxxopts vs pulling them in through CMake?

Using Vendor version:
Pros:
  - these libraries are guaranteed to be stable
  - any dependencies they have will automatically be installed by the vendor package manager
  - this is standard practice in the industry
  - these libraries will be kept up to date and compatible with the rest of the system
  - any plugins that depend on these libraries will be using the same version and avoids conflicts and collisions

Cons: 
  - if there is a specific version you need this will not work, but then one would have to think about why a specific version is necessary

Building through CMake
Pros:
  - the specific version is under your control (but, again one would need to think about why a specific version is needed)
 
Cons:
  - you may need to worry about installing any dependencies of these libraries manually
  - system compatibility is not guaranteed in the long term
  - you will need to manually update the cmake files when a new version is needed
  - Build times will be longer
  - there could be conflicts with vendor installed versions at link/run time
  - not all libraries are easy or friendly to build from source

All in all, I am a big fan of vendor supplied libraries as they make management much easier and stable long term. There are times when building one of these from source is necessary (e.g. OpenCV on CentOS does not provide the viz module, so we build it from source in that instance), but these should be kept to a minimum.

CMakeLists.txt
astro-friedel: Changing that should be ok. In the long term we should have a standard demo data file that is part of the github repo, rather than relying on downloading from somewhere.

charmoniumQ:
>> What are the pros/cons of vendoring dependencies like sqlite3pp, concurrentqueue, cxxopts vs pulling them in through CMake?
> All in all, I am a big fan of vendor supplied libraries

I meant vendoring as copying the source code of third-party library into our source-tree ([ref](https://stackoverflow.com/a/39643873/1078199)). 

If this is your preference, [Concurrentqueue](https://launchpad.net/ubuntu/jammy/+source/concurrentqueue) and [cxxopts](https://launchpad.net/ubuntu/jammy/+source/cxxopts) do have Ubunut packages we could switch to (I don't know about Fedora or CentOS).

Sqlite3pp has no Ubuntu package, so it can either remain vendored in our repository or we can pull down a copy from the internet.

wsherman64:
A note on the OpenCV need of the "vis module".  If this is what I'm
remembering, the requirement of the OpenCV "vis" module is actually a
problem for users of VTK who want a modern version of VTK.

My recollection is that the OpenCV vis module wants version 6.x of VTK,
but VTK is now on the 9.x version, so OpenCV is massively out of date
in that regard.  It'd be better if we could avoid requiring the ancient VTK.
I'm not even sure there's a great use of OpenCV's vis module in ILLIXR.

	Bill

astro-friedel:
@charmoniumQ 
> > > What are the pros/cons of vendoring dependencies like sqlite3pp, concurrentqueue, cxxopts vs pulling them in through CMake?
> > > All in all, I am a big fan of vendor supplied libraries
> 
> I meant vendoring as copying the source code of third-party library into our source-tree ([ref](https://stackoverflow.com/a/39643873/1078199)).
> 
> If this is your preference, [Concurrentqueue](https://launchpad.net/ubuntu/jammy/+source/concurrentqueue) and [cxxopts](https://launchpad.net/ubuntu/jammy/+source/cxxopts) do have Ubunut packages we could switch to (I don't know about Fedora or CentOS).
> 
> Sqlite3pp has no Ubuntu package, so it can either remain vendored in our repository or we can pull down a copy from the internet.

All of these are very small projects (some even just header only) so I could go either way (vendoring or using the OS repo) for these. For Sqlite3pp vendoring would be the only option. Ideally we would make these submodules, rather than having the code directly in our repo.

qinjunj (commented):
Hi Doug, the file `build/_deps/kimera_vio/tmp/Kimera_VIO-gitclone.cmake` has empty paths for some git commands. I'm sending the file to your email.

jianxiapyh:
1. I am also running into the empty paths issue qinjunj mentioned above. 
2. It seems the current build assumes that the user has OpenCV prebuilt in their system. However, since ILLIXR still runs a specific version of OpenCV (3.4.6) should we install OpenCV in the build script?

astro-friedel:
@qinjunj 
> Hi Doug, the file `build/_deps/kimera_vio/tmp/Kimera_VIO-gitclone.cmake` has empty paths for some git commands. I'm sending the file to your email.

This is now fixed

astro-friedel:
@jianxiapyh 
> 1. I am also running into the empty paths issue qinjunj mentioned above.

This has now been fixed.

> 2. It seems the current build assumes that the user has OpenCV prebuilt in their system. However, since ILLIXR still runs a specific version of OpenCV (3.4.6) should we install OpenCV in the build script?

Yes, this build system assumes that OpenCV is on the system as it is available through the OS repos. The current code should build with OpenCV 3.x or 4.x. I do not want to attach us to a specific version, especially not 3.6 as it is very old and not supported by several of the other 3rd party codes that we use.

plugins/offload_vio/CMakeLists.txt
```diff
@@ -0,0 +1,6 @@
+find_package(eCAL)
```
qinjunj, line 1: Missing `ecal`, `zed` and `glm` package.

astro-friedel: Can you clarify? The zed and eCAL dependencies are given on the build instructions. The glam is missing and I have added that one.

src/main.cpp
```diff
@@ -0,0 +1,75 @@
+#include "illixr.hpp"
+
+#include <csignal>
+
```
hungdche, line 4: I think you have to include `iostream` to use `cout`. At least that's the issue in my local machine

profiles/native_vulkan.yaml
Jebbly, line 1: Same here, `native_vulkan.yaml` -> `native_vk.yaml`

cmake/GetMonado_gl.cmake
```diff
@@ -0,0 +1,46 @@
+# module to download, build and install the Monado_gl ILLIXR plugin
+
+# get dependencies
+pkg_check_modules(glslang glslang)
```
Jebbly, line 4: Not sure if my understanding of my issue is complete, but my CMake build is failing here. It may have to do with my distro - I'm not very familiar with pkg-config, but this issue seems to suggest that certain distros applied their own patches to add support for this: https://github.com/KhronosGroup/glslang/issues/1715.

Running `pkg-config --list-all | grep glslang` also doesn't seem to return anything for me, even though I have `glslang-tools` installed (and verified that I can run `glslang`). As an alternative to this, it seems like [find_package(Vulkan)](https://cmake.org/cmake/help/latest/module/FindVulkan.html) offers some information for utilities like glslang in newer CMake versions, so we could consider requiring a higher version of CMake?

Jebbly: To follow up on this, the actual error that I encounter afterwards is this:
```
CMake Error at cmake/GetMonado_gl.cmake:6 (find_package):
  By not providing "Findglslang.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "glslang", but
  CMake did not find one.

  Could not find a package configuration file provided by "glslang"
  (requested version 11) with any of the following names:

    glslangConfig.cmake
    glslang-config.cmake

  Add the installation prefix of "glslang" to CMAKE_PREFIX_PATH or set
  "glslang_DIR" to a directory containing one of the above files.  If
  "glslang" provides a separate development package or SDK, be sure it has
  been installed.
Call Stack (most recent call first):
  cmake/HelperFunctions.cmake:262 (include)
  plugins/timewarp_gl/CMakeLists.txt:2 (get_external_for_plugin)
```

jianxiapyh:
Is the native group still enabled? When I use it, it shows every plugin is disabled.

cmake/GetMonado_gl.cmake
jianxiapyh: I encountered this error, but after I installed glslang-dev, this error is gone. But then I am running into additional missing libraries: 1. xcb-glx0-dev 2. libx11-xcb-dev 3. libxcb-randr0-dev 4. xkbcommon-dev. Finally, I am reaching the error:
```
CMake Error at cmake/GetMonado_gl.cmake:20 (find_package):
  By not providing "FindOpenXR.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "OpenXR", but
  CMake did not find one.

  Could not find a package configuration file provided by "OpenXR" with any
  of the following names:

    OpenXRConfig.cmake
    openxr-config.cmake

  Add the installation prefix of "OpenXR" to CMAKE_PREFIX_PATH or set
  "OpenXR_DIR" to a directory containing one of the above files.  If "OpenXR"
  provides a separate development package or SDK, be sure it has been
  installed.
Call Stack (most recent call first):
  cmake/HelperFunctions.cmake:262 (include)
  plugins/timewarp_gl/CMakeLists.txt:2 (get_external_for_plugin)
```

include/illixr/vk_util/render_pass.hpp
```diff
@@ -40,7 +42,7 @@ class timewarp : public render_pass {
      * @param input_texture_vulkan_coordinates Whether the input texture is in Vulkan coordinates.
      */
     virtual void setup(VkRenderPass render_pass, uint32_t subpass, std::array<std::vector<VkImageView>, 2> buffer_pool,
-                       bool input_texture_vulkan_coordinates = true) = 0;
```
shg8, line 43: Why is this removed? This would flip the texture.

common/common.mk
```diff
@@ -1,89 +0,0 @@
...
-CXX := clang++-10
-STDCXX ?= c++17
-CFLAGS := $(CFLAGS) -DGLSL_VERSION='"330 core"'
-RM := rm -f
-
-## A compilation flag for selecting the Monado integration mode
-#> If ILLIXR_MONADO_VULKAN is set to 'ON', disable OpenGL initializatin
-#> Else ILLIXR_MONADO_VULKAN is empty, enabling the monado_integration compatible compilation
-ifeq ($(ILLIXR_MONADO_VULKAN),ON)
-	MONADO_FLAGS := -DILLIXR_MONADO_MAINLINE
```
shg8, line 12: I wasn't able to find where this is defined in the Cmake files. This needs to be set for both the gl and vk flavors of Monado.

shg8 (changes requested):

common/common.mk
Jebbly: This may be an adjustment to the OpenGL rebase, which defines `ILLIXR_MONADO` and `ILLIXR_VULKAN` as separate defines.

cmake/GetMonado_gl.cmake
Jebbly: Unfortunately, neither glslang-dev nor glslang-tools are working for me. I'm inclined to add a FindGlslang.cmake file because the Vulkan config will also depend on it.

Jebbly:
> Is the native group still enabled? When I use it, it shows every plugin is disabled. Also, the current ILLIXR Installation Instructions with Group: None is incorrect because it requires the value to be set to on to use it. E.g., I can't get CMake to run with -DUSE_GLDEMO (provided by website command) unless I set the value (-DUSE_GLDEMO=on)

It seems like the `BUILD_GROUP` define has been replaced with `YAML_FILE`,so you would specify a specific yaml profile instead, e.g. you can use `-DYAML_FILE=profiles/native_gl.yaml`

src/main.cpp
astro-friedel: Thanks for catching this.

docs/getting_started.md
```diff
@@ -1,117 +1,472 @@
...
 
-    Update the submodules.
-    Submodules are git repositories inside a git repository that need to be pulled down separately:
+    <button class="code-box-copy__btn" data-clipboard-target="#clone" title="Copy"></button>
+    <pre class="language-shell" id="clone">git clone https://github.com/ILLIXR/ILLIXR</pre>
 
-    <!--- language: lang-none -->
+</div>
+<br>
+### Install dependencies
```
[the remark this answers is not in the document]
astro-friedel: Can you show me a screen shot? What browser are you using?

profiles/native_vulkan.yaml
astro-friedel: done

cmake/GetMonado_gl.cmake
astro-friedel: @Jebbly What Operating systems and versions are you using?

astro-friedel: @jianxiapyh All of those libraries are listed as dependencies in the documentation as is libopenxr-dev which should provide OpenXRConfig.cmake

astro-friedel:
@jianxiapyh The groups have been converted to profile.yaml files. To use the native profile set -DYAML_FILE=profiles/native_gl.yaml (note that all profile files are generated from plugins/plugins.yaml). I fixed the documentation to include the =ON part of the command.

include/illixr/vk_util/render_pass.hpp
astro-friedel: It was removed because default arguments on virtual functions are strongly discouraged due to potential issues with overloaded versions of the functions. All this change should do is force all of the calls that relied on the default value to add true to their argument list.

common/common.mk
astro-friedel: Once the OpenGL rebase is merged into main its changes will be reflected in this branch.

cmake/GetMonado_gl.cmake
jianxiapyh: @astro-friedel , I don't think libopenxr-dev is in 20.04

jianxiapyh: @astro-friedel 
I found a solution to this OpenXRConfig. According to https://monado.freedesktop.org/packages-ubuntu.html
For 20.04, we need to add
```
sudo add-apt-repository ppa:monado-xr/monado
sudo apt-get update|
```
in order to install libopenxr-dev

astro-friedel: Yes, adding that repo has been part of the instructions for 20.04 for a while.

common/common.mk
astro-friedel: The addition of the -DILLIXR_MONADO_MAINLINE flag is now part of the build system.

jianxiapyh:
I am getting the following error
```
terminate called after throwing an instance of 'YAML::TypedBadConversion<bool>'
  what():  yaml-cpp: error at line 7, column 1: bad conversion
```

qinjunj:
I'm having similar issues as Yihan. The yaml file I use is as follows ```# This file was auto generated, take caution if manually editing.
duration: 60
data: ../V1_02_medium/mav0/
demo_data: ../demo_data/
plugins: offline_imu,offline_cam,slam2,gtsam_integrator,pose_prediction,ground_truth_slam,gldemo,debugview,offload_data,timewarp_gl,audio_pipeline```

cmake/GetMonado_gl.cmake
Jebbly: > @Jebbly What Operating systems and versions are you using?

I'm on Ubuntu 20.04, and still running into this issue.

hungdche:
I'm not sure if the issue I have is similar, but here it goes. I run (from `ILLIXR/build`)
```
./main.opt.exe --duration=60 --data=../data/mav0/ --demo_data=../demo_data/ --enable_offload=false --enable_alignment=false --enable_verbose_errors=true --enable_pre_sleep=false --yaml=../illixr.yaml
```
and the error is:
```
Reading ../illixr.yaml
terminate called after throwing an instance of 'std::logic_error'
  what():  basic_string::_M_construct null not valid
Aborted (core dumped)

```

cmake/GetMonado_gl.cmake
Jebbly: Okay, I think I've finally resolved the issue. The issue has to do with installing the Vulkan SDK according to this documentation: https://vulkan.lunarg.com/doc/view/latest/linux/getting_started_ubuntu.html

When that additional repository is added, installing glslang causes an error for some reason (likely due to versioning, but not entirely sure). It would be good to document this behavior somewhere because it seems likely for someone to encounter this in the future.

include/illixr/vk_util/render_pass.hpp
shg8: Would you also be able to make the corresponding changes in monado_vulkan_integration since this function is used there? Otherwise, we could remove the virtual qualifier and add back the default.

shg8: Also, `native_renderer` uses this function as well. So let's just remove the virtual and add back the default value.

shg8:
There seems to be a problem when building on Fedora 38.

```
[ 92%] Performing configure step for 'OpenVINS'
cd /home/steven/Documents/ILLIXR/ILLIXR/build/_deps/OpenVINS/src/OpenVINS-build && /usr/bin/cmake -DCMAKE_INSTALL_PREFIX=/home/steven/Documents/ILLIXR/ILLIXR/install -DCMAKE_CXX_FLAGS=-L/home/steven/Documents/ILLIXR/ILLIXR/install/lib -DCMAKE_BUILD_TYPE=Release -DILLIXR_ROOT=/home/steven/Documents/ILLIXR/ILLIXR/include -DILLIXR_BUILD_SUFFIX=.opt -DILLIXR_INTEGRATION=ON "-GUnix Makefiles" -S /home/steven/Documents/ILLIXR/ILLIXR/build/_deps/OpenVINS/src/OpenVINS -B /home/steven/Documents/ILLIXR/ILLIXR/build/_deps/OpenVINS/src/OpenVINS-build
-- EIGEN VERSION: 3.4.0
-- OPENCV VERSION: 4.7.0
-- BOOST VERSION: 1.78.0
CMake Warning at ov_core/CMakeLists.txt:33 (message):
  CATKIN NOT FOUND BUILDING WITHOUT ROS!


-- EIGEN VERSION: 3.4.0
-- OPENCV VERSION: 4.7.0
-- BOOST VERSION: 1.78.0
CMake Warning at ov_msckf/CMakeLists.txt:36 (message):
  CATKIN NOT FOUND BUILDING WITHOUT ROS!


CMake Warning at ov_msckf/CMakeLists.txt:79 (message):
  MANUALLY LINKING TO OV_CORE LIBRARY....


-- Configuring done (0.0s)
-- Generating done (0.0s)
-- Build files have been written to: /home/steven/Documents/ILLIXR/ILLIXR/build/_deps/OpenVINS/src/OpenVINS-build
cd /home/steven/Documents/ILLIXR/ILLIXR/build/_deps/OpenVINS/src/OpenVINS-build && /usr/bin/cmake -E touch /home/steven/Documents/ILLIXR/ILLIXR/build/_deps/OpenVINS/src/OpenVINS-stamp/OpenVINS-configure
[ 94%] Performing build step for 'OpenVINS'
cd /home/steven/Documents/ILLIXR/ILLIXR/build/_deps/OpenVINS/src/OpenVINS-build && /usr/bin/gmake
gmake[3]: Entering directory '/home/steven/Documents/ILLIXR/ILLIXR/build/_deps/OpenVINS/src/OpenVINS-build'
[  4%] Building CXX object ov_core/CMakeFiles/ov_core_lib.dir/src/track/TrackAruco.cpp.o
In file included from /home/steven/Documents/ILLIXR/ILLIXR/build/_deps/OpenVINS/src/OpenVINS/ov_core/src/track/TrackAruco.cpp:21:
/home/steven/Documents/ILLIXR/ILLIXR/build/_deps/OpenVINS/src/OpenVINS/ov_core/src/track/TrackAruco.h: In constructor ‘ov_core::TrackAruco::TrackAruco()’:
/home/steven/Documents/ILLIXR/ILLIXR/build/_deps/OpenVINS/src/OpenVINS/ov_core/src/track/TrackAruco.h:46:84: error: no match for ‘operator=’ (operand types are ‘cv::Ptr<cv::aruco::Dictionary>’ and ‘cv::aruco::Dictionary’)
   46 |             aruco_dict = cv::aruco::getPredefinedDictionary(cv::aruco::DICT_6X6_250);
      |                                                                                    ^
In file included from /usr/include/opencv4/opencv2/core/cvstd.hpp:81,
                 from /usr/include/opencv4/opencv2/core/base.hpp:58,
                 from /usr/include/opencv4/opencv2/core.hpp:53,
                 from /usr/include/opencv4/opencv2/objdetect/aruco_dictionary.hpp:7,
                 from /usr/include/opencv4/opencv2/objdetect/aruco_detector.hpp:7,
                 from /usr/include/opencv4/opencv2/aruco.hpp:7,
                 from /home/steven/Documents/ILLIXR/ILLIXR/build/_deps/OpenVINS/src/OpenVINS/ov_core/src/track/TrackAruco.h:25:
/usr/include/opencv4/opencv2/core/cvstd_wrapper.hpp:122:38: note: candidate: ‘template<class Y> cv::Ptr<T>& cv::Ptr<T>::operator=(const cv::Ptr<Y>&) [with T = cv::aruco::Dictionary]’
  122 |     template<typename Y> inline Ptr& operator=(const Ptr<Y>& o) { std::shared_ptr<T>::operator =(o); return *this; }
      |                                      ^~~~~~~~
/usr/include/opencv4/opencv2/core/cvstd_wrapper.hpp:122:38: note:   template argument deduction/substitution failed:
/home/steven/Documents/ILLIXR/ILLIXR/build/_deps/OpenVINS/src/OpenVINS/ov_core/src/track/TrackAruco.h:46:84: note:   ‘cv::aruco::Dictionary’ is not derived from ‘const cv::Ptr<_Tp>’
   46 |             aruco_dict = cv::aruco::getPredefinedDictionary(cv::aruco::DICT_6X6_250);
      |                                                                                    ^
/usr/include/opencv4/opencv2/core/cvstd_wrapper.hpp:121:10: note: candidate: ‘cv::Ptr<T>& cv::Ptr<T>::operator=(const cv::Ptr<T>&) [with T = cv::aruco::Dictionary]’
  121 |     Ptr& operator=(const Ptr& o) { std::shared_ptr<T>::operator =(o); return *this; }
      |          ^~~~~~~~
/usr/include/opencv4/opencv2/core/cvstd_wrapper.hpp:121:31: note:   no known conversion for argument 1 from ‘cv::aruco::Dictionary’ to ‘const cv::Ptr<cv::aruco::Dictionary>&’
  121 |     Ptr& operator=(const Ptr& o) { std::shared_ptr<T>::operator =(o); return *this; }
      |                    ~~~~~~~~~~~^
/home/steven/Documents/ILLIXR/ILLIXR/build/_deps/OpenVINS/src/OpenVINS/ov_core/src/track/TrackAruco.h:47:59: error: ‘create’ is not a member of ‘cv::aruco::DetectorParameters’
   47 |             aruco_params = cv::aruco::DetectorParameters::create();
      |                                                           ^~~~~~
/home/steven/Documents/ILLIXR/ILLIXR/build/_deps/OpenVINS/src/OpenVINS/ov_core/src/track/TrackAruco.h: In constructor ‘ov_core::TrackAruco::TrackAruco(int, bool)’:
/home/steven/Documents/ILLIXR/ILLIXR/build/_deps/OpenVINS/src/OpenVINS/ov_core/src/track/TrackAruco.h:57:84: error: no match for ‘operator=’ (operand types are ‘cv::Ptr<cv::aruco::Dictionary>’ and ‘cv::aruco::Dictionary’)
   57 |             aruco_dict = cv::aruco::getPredefinedDictionary(cv::aruco::DICT_6X6_250);
      |                                                                                    ^
/usr/include/opencv4/opencv2/core/cvstd_wrapper.hpp:122:38: note: candidate: ‘template<class Y> cv::Ptr<T>& cv::Ptr<T>::operator=(const cv::Ptr<Y>&) [with T = cv::aruco::Dictionary]’
  122 |     template<typename Y> inline Ptr& operator=(const Ptr<Y>& o) { std::shared_ptr<T>::operator =(o); return *this; }
      |                                      ^~~~~~~~
/usr/include/opencv4/opencv2/core/cvstd_wrapper.hpp:122:38: note:   template argument deduction/substitution failed:
/home/steven/Documents/ILLIXR/ILLIXR/build/_deps/OpenVINS/src/OpenVINS/ov_core/src/track/TrackAruco.h:57:84: note:   ‘cv::aruco::Dictionary’ is not derived from ‘const cv::Ptr<_Tp>’
   57 |             aruco_dict = cv::aruco::getPredefinedDictionary(cv::aruco::DICT_6X6_250);
      |                                                                                    ^
/usr/include/opencv4/opencv2/core/cvstd_wrapper.hpp:121:10: note: candidate: ‘cv::Ptr<T>& cv::Ptr<T>::operator=(const cv::Ptr<T>&) [with T = cv::aruco::Dictionary]’
  121 |     Ptr& operator=(const Ptr& o) { std::shared_ptr<T>::operator =(o); return *this; }
      |          ^~~~~~~~
/usr/include/opencv4/opencv2/core/cvstd_wrapper.hpp:121:31: note:   no known conversion for argument 1 from ‘cv::aruco::Dictionary’ to ‘const cv::Ptr<cv::aruco::Dictionary>&’
  121 |     Ptr& operator=(const Ptr& o) { std::shared_ptr<T>::operator =(o); return *this; }
      |                    ~~~~~~~~~~~^
/home/steven/Documents/ILLIXR/ILLIXR/build/_deps/OpenVINS/src/OpenVINS/ov_core/src/track/TrackAruco.h:58:59: error: ‘create’ is not a member of ‘cv::aruco::DetectorParameters’
   58 |             aruco_params = cv::aruco::DetectorParameters::create();
      |                                                           ^~~~~~
gmake[5]: *** [ov_core/CMakeFiles/ov_core_lib.dir/build.make:132: ov_core/CMakeFiles/ov_core_lib.dir/src/track/TrackAruco.cpp.o] Error 1
gmake[4]: *** [CMakeFiles/Makefile2:118: ov_core/CMakeFiles/ov_core_lib.dir/all] Error 2
gmake[3]: *** [Makefile:136: all] Error 2
gmake[3]: Leaving directory '/home/steven/Documents/ILLIXR/ILLIXR/build/_deps/OpenVINS/src/OpenVINS-build'
gmake[2]: *** [plugins/openvins/CMakeFiles/OpenVINS.dir/build.make:89: _deps/OpenVINS/src/OpenVINS-stamp/OpenVINS-build] Error 2
gmake[2]: Leaving directory '/home/steven/Documents/ILLIXR/ILLIXR/build'
gmake[1]: *** [CMakeFiles/Makefile2:734: plugins/openvins/CMakeFiles/OpenVINS.dir/all] Error 2
gmake[1]: Leaving directory '/home/steven/Documents/ILLIXR/ILLIXR/build'
gmake: *** [Makefile:139: all] Error 2

```

rsingh76:
1. Native-GL works (but not when "group" is used for cmake config. It also doesn't work with the yaml file)
2. Native-VK - - compiles but throws a runtime error (std::out_of_range)
3. Monado-GL - compiles, but the executable is not generated
4. Monado-VK - compiles, but hits segfault after com_main_create_system_compositor init
Which one of these works on your machine?

shg8:
Also, the docs seems to still suggest native_vulkan instead native_vk in the command generation.

docs/getting_started.md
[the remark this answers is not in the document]
astro-friedel: Added a new line

astro-friedel:
@qinjunj I have fixed the documentation rendering issue, and the reading of the yaml file

astro-friedel:
@hungdche @jianxiapyh I have made changes to the way the yaml file is read. I hope this fixes the issues you are seeing

astro-friedel:
While running on Fedora is nice it is not my main concern right now (I 
added to the documentation that we only officially support Ubuntu at 
this time)

____________________________________________________________________________
Douglas Friedel, Ph.D.
Senior Research Software Engineer
University of Illinois
National Center for Supercomputing Applications
Department of Astronomy
217-333-9378

astro-friedel:
Can you give some more specifics? (e.g. the command line(s) you used, 
etc.) The 'group' idea has be superseded by profiles which are all 
defined in a single file plugins/plugins.yaml This way only a single 
file needs to be modified to add a new profile and or plugin. The yaml 
files in the profiles directory are now auto generated by cmake from 
plugins/plugins.yaml

____________________________________________________________________________
Douglas Friedel, Ph.D.
Senior Research Software Engineer
University of Illinois
National Center for Supercomputing Applications
Department of Astronomy
217-333-9378

profiles/native_vk.yaml
```diff
@@ -0,0 +1,9 @@
+# This file was auto generated, take caution if manually editing.
+plugins: audio_pipeline,offline_cam,offline_imu,gtsam_integrator,pose_prediction,display_vk,timewarp_vk,vkdemo,native_renderer,debugview
```
shg8, line 2: There are dependency issues with this loading order.
`pose_prediction,display_vk,timewarp_vk,vkdemo,audio_pipeline,debugview,gtsam_integrator,offline_cam,offline_imu,native_renderer` worked for me

shg8: Additionally, the automatically generated illixr.yaml seems to point to the wrong path for `demo_data`. Either the path is wrong or the folder is not copied.

plugins/vkdemo/plugin.cpp
```diff
@@ -0,0 +1,983 @@
...
+    }
+
+    void create_pipeline(VkRenderPass render_pass, uint32_t subpass) {
+        if (pipeline != VK_NULL_HANDLE) {
+            throw std::runtime_error("timewarp_vk::create_pipeline: pipeline already created");
+        }
+
+        auto           folder = std::string(SHADER_FOLDER);
+        VkShaderModule vert =
+            vulkan_utils::create_shader_module(ds->vk_device, vulkan_utils::read_file(folder + "/demo.vert.spv"));
```
shg8, line 750: `SHADER_FOLDER` no longer points to a valid directory. The system crashes on launch.

shg8 (changes requested):
1. For `native_vk`, it seems that shaders weren't compiled / weren't copied to the correct path / path isn't updated in code.
2. `monado_vk` crashes on launch as well. I got
```
terminate called after throwing an instance of 'std::runtime_error'
  what():  Invalid conversion from std::string to bool
Aborted (core dumped)
```

jianxiapyh:
@astro-friedel I ran the following builds and here are the errors I am facing
1. native-gl
cmake command:
```
 cmake .. -DCMAKE_INSTALL_PREFIX=some location -DYAML_FILE=profiles/native_gl.yaml
```
and executes with the following
```
./main.opt.exe --yaml=/home/yihan/Research/ILLIXR/illixr.yaml --vis=openvins
```
Error encountered:
```
Reading /home/yihan/Research/ILLIXR/illixr.yaml
Getting second/third order dodecahedron
Getting second/third order dodecahedron
terminate called after throwing an instance of 'std::out_of_range'
  what():  _Map_base::at
Aborted (core dumped)
```
gdb backtrace
```
__GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
50      ../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
(gdb) bt
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x00007ffff7b7b859 in __GI_abort () at abort.c:79
#2  0x00007ffff7e048d1 in ?? () from /lib/x86_64-linux-gnu/libstdc++.so.6
#3  0x00007ffff7e1037c in ?? () from /lib/x86_64-linux-gnu/libstdc++.so.6
#4  0x00007ffff7e103e7 in std::terminate() () from /lib/x86_64-linux-gnu/libstdc++.so.6
#5  0x00007ffff7e10699 in __cxa_throw () from /lib/x86_64-linux-gnu/libstdc++.so.6
#6  0x00007ffff7e0733c in std::__throw_out_of_range(char const*) () from /lib/x86_64-linux-gnu/libstdc++.so.6
#7  0x00007ffff473e1fc in std::__detail::_Map_base<std::type_index, std::pair<std::type_index const, std::shared_ptr<ILLIXR::phonebook::service> const>, std::allocator<std::pair<std::type_index const, std::shared_ptr<ILLIXR::phonebook::service> const> >, std::__detail::_Select1st, std::equal_to<std::type_index>, std::hash<std::type_index>, std::__detail::_Mod_range_hashing, std::__detail::_Default_ranged_hash, std::__detail::_Prime_rehash_policy, std::__detail::_Hashtable_traits<false, false, true>, true>::at(std::type_index const&) const ()
   from /home/yihan/Research/cmake_build/lib/libplugin.debugview.opt.so
#8  0x00007ffff47469a7 in debugview::debugview(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, ILLIXR::phonebook*) () from /home/yihan/Research/cmake_build/lib/libplugin.debugview.opt.so
#9  0x00007ffff4725746 in this_plugin_factory () from /home/yihan/Research/cmake_build/lib/libplugin.debugview.opt.so
#10 0x00007ffff7fb2673 in std::back_insert_iterator<std::vector<std::unique_ptr<ILLIXR::plugin, std::default_delete<ILLIXR::plugin> >, std::allocator<std::unique_ptr<ILLIXR::plugin, std::default_delete<ILLIXR::plugin> > > > > std::transform<__gnu_cxx::__normal_iterator<ILLIXR::plugin* (* const*)(ILLIXR::phonebook*), std::vector<ILLIXR::plugin* (*)(ILLIXR::phonebook*), std::allocator<ILLIXR::plugin* (*)(ILLIXR::phonebook*)> > >, std::back_insert_iterator<std::vector<std::unique_ptr<ILLIXR::plugin, std::default_delete<ILLIXR::plugin> >, std::allocator<std::unique_ptr<ILLIXR::plugin, std::default_delete<ILLIXR::plugin> > > > >, runtime_impl::load_so(std::vector<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::allocator<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > > const&)::{lambda(auto:1 const&)#3}>(__gnu_cxx::__normal_iterator<ILLIXR::plugin* (* const*)(ILLIXR::phonebook*), std::vector<ILLIXR::plugin* (*)(ILLIXR::phonebook*), std::allocator<ILLIXR::plugin* (*)(ILLIXR::phonebook*)> > >, __gnu_cxx::__normal_iterator<ILLIXR::plugin* (* const*)(ILLIXR::phonebook*), std::vector<ILLIXR::plugin* (*)(ILLIXR::phonebook*), std::allocator<ILLIXR::plugin* (*)(ILLIXR::phonebook*)> > >, runtime_impl::load_so(std::vector<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::allocator<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > > const&)::{lambda(auto:1 const&)#3}, runtime_impl::load_so(std::vector<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::allocator<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > > const&)::{lambda(auto:1 const&)#3}) ()
   from /home/yihan/Research/cmake_build/lib/libplugin.main.opt.so
#11 0x00007ffff7fb2b09 in runtime_impl::load_so(std::vector<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::allocator<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > > const&) ()
   from /home/yihan/Research/cmake_build/lib/libplugin.main.opt.so
#12 0x00007ffff7f8e7ee in ILLIXR::run(cxxopts::ParseResult const&) () from /home/yihan/Research/cmake_build/lib/libplugin.main.opt.so
#13 0x0000555555564555 in main ()

```

jianxiapyh:
@astro-friedel in native_vulkan, 
cmake command 
```
cmake .. -DCMAKE_INSTALL_PREFIX=/home/yihan/Research/cmake_build -DYAML_FILE=profiles/native_vulkan.yaml
```
build error:
```
make[5]: Entering directory '/home/yihan/Research/ILLIXR/build/_deps/audio_pipeline/src/Audio_Pipeline-build'
[ 14%] Building CXX object CMakeFiles/plugin.audio_pipeline.opt.dir/src/audio.cpp.o
/usr/bin/c++  -DAUDIO_SAMPLES=\"/home/yihan/Research/cmake_build/etc/audio_pipeline\" -DILLIXR_INTEGRATION -Dplugin_audio_pipeline_opt_EXPORTS -I/home/yihan/Research/cmake_build/include -I/home/yihan/Research/ILLIXR/build/_deps/audio_pipeline/src/Audio_Pipeline/include -I/home/yihan/Research/ILLIXR/include  -L/home/yihan/Research/cmake_build/lib -L/home/yihan/Research/cmake_build/lib64 -O3 -DNDEBUG -fPIC   -std=gnu++17 -o CMakeFiles/plugin.audio_pipeline.opt.dir/src/audio.cpp.o -c /home/yihan/Research/ILLIXR/build/_deps/audio_pipeline/src/Audio_Pipeline/src/audio.cpp
In file included from /home/yihan/Research/ILLIXR/build/_deps/audio_pipeline/src/Audio_Pipeline/include/audio.h:3,
                 from /home/yihan/Research/ILLIXR/build/_deps/audio_pipeline/src/Audio_Pipeline/src/audio.cpp:4:
/home/yihan/Research/ILLIXR/build/_deps/audio_pipeline/src/Audio_Pipeline/include/sound.h:6:10: fatal error: spatialaudio/Ambisonics.h: No such file or directory
    6 | #include <spatialaudio/Ambisonics.h>
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
```
**This error will not trigger if I build and install native_gl first** 
In addition, I am not seeing monado-service-gl, in my install directory

jianxiapyh:
@astro-friedel when I build offload_device.yaml and offload_server.yaml
1. offload_device.yaml
cmake command 
```
cmake .. -DCMAKE_INSTALL_PREFIX=/home/yihan/Research/cmake_build -DYAML_FILE=profiles/offload_device.yaml
```
execution command (I didn't include --vins part since device side do not need openVINs)
```
./main.opt.exe --yaml=/home/yihan/Research/ILLIXR/illixr.yaml
```
gdb backtrace
```
terminate called after throwing an instance of 'std::runtime_error'
  what():  Invalid conversion from std::string to bool

Program received signal SIGABRT, Aborted.
__GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
50      ../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
(gdb) gdb
Undefined command: "gdb".  Try "help".
(gdb) bt
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x00007ffff7b7b859 in __GI_abort () at abort.c:79
#2  0x00007ffff7e048d1 in ?? () from /lib/x86_64-linux-gnu/libstdc++.so.6
#3  0x00007ffff7e1037c in ?? () from /lib/x86_64-linux-gnu/libstdc++.so.6
#4  0x00007ffff7e103e7 in std::terminate() () from /lib/x86_64-linux-gnu/libstdc++.so.6
#5  0x00007ffff7e10699 in __cxa_throw () from /lib/x86_64-linux-gnu/libstdc++.so.6
#6  0x00007ffff7f8c5cb in _GLOBAL__sub_I_plugin.cpp.cold () from /home/yihan/Research/cmake_build/lib/libplugin.main.opt.so
#7  0x00007ffff7fe0b9a in call_init (l=<optimized out>, argc=argc@entry=2, argv=argv@entry=0x7fffffffdb68, env=env@entry=0x7fffffffdb80)
    at dl-init.c:72
#8  0x00007ffff7fe0ca1 in call_init (env=0x7fffffffdb80, argv=0x7fffffffdb68, argc=2, l=<optimized out>) at dl-init.c:30
#9  _dl_init (main_map=0x7ffff7ffe190, argc=2, argv=0x7fffffffdb68, env=0x7fffffffdb80) at dl-init.c:119
#10 0x00007ffff7fd013a in _dl_start_user () from /lib64/ld-linux-x86-64.so.2
#11 0x0000000000000002 in ?? ()
#12 0x00007fffffffdf0d in ?? ()
#13 0x00007fffffffdf3f in ?? ()
#14 0x0000000000000000 in ?? ()
```

jianxiapyh:
@astro-friedel 
for monado_vk.yaml
1. the plugin selector still shows old yaml file (monado_vulkan.yaml)
2. I thought we are using cmake version 3.16 since that is the default version. However, the timewarp_vk asks for a minimum of 3.17
3. permission error
build error:
```
make[2]: Entering directory '/home/yihan/Research/ILLIXR/build'
[  4%] Generating shaders/tw.frag.spv
cd /home/yihan/Research/ILLIXR/build/plugins/timewarp_vk && /usr/bin/cmake -E make_directory /home/yihan/Research/ILLIXR/build/plugins/timewarp_vk/shaders/
cd /home/yihan/Research/ILLIXR/build/plugins/timewarp_vk && ../../../plugins/timewarp_vk/shaders/tw.frag -V -o /home/yihan/Research/ILLIXR/build/plugins/timewarp_vk/shaders/tw.frag.spv
/bin/sh: 1: ../../../plugins/timewarp_vk/shaders/tw.frag: Permission denied
make[2]: *** [plugins/timewarp_vk/CMakeFiles/Timewarp_VK_Shaders.dir/build.make:66: plugins/timewarp_vk/shaders/tw.frag.spv] Error 126
make[2]: Leaving directory '/home/yihan/Research/ILLIXR/build'
make[1]: *** [CMakeFiles/Makefile2:587: plugins/timewarp_vk/CMakeFiles/Timewarp_VK_Shaders.dir/all] Error 2
make[1]: Leaving directory '/home/yihan/Research/ILLIXR/build'
make: *** [Makefile:133: all] Error 2
```
build command:
```
cmake .. -DCMAKE_INSTALL_PREFIX=/home/yihan/Research/cmake_build -DYAML_FILE=profiles/monado_vk.yaml
```

profiles/native_vk.yaml
astro-friedel: I fixed the plugin order in native_vk.
The demo data are currently not installed but I did fix where the demo_data entry is pointing

plugins/vkdemo/plugin.cpp
astro-friedel: I have fixed this. The shader files are now installed in $CMAKE_INSTALL_PREFIX/share/illixr/shaders and the proper location is given to the vkdemo and timewarp_vk code at compilation time. Just be sure to run make install

astro-friedel:
@shg8 Regarding the monado_vk crash, I removed an erroneous plugin from the list which may have been causing the crash.

astro-friedel:
@jianxiapyh Regarding your first message: seeing something with an 'std::out_of_range' error usually indicates that the plugins are not specified in the correct order. I may not have gotten them all correct as I had to guess in some instances. Regarding several of the others and include issues, I have never encountered these issues, nor the permissions issues. Can you give me any more info?

I will fix the plugin selector for tha yaml file names.

I have fixed the cmake version issue with timewarp_vk

In the future please build in debug mode and run with verbose_errors=true, this will give me far more information that a gdb backtrace.

astro-friedel:
@jianxiapyh Regarding the monado-service-gl, This is not in my codebase yet as the PR it relies on (#404) has not been merged into master yet, so I have not merged it into the cmake branch.

shg8:
I had to deviate from the commands in the docs when building. Specifically, 
1. `cmake --build -- -j4`: I had to replace -- with ./
2. `cmake --install`: I had to append "./"

shg8:
For `native_vk`, Monado_vk and OpenXR_APP seem to be incorrectly marked as install 
Screenshot from 2023-09-14 16-52-13

shg8:
> For `native_vk`, Monado_vk and OpenXR_APP seem to be incorrectly marked as install Screenshot from 2023-09-14 16-52-13

Also, illixr.yaml in the root folder didn't seem to reflect the correct plugin order in`native_vk.yaml` after cmake configure.

shg8:
Vulkan validation layers should be required on all vulkan versions. For ubuntu, the package is `vulkan-validationlayers`

astro-friedel:
@shg8 
I will look into the cmake command format
I will also look into why monado_vk is tagged for native_vk.yaml
The order should be the same as it is just copied I believe, I will look into this as well.
For the vulkan-validationlayers package, is just having it on the system all we need, or do things need to like to it as well?

shg8:
> @shg8 I will look into the cmake command format I will also look into why monado_vk is tagged for native_vk.yaml The order should be the same as it is just copied I believe, I will look into this as well. For the vulkan-validationlayers package, is just having it on the system all we need, or do things need to like to it as well?

Thanks!

For validation, we just need it system wide.

shg8:
> @shg8 I will look into the cmake command format I will also look into why monado_vk is tagged for native_vk.yaml The order should be the same as it is just copied I believe, I will look into this as well. For the vulkan-validationlayers package, is just having it on the system all we need, or do things need to like to it as well?

When running the script, I got
```
Register N6ILLIXR13record_loggerE
Register N6ILLIXR8gen_guidE
Register N6ILLIXR11switchboardE
Register N6ILLIXR9StoplightE
Register N6ILLIXR13RelativeClockE
Creating: slow_pose for N6ILLIXR9pose_typeE
Getting second/third order dodecahedron
Getting second/third order dodecahedron
Creating: imu_integrator_input for N6ILLIXR20imu_integrator_inputE
Creating: imu_raw for N6ILLIXR12imu_raw_typeE
Creating: imu for N6ILLIXR8imu_typeE
Thread 139954044204800 start
Creating: cam for N6ILLIXR8cam_typeE
${ILLIXR_DATA}/cam0/data.csv (/opt/ILLIXR-vk/data/mav0/cam0/data.csv) is not a good path
** ERROR ** 
```

astro-friedel:
@shg8, where is the data file on your system? Also did you run `make install`? before running the script?

astro-friedel:
@shg8 The monado is no longer listed when not requested (e.g. native_gl or nataive_vk) and the order of the plugins in illixr.yaml now matches that of the input yaml file

shg8:
> @shg8, where is the data file on your system? Also did you run `make install`? before running the script?

I believe cmake downloaded the data file in the build folder. I did do install before that.

shg8:
Another issue is that `native_vk` doesn't exit cleanly. Seems to be related to pthread as well.

```
terminate called without an active exception
 
Thread 1 "main.dbg.exe" received signal SIGABRT, Aborted.
__pthread_kill_implementation (no_tid=0, signo=6, threadid=140737344440128) at ./nptl/pthread_kill.c:44
44      ./nptl/pthread_kill.c: No such file or directory.
(gdb) bt
#0  __pthread_kill_implementation (no_tid=0, signo=6, threadid=140737344440128) at ./nptl/pthread_kill.c:44
#1  __pthread_kill_internal (signo=6, threadid=140737344440128) at ./nptl/pthread_kill.c:78
#2  __GI___pthread_kill (threadid=140737344440128, signo=signo@entry=6) at ./nptl/pthread_kill.c:89
#3  0x00007ffff793b476 in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
#4  0x00007ffff79217f3 in __GI_abort () at ./stdlib/abort.c:79
#5  0x00007ffff7be3b9e in ?? () from /lib/x86_64-linux-gnu/libstdc++.so.6
#6  0x00007ffff7bef20c in ?? () from /lib/x86_64-linux-gnu/libstdc++.so.6
#7  0x00007ffff7bef277 in std::terminate() () from /lib/x86_64-linux-gnu/libstdc++.so.6
#8  0x00007ffff7ed10a5 in std::thread::~thread (this=0x555555f329b8, __in_chrg=<optimized out>) at /usr/include/c++/11/bits/std_thread.h:152
#9  0x00007fffebda115a in display_vk_plugin::~display_vk_plugin (this=0x555555f32960, __in_chrg=<optimized out>) at /home/steven/Documents/ILLIXR/plugins/display_vk/plugin.cpp:182
#10 0x00007fffebda1186 in display_vk_plugin::~display_vk_plugin (this=0x555555f32960, __in_chrg=<optimized out>) at /home/steven/Documents/ILLIXR/plugins/display_vk/plugin.cpp:182
#11 0x00007ffff7f16f10 in std::default_delete<ILLIXR::plugin>::operator() (this=0x555555d7e930, __ptr=0x555555f32960) at /usr/include/c++/11/bits/unique_ptr.h:85
#12 0x00007ffff7f142cc in std::unique_ptr<ILLIXR::plugin, std::default_delete<ILLIXR::plugin> >::~unique_ptr (this=0x555555d7e930, __in_chrg=<optimized out>) at /usr/include/c++/11/bits/unique_ptr.h:361
#13 0x00007ffff7f1e733 in std::_Destroy<std::unique_ptr<ILLIXR::plugin, std::default_delete<ILLIXR::plugin> > > (__pointer=0x555555d7e930) at /usr/include/c++/11/bits/stl_construct.h:151
#14 0x00007ffff7f1b19b in std::_Destroy_aux<false>::__destroy<std::unique_ptr<ILLIXR::plugin, std::default_delete<ILLIXR::plugin> >*> (__first=0x555555d7e930, __last=0x555555d7e960) at /usr/include/c++/11/bits/stl_construct.h:163
#15 0x00007ffff7f161aa in std::_Destroy<std::unique_ptr<ILLIXR::plugin, std::default_delete<ILLIXR::plugin> >*> (__first=0x555555d7e910, __last=0x555555d7e960) at /usr/include/c++/11/bits/stl_construct.h:196
#16 0x00007ffff7f13a3d in std::_Destroy<std::unique_ptr<ILLIXR::plugin, std::default_delete<ILLIXR::plugin> >*, std::unique_ptr<ILLIXR::plugin, std::default_delete<ILLIXR::plugin> > > (__first=0x555555d7e910, __last=0x555555d7e960)
    at /usr/include/c++/11/bits/alloc_traits.h:848
#17 0x00007ffff7f0f46f in std::vector<std::unique_ptr<ILLIXR::plugin, std::default_delete<ILLIXR::plugin> >, std::allocator<std::unique_ptr<ILLIXR::plugin, std::default_delete<ILLIXR::plugin> > > >::~vector (this=0x5555556bde50, __in_chrg=<optimized out>)
    at /usr/include/c++/11/bits/stl_vector.h:680
#18 0x00007ffff7f0b602 in runtime_impl::~runtime_impl (this=0x5555556bddc0, __in_chrg=<optimized out>) at /home/steven/Documents/ILLIXR/src/runtime_impl.cpp:122
#19 0x00007ffff7f0b666 in runtime_impl::~runtime_impl (this=0x5555556bddc0, __in_chrg=<optimized out>) at /home/steven/Documents/ILLIXR/src/runtime_impl.cpp:122
#20 0x00007ffff7ece0a1 in ILLIXR::run (options=...) at /home/steven/Documents/ILLIXR/src/plugin.cpp:123
#21 0x0000555555616280 in main (argc=3, argv=0x7fffffffd928) at /home/steven/Documents/ILLIXR/src/main.cpp:75
```

astro-friedel:
@shg8 what options did you give to cmake? Can you locate the data file on your machine?

Regarding the bad exit from native_vk, do you have any idea which plugin caused the issue? There is at least one other profile that does not exit cleanly, so I think a general look at how the plugins shut down would be a good idea, but not necessary for this release.

qinjunj:
Can we add default data path and demo_data path to the yaml files in profiles? For e.g., upzip the downloaded data.zip and set `data_path: data/mav0`, `demo_data: ../demo_data`

cmake/GetMonado_vk.cmake
```diff
@@ -0,0 +1,36 @@
...
+
+        # custom install command to get the name of the plugin correct
+        INSTALL_COMMAND make install && ln -sf ${CMAKE_INSTALL_PREFIX}/lib/libopenxr_monado_vk.so ${CMAKE_INSTALL_PREFIX}/lib/libopenxr_monado_vk${ILLIXR_BUILD_SUFFIX}.so
```
shg8, line 32: We should do `cmake --install ./` here in case that the user's not using make.

cmake/GetMonado_gl.cmake
```diff
@@ -0,0 +1,46 @@
...
+        # custom install command to get the name of the plugin correct
+        INSTALL_COMMAND make install && ln -sf ${CMAKE_INSTALL_PREFIX}/lib/libopenxr_monado_gl.so ${CMAKE_INSTALL_PREFIX}/lib/libopenxr_monado_gl${ILLIXR_BUILD_SUFFIX}.so
```
shg8, line 42: Same here

shg8 (changes requested):

cmake/GetMonado_vk.cmake
astro-friedel: fixed

cmake/GetMonado_gl.cmake
astro-friedel: fixed

shg8 (approved):

astro-friedel:
@qinjunj You can specify _data_ and _demo_data_ entries for each profile. Keep in mind that if the _data_ needs to be downloaded it must be a url, not a local file path.

qinjunj:
> @qinjunj You can specify _data_ and _demo_data_ entries for each profile. Keep in mind that if the _data_ needs to be downloaded it must be a url, not a local file path.

Yeah I can update that myself. Was only wondering if we should set a default value so that the user will have something to run out of box. After all we have demo_data already there and `data.zip` downloaded with cmake. But if you prefer not and have that specified clearly in the doc, I'm okay with that. Also for the `??` part in the doc, you can put `--enable_offload, save rendered frames`, `enable_alignment, apply estimated alignment parameters to poses in pose_lookup`, `enable_pre_sleep, activate sleeping at application start for e.g. attaching gdb`.

docs/docs/plugin_README/offload_vio.md
```diff
@@ -14,15 +14,22 @@ and the device running in a different terminal. **Each terminal must be running
...
-Start ILLIXR using the runner with configs/offload-server.yaml
-for one terminal and configs/offload-server.yaml on the other terminal. The default configuration within the
-config files will run OpenVINS on the server and will feed it with the EuRoC dataset from the device.
+Start ILLIXR using
+``` bash
+main.opt.exe --yaml=profiles/offload-server.yaml --data=<> --demo_data=<>
+```
+for one terminal and 
+``` bash
+main.opt.exe --yaml=profiles/offload-device.yaml --data=<> --demo_data=<>
```
jianxiapyh, line 23: Have we tested this?

docs/docs/virtualization.md
```diff
@@ -0,0 +1,62 @@
...
+
+See the [Install dependencies][13] section for details on installing QEMU.
+
+
+## Setup Ubuntu in the VM
+
+Run `ILLIXR/qemu/qemu.sh` to download [Ubuntu 22.04][12], create a virtual hard drive
+    (`illixr.qcow2`), and launch `qemu`.
+
+Your VM image will be created at `ILLIXR/qemu/illixr.qcow2`.
```
jianxiapyh, line 14: TODO test this

docker/README.md
```diff
@@ -0,0 +1,44 @@
+This directory contains Docker files which can be used to create Docker images for any of the supported OS versions.
+Each container has the default user `illixr`, and a script called `cmk.sh` in the home directory. This script can be used to configure, build, and install ILLIXR in the container.
+Note that this requires [Docker](https://docker.com) to be installed.
+
+To build the containers run any of the following (depending on what OS's you need):
+
+**Ubuntu 20.04**
```
jianxiapyh, line 7: TODO test them later

jianxiapyh (approved):

docker/README.md
jianxiapyh: I will go over the remaining later

docs/docs/plugin_README/offload_vio.md
qinjunj: I tested it on my laptop and it works fine. Please try on your side if you have time.

PR #385, opened 2023-05-18
Make Phonebook and Switchboard compatible with Android Studio Compiler

Madhuparna04:
There are two main changes in this commit:

1. Switchboard: Change all dynamic_pointer_cast to reinterpret_pointer_cast so that switchboard works with Android Studio Compiler. 
2. Phonebook: A struct compiled in different compilation units (plugins) may not have the same type index assigned by the Android compiler (or other compilers) thus use type_index.name as the key to the hashmap _m_registry. 
type_index.name seems to be working in both Desktop and Android Studio.

### DIFF
--- a/common/phonebook.hpp
+++ b/common/phonebook.hpp
@@ -100,8 +100,8 @@ public:
 #ifndef NDEBUG
         std::cerr << "Register " << type_index.name() << std::endl;
 #endif
-        assert(_m_registry.count(type_index) == 0);
-        _m_registry.try_emplace(type_index, impl);
+        assert(_m_registry.count(type_index.name()) == 0);
+        _m_registry.try_emplace(type_index.name(), impl);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


charmoniumQ:
Change 2 seems fine.

Change 1 makes me suspicious. That makes me think we're doing something that is actually undefined behavior that happens to work in our case (type punning). Can types that are structurally equivalent, but exist in different compilation units freely convertible?

Still, I understand we haven't figured out how to make ILLIXR compile Android Studio _without_ Change 1, and the show must go on. I only request that at every `reinterpret_pointer_cast`, add a comment with a link to this GitHub comment. If somebody finds a crash at that line 5 years from now, they might at least have a hint.

```
// This used to be a dynamic_pointer_cast, but we changed it to reinterpret_pointer_cast to make it work in Android Studio
// See https://github.com/ILLIXR/ILLIXR/pull/385
```

charmoniumQ (approved):
Approved, if you add the above comments to every `reinterpret_pointer_cast`.

astro-friedel:
It may be safer to use ifdef statements to only make the changes when 
compiling for Android, rather than change something that is known to work.

____________________________________________________________________________
Douglas Friedel, Ph.D.
Senior Research Software Engineer
University of Illinois
National Center for Supercomputing Applications
Department of Astronomy
217-333-9378

charmoniumQ:
Either ifdef or {comment and retest changed version with the old setup (GCC on Linux)} would be fine with me.

Madhuparna04:
Thank you for the comments. I think using ifdef would be a good choice. But in that case, this change should accompany all the other Android-related changes. Since that will take some time, I am closing this PR for now until all the other Android changes are ready to be merged.

PR #387, opened 2023-05-27
Minor fixes

wermos:
While working on my XROS project, I found a couple of things that I felt could be improved upon, and Yihan suggested turning these suggestions into a PR, so I did.

The main things I changed were:
1. The `RelativeClock` implementation was a bit hard to follow, so I modified the implementation a little bit.
2. I removed some unused headers in `common/data_format.hpp`.
3. When I was trying to compile a simple `main.cpp` which included `common/data_format.hpp` with GCC 9.4.0, I got a "Declaration of `time_point` changes meaning of `time_point`" error on [this line](https://github.com/ILLIXR/ILLIXR/blob/e17e6c5679822a4f021d0a6f85628b5ab698b384/common/relative_clock.hpp#L105). After some research, it turns out that [GCC emits a warning when this happens whereas Clang doesn't](https://stackoverflow.com/a/15538759), so I fixed it.

### DIFF
--- a/common/data_format.hpp
+++ b/common/data_format.hpp
@@ -1,8 +1,5 @@
 #pragma once
 
-#include <boost/optional.hpp>
-#include <chrono>
-#include <iostream>
 #include <memory>
 #include <opencv2/core/mat.hpp>
 #undef Success // For 'Success' conflict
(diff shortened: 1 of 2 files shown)


astro-friedel:
Closing this as it mey not be needed any more. The header cleanup has already been done with the cmake conversion.

PR #392, opened 2023-07-18
Offload rebase

qinjunj:
Updates to the offloading VIO plugins:

- TCP implementation with separated IMU and Camera types
- H264 codec


jianxiapyh:
I am running into this issue running this PR. GStreamer compression is getting stuck here \
`    cv.wait(lock, [this]() { 
     return img_ready;
     });
` \
This issue happened before with an earlier GStreamer version as well. \ 
Running non-compression method does work. However, @qinjunj  need to uncomment these two lines in the PR \
`            /** NO COMPRESSION **/
            //cam_data->set_img0_data((void*) img0.data, img0_size);
            //cam_data->set_img1_data((void*) img1.data, img0_size);
            /** NO COMPRESSION END **/
`

common/data_format.hpp
```diff
@@ -160,14 +185,23 @@ struct rendered_frame : public switchboard::event {
 };
 
 struct hologram_input : public switchboard::event {
-    uint seq;
+    int seq;
```
shg8, line 188: Why int instead of uint?

common/network/socket_instance.hpp
```diff
@@ -0,0 +1,20 @@
...
+#define SOCKETINSTANCE_HPP
+
+#include "address.hpp"
+#include "exception.hpp"
+#include "socket.hpp"
+
+/* class for network sockets (UDP, TCP, etc.) */
+class SocketInstance {
+public:
+    static UDPSocket socket;
```
shg8, line 13: Is the default UDP?

common/network/util.hpp
```diff
@@ -0,0 +1,209 @@
...
+
+    unique_ptr<char, Free> cwd_ptr{get_current_dir_name()};
+    if (not cwd_ptr) {
+        throw unix_error("getcwd");
+    }
+
+    return cwd_ptr.get();
+}
+
+// class TemporarilyUnprivileged {
```
shg8, line 185: Why is this commented? Remove if not used?

configs/openxr_app.yaml
```diff
@@ -2,5 +2,8 @@ app:
   src_path:
     git_repo : https://gitlab.freedesktop.org/monado/demos/openxr-simple-example
     version  : master
-  bin_subpath: build/openxr-example
+  # bin_subpath: Materials_VR.x86_64
+  bin_subpath: Sponza_VR.x86_64
```
shg8, line 6: Should we stick with openxr simple example since we want to run it out of the box?

ground_truth_slam/plugin.cpp
```diff
@@ -6,10 +6,18 @@
 #include "data_loading.hpp"
 
 #include <chrono>
+#include <filesystem>
+#include <fstream>
 #include <iomanip>
 #include <thread>
 
 using namespace ILLIXR;
+#define ViconRoom1Easy      1403715273262142976
```
shg8, line 15: Maybe add a comment explaining what these are?

shg8 (changes requested):
Some minor changes. Otherwise LGTM

common/network/socket_instance.hpp
qinjunj: No, this is not the one actually used in offload_vio ([used here](https://github.com/ILLIXR/ILLIXR/blob/0329c6b5ad6b3a505e06c7a3ccdbd912fb0dae75/offload_vio/device_tx/plugin.cpp#L232)

configs/openxr_app.yaml
qinjunj: yes, will update.

ground_truth_slam/plugin.cpp
qinjunj: added, also explained later in the code.

qinjunj:
Closed. Resolved in #405

PR #393, opened 2023-07-21
Generalized dataset plugin

wermos:
This PR adds the generalized dataset plugin to ILLIXR.


wermos:
I think we can start the code review process on this. I still need some help with the build system. Other than that, it's good to go.

wermos:
Ping @jianxiapyh @hungdche for review

hungdche:
`gtsam_integrator` fails to compile in my local machine because `data_format` is being modified.

PR #394, opened 2023-07-22
Fix zed

qinjunj:
Fix the shaking issue with ZED after separating the imu and camera. 

The changes concern the follows:

- How to set IMU and camera timestamps in ZED plugin. Previously (before separating IMU and camera) all camera images were given the timestamp of the nearest IMU sample. This helps making sure that IMU and Camera streams are on the same time frame and share the same starting time point, but the camera timestamps are then skewed. I wonder if there's any downstream functions (OpenVINS for e.g.) that require all camera images having corresponding IMU sample? 
- The use of cam_to_imu time offset in IMU integrator. The old implementation applies a zero offset to the starting point of the integration and a OpenVINS estimated offset to the ending point. The offset will only take effects on the ending point if it's negative because the IMU samples for future timestamps are not available. But why would we want to integrate to a past time point rather than the current time point?

Please test with the [fix-zed branch](https://github.com/ILLIXR/open_vins/tree/fix-zed) of OpenVINS.

### DIFF
--- a/gtsam_integrator/plugin.cpp
+++ b/gtsam_integrator/plugin.cpp
@@ -17,6 +17,9 @@
 #include <iomanip>
 #include <thread>
 
+#include <filesystem>
+#include <fstream>
+
 using namespace ILLIXR;
 // IMU sample time to live in seconds
 constexpr duration IMU_TTL{std::chrono::seconds{5}};
@@ -31,6 +34,14 @@ public:
         , _m_clock{pb->lookup_impl<RelativeClock>()}
         , _m_imu_integrator_input{sb->get_reader<imu_integrator_input>("imu_integrator_input")}
         , _m_imu_raw{sb->get_writer<imu_raw_type>("imu_raw")} {
+
+        if (!std::filesystem::exists(data_path)) {
+            if (!std::filesystem::create_directory(data_path)) {
+                std::cerr << "Failed to create data directory.";
+            }
+        }
+        raw_csv.open(data_path + "/imu_raw.csv");
+
         sb->schedule<imu_type>(id, "imu", [&](switchboard::ptr<const imu_type> datum, size_t) {
             callback(datum);
         });
@@ -59,6 +70,9 @@ private:
 
     [[maybe_unused]] time_point last_cam_time;
     duration                    last_imu_offset;
+
+    const std::string data_path = std::filesystem::current_path().string() + "/recorded_data";
+    std::ofstream raw_csv;
 
     /**
      * @brief Wrapper object protecting the lifetime of IMU integration inputs and biases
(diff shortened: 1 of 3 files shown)


gtsam_integrator/plugin.cpp
```diff
@@ -176,8 +176,11 @@ class gtsam_integrator : public plugin {
 
         assert(_pim_obj != nullptr && "_pim_obj should not be null");
 
+        // TODO last_imu_offset is 0, t_offset only take effects when it's negative.
```
rsingh76, line 179: Does this make any difference in quality?

zed/src/main.cpp
```diff
@@ -111,22 +123,14 @@ class zed_camera_thread : public threadloop {
...
         // Retrieve images
         zedm->retrieveImage(imageL_zed, VIEW::LEFT_GRAY, MEM::CPU, image_size);
         zedm->retrieveImage(imageR_zed, VIEW::RIGHT_GRAY, MEM::CPU, image_size);
         zedm->retrieveMeasure(depth_zed, MEASURE::DEPTH, MEM::CPU, image_size);
         zedm->retrieveImage(rgb_zed, VIEW::LEFT, MEM::CPU, image_size);
 
-        _m_cam.put(_m_cam.allocate<cam_type>({cam_time_point, cv::Mat{imageL_ocv}, cv::Mat{imageR_ocv}}));
-
-        _m_rgb_depth.put(_m_rgb_depth.allocate<rgb_depth_type>({cam_time_point, cv::Mat{rgb_ocv}, cv::Mat{depth_ocv}}));
+        _m_cam.put(_m_cam.allocate<cam_type_zed>(
```
rsingh76, line 132: Shouldn't this be deepcopy?

zed/src/main.cpp
```diff
@@ -179,7 +186,9 @@ class zed_imu_thread : public threadloop {
             _m_first_imu_time  = imu_time;
             _m_first_real_time = _m_clock->now();
         }
-        time_point imu_time_point{*_m_first_real_time + std::chrono::nanoseconds(imu_time - *_m_first_imu_time)};
+        time_point imu_time_point{std::chrono::nanoseconds(imu_time - *_m_first_imu_time)};
```
rsingh76, line 189: Can you add some comments here, also why are we going to sleep?

rsingh76:
Overall, this is significantly better than the master, shaking is gone. But it is both jittery and laggy inside the headset.

gtsam_integrator/plugin.cpp
qinjunj: For EuRoc, `t_offset` is very small (less than 1ms) so no difference made. For ZED, this value ranges from -10ms to -60ms and a big difference there could be. My experience is that not adding this offset gives a better headset experience. I wonder if ZED needs a calibration to update the camera extrinsics, intrinsics, gyroscope/accelerator noises and random walks, and so on. These values contribute to the estimates of the `t_offset` which is so off right now.

zed/src/main.cpp
qinjunj: Yes it should be.

rsingh76 (approved):
Tested all the apps on the headset

jianxiapyh (dismissed):

rsingh76 (approved):

jianxiapyh (approved):

PR #396, opened 2023-07-26
Vulkan Implementation of ILLIXR's Rendering Pipeline with Monado Integration

shg8:
This PR contains a reimplementation of ILLIXR's rendering pipeline using the Vulkan graphics API, providing two new targets in the configuration folder, `monado-vulkan` and `native-vulkan`, which corresponds to the antecedent `monado` and `native` targets. More specifically, this PR contains the following changes:
* Replaces plugins using OpenGL with their Vulkan counterpart
    * `gldemo` -> `vkdemo`
    * `timewarp_gl` -> `timewarp_vk`
* Add new plugins in support of the new architecture
    * `display_vk` initializes Vulkan context and swapchain
    * `native_renderer` handles main render loop, resource creation, and synchronization
* `vkdemo` and `timewarp_vk` are designed as services, meaning that they have to be scheduled externally. `native_renderer` and Monado performs the scheduling in their respective versions.
* New data structures are added to the runtime. No existing data structures have been modified in a way that breaks compatibility.


native_renderer/README.md
Jebbly, line 1: I'm inclined to maintain symmetry with our OpenGL version, such that `vkdemo` is the threadloop. I.e. can we merge `vkdemo` with `native_renderer` and just have that renderer call `timewarp_vk`?

Jebbly (commented):
Overall, looks like it's in a good state. I left some comments that are primarily nitpicky or for future PRs. Other than that, I think a question for discussion is how much parity we want to maintain with our OpenGL version.

common/vk_util/render_pass.hpp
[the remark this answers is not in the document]
shg8: Yeah, it definitely would seem that way since we only have one warping algorithm right now. However, since both timewarp_vk and vkdemo are designed as services, we need to have these abstractions in order to register them in phonebook without having the whole logic in the commons folder.

common/vk_util/display_sink.hpp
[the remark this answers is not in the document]
shg8: With the current design, we can't since Monado also creates a display_sink.

common/vk_util/third_party/vk_mem_alloc.h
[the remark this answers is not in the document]
shg8: We can definitely consider better ways to ship the libraries we sue, but for this specific one it's going to be challenging as it's distributed in GitHub releases as a single header library.

configs/monado.yaml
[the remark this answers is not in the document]
shg8: Yes, I'm thinking that you could rename it to monado-gl in yoru branch so that we have both.

configs/native-gl.yaml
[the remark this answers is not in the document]
shg8: Up to Doug since he's going to convert all of these to his build system. I think either is fine.

configs/openxr_app.yaml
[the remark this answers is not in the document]
shg8: Yes, feel free to change is in your branch, but I think Doug's taking care of it anyways.

common/record_logger.hpp
```diff
@@ -160,7 +160,7 @@ class record {
 #ifndef NDEBUG
         if (rh && !data_use_indicator_.is_used()) {
             std::cerr << "Record was deleted without being logged." << std::endl;
-            abort();
+            // abort();
```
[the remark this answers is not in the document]
shg8: I think it should be fine. I ran into some problems with plugin initialization orders in which the sqlite thread is launched after records have started coming in.

configs/monado-vulkan.yaml
```diff
@@ -1,12 +1,7 @@
 plugin_groups:
+  - !include "rt_slam_plugins.yaml"
```
rsingh76, line 2: did you miss audio?
In master it is added as a part of core_plugins.yaml

configs/monado.yaml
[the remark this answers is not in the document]
rsingh76: still, let's not delete it in this merge.

configs/native-vulkan.yaml
```diff
@@ -0,0 +1,35 @@
+plugin_groups:
+  - !include "rt_slam_plugins.yaml"
+  - plugin_group:
+      - path: display_vk/
+      - path: timewarp_vk/
+      - path: vkdemo/
```
rsingh76, line 6: you can either add core_plugins.yaml or audio here

runner/runner/main.py
```diff
@@ -212,13 +219,12 @@ def load_monado(config: Mapping[str, Any]) -> None:
...
-    is_mainline: bool = bool(config["action"]["is_mainline"])
-    build_runtime(config, "so", is_mainline=is_mainline)
+    is_vulkan: bool = bool(config["action"]["is_vulkan"])
+    build_runtime(config, "so", is_vulkan = True)
 
     def process_plugin(plugin_config: Mapping[str, Any]) -> Path:
-        if is_mainline:
-            plugin_config.update(ILLIXR_MONADO_MAINLINE="ON")
-        return build_one_plugin(config, plugin_config)
+        plugin_config.update(ILLIXR_MONADO="ON")
```
rsingh76, line 226: Is this redundant in the build_one_plugin?

runner/runner/main.py
```diff
@@ -232,17 +238,19 @@ def process_plugin(plugin_config: Mapping[str, Any]) -> Path:
...
         XR_RUNTIME_JSON=str(monado_path / "build" / "openxr_monado-dev.json"),
+        XRT_TRACING="true",
+        KIMERA_ROOT=config["action"]["kimera_path"],
+        AUDIO_ROOT=config["action"]["audio_path"],
     )
 
     ## For CMake
     monado_build_opts: Mapping[str, str] = dict(
-        CMAKE_BUILD_TYPE=cmake_profile,
+        CMAKE_BUILD_TYPE="Debug",
```
rsingh76, line 248: Shouldn't be hardcoded to Debug

runner/runner/main.py
```diff
@@ -252,85 +260,104 @@ def process_plugin(plugin_config: Mapping[str, Any]) -> Path:
...
-        openxr_app_path     = None
-        openxr_app_bin_path = pathify(openxr_app_obj["app"], root_dir, cache_path, True, False)
-
-    ## Compile the OpenXR app if we received an 'app' with 'src_path'
-    if openxr_app_path:
-        cmake(
-            openxr_app_path,
-            openxr_app_path / "build",
-            dict(CMAKE_BUILD_TYPE=cmake_profile, **openxr_app_config),
+    ## Compile OpenXR apps
```
rsingh76, line 263: You're planning to compile all the OpenXR apps?
Does this also include the 2 Unreal apps that we have

runner/runner/main.py
```diff
@@ -252,85 +260,104 @@ def process_plugin(plugin_config: Mapping[str, Any]) -> Path:
...
-            Path(env_monado_service['XDG_RUNTIME_DIR'] + "/monado_comp_ipc").unlink(missing_ok=True)
-            Path("/tmp/monado_comp_ipc").unlink(missing_ok=True)
-
-        print("\nstdout:\n")
-        sys.stdout.buffer.write(outs)
+    ## Give the Monado service some time to boot up and the user some time to initialize VIO
+    time.sleep(10)
+
+    # Launch all OpenXR apps after the service is launched
+    for openxr_app_bin_path in openxr_app_bin_paths:
```
rsingh76, line 342: Should we launch all the apps?

common/record_logger.hpp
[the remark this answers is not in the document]
rsingh76: Add detailed comments here

rsingh76:
Overall, this requires a lot of comments in all the places.
Why were certain decisions (inheritance/classes/functions) made and the rationale behind those decisions.
I am going to be nit-picky about this.

common/vk_util/display_sink.hpp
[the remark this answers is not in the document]
rsingh76: Can you add detailed comments here.
Why did older version didn't need this or Why does Vulkan implementation need this (specially with Monado)?
You can also add any future plans that you might have (This goes with all the new classes you created).

common/vk_util/render_pass.hpp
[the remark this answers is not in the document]
rsingh76: Again, can you add detailed comments here. Why did older version didn't need this or Why does Vulkan implementation need this (specially with Monado)? You can also add any future plans that you might have (This goes with all the new classes you created).

display_vk/README.md
```diff
@@ -0,0 +1,3 @@
+# display_vk
+## Summary
+`display_vk` defines a service called `display_vk` which inherits from `display_sink`. The `display_vk` class is responsible for setting up and managing a Vulkan-based display sink. The `setup_vk()` function initializes GLFW and Vulkan, while the `recreate_swapchain()` function recreates the Vulkan swapchain. The `setup_vk()` function creates a Vulkan instance using the `vkb::InstanceBuilder` class from the `vkb` library. The instance is configured with validation layers and a debug callback function.
```
rsingh76, line 3: Add minor details here like - 
1. is this needed for native or monado or both? if this is needed by only one, then why? What was the reasoning behind this class/inheritance etc.
2. Why do we need to recreate Vulkan Swapchain in this case

native_renderer/plugin.cpp
```diff
@@ -0,0 +1,477 @@
+#include <cassert>
```
rsingh76, line 1: The file has barely any comments. please add detailed comments to what each function/code chunk is doing

timewarp_vk/shaders/tw.frag
```diff
@@ -0,0 +1,14 @@
+#version 450
```
rsingh76, line 1: Add some more details here and in the .vert file

vkdemo/shaders/demo.frag
```diff
@@ -0,0 +1,12 @@
+#version 450
```
rsingh76, line 1: Same as timewarp comment - add more details

vkdemo/plugin.cpp
```diff
@@ -0,0 +1,793 @@
...
+private:
+    void update_uniform(const pose_type& fp, int eye) {
+        Eigen::Matrix4f modelMatrix = Eigen::Matrix4f::Identity();
+
+        auto pose = fp;
+
+        Eigen::Matrix3f head_rotation_matrix = pose.orientation.toRotationMatrix();
+
+        constexpr int LEFT_EYE = 0;
+
```
rsingh76, line 177: Appropriate level of commenting - it would be good to have this level of comments in other parts of the code

timewarp_vk/utils/hmd.hpp
```diff
@@ -0,0 +1,58 @@
+#ifndef _HMD_H
```
rsingh76, line 1: Please add detailed comments like why we needed this addition. How is this used etc. Similar to what I have said in otehr parts of the code

timewarp_vk/plugin.cpp
```diff
@@ -0,0 +1,752 @@
...
+
+        vkDestroyPipelineLayout(ds->vk_device, pipeline_layout, nullptr);
+        pipeline_layout = VK_NULL_HANDLE;
+
+        vkDestroyDescriptorPool(ds->vk_device, descriptor_pool, nullptr);
+        descriptor_pool = VK_NULL_HANDLE;
+    }
+
+    virtual void update_uniforms(const pose_type render_pose) override {
+        // Generate "starting" view matrix, from the pose sampled at the time of rendering the frame
```
rsingh76, line 158: Good level of comments, please emulate this in the rest of the code.

configs/openxr_app.yaml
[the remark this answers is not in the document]
rsingh76: Please confirm with Doug

common/vk_util/render_pass.hpp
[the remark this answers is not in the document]
Jebbly: Could we just have timewarp_vk and vk_demo register themselves in their respective plugins? It feels a little more readable to me that way because most other plugins handle that, with pose_prediction being one of the exceptions since it has multiple implementations.

common/vk_util/third_party/vk_mem_alloc.h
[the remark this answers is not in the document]
Jebbly: Okay, let's leave this for the CMake PR then.

native_renderer/README.md
Jebbly: I think the main argument for keeping one of them, e.g. timewarp_vk, as a thread loop is so that it might be able to at least run asynchronously when running without Monado. Also in terms of naming, it might get confusing again since `native-gl.yaml` wouldn't be using the native renderer.

common/vk_util/render_pass.hpp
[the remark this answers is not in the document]
shg8: Currently, timewarp_vk and vkdeo do register themselves in their respective plugins. There are two classes in the plugin file, one service and one plugin, similar to pose_prediction.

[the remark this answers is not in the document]
shg8: I think headers should be placed under commons. We probably don't want Monado to import a plugin header and become plugin-dependent.

native_renderer/README.md
shg8: That's a good point. Timewarp is did synchronous for vkdemo in the release for the following reasons that I considered:
1. For timewarp to have its own main loop, we'd have to do explicit synchronization between vkdemo and timewarp, which could be fairly complicated. 
2. Having a main loop is not enough. In fact, it needs to have its own queue with a higher priority than the app queue in order for it to choke a fully saturated app workload. There are further complications to achieve this:
    * The Vulkan spec doesn't guarantee queue priorities. They are provided to the drivers, and the drivers may elect to do whatever they want.
    * Maybe @rsingh76 can chime in, but to properly make it async a lot of scheduling is involved, which is out of the scope of this release
3. vkdemo's workload should never require async warping. In the original docs, it is only meant as a super lightweight app to test the pipeline. For any heavier apps, I think we should focus on OpenXR support.

configs/monado-vulkan.yaml
shg8: I added them back in. Had to remove timewarp_gl from core plugins.

runner/runner/main.py
shg8: I haven't changed this part since I copied Jeffrey's script.

runner/runner/main.py
shg8: Good catch

runner/runner/main.py
shg8: I think Jeffrey added this to support multi-client. Hasn't changed this part from this script.

native_renderer/README.md
Jebbly: Fair enough, any workarounds are probably out of the scope for this PR.

common/vk_util/render_pass.hpp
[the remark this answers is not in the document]
Jebbly: Hm, if it's something that Monado needs then it would be hard to work around. But since we're already kind of hacky with our Monado integration, I think it would be preferred to make our native repo as clean and understandable as possible and do what we need with our Monado integration for the time being. 

Personally, I think it might even be more understandable if we mimic what `timewarp-semaphore` is doing where `timewarp` is a plugin again but scheduled synchronously to a topic. Let's see what @qinjunj @rsingh76 has to say about this.

runner/runner/main.py
Jebbly: This is fine, some plugins (e.g. timewarp_gl) might want to use ifdefs on their own. Although we may potentially want to find a better solution in the future.

runner/runner/main.py
Jebbly: Yes, this also addresses the previous comment because all potentially compilable OpenXR apps (like openxr-simple-example) will be built here now. It's a bit rough around the edges but I think it's okay to keep for now, unless we don't want to support multiple clients yet.

runner/runner/main.py
Jebbly: I believe this is addressed in the comment below.

timewarp_vk/utils/hmd.hpp
shg8: That's just equivalent to `#pragma once`, which prevents double inclusion.

common/vk_util/render_pass.hpp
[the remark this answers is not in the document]
shg8: I think the challenge here with Vulkan is that we can't just pass around a texture handle anymore due to synchronization. If we have it on a topic, we'd have to verify that it's not being used. That involves device to host synchronization. 

Right now, we have implicit sync like this:
`app pass (GPU) -> layout transition to shader read (GPU) -> tw (GPU) -> layout transition to color attachment (GPU)`

However, if we mimic timewarp-semaphore, we'll have 
`app pass (GPU) -> layout transition to shader read (GPU) -> fence signaled (GPU->CPU) -> publish on switchboard (CPU->CPU) -> tw submits command buffer (CPU->GPU) -> tw (GPU) -> layout transition to color attachment (GPU)`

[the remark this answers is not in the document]
Jebbly: Yeah, it's definitely a bit more annoying to handle than the OpenGL side where it's not as picky about synchronization. I'm slightly willing to compromise here, but I think we should handle it better in the future.

For example, we could implicitly guarantee that any buffer in the event stream is no longer being written to by timewarp, and timewarp is writing to another buffer at the moment. Then when we publish again, timewarp should make sure that it first updates the topic before starting to write again to the previous buffer. 

I imagine that, depending on how we want to go, timewarp will need to submit its own command buffer at some point. The idea here would be that we would need some fence for the first publish to switchboard, but afterward, timewarp could run asynchronously.  It's not a terrible trade-off in my opinion since Monado was already using fences around its layer compositor. We could potentially also consider how a rendergraph builder might fit into all of this. 

Again could be handled in a future fix since I'm a bit fuzzy on the details still, need to take a closer look at monado_vulkan_integration.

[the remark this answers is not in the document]
shg8: Yeah, I agree we can come back to this at a future date.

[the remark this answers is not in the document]
Jebbly: For this PR though, let's at least rename `timewarp` to `monado_timewarp` or something so it's more clear that it's only being used for monado.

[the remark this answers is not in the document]
shg8: But native also uses it, no?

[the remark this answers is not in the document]
Jebbly: Oh right, maybe `compositor_vk` then. Also works better since we do use `timewarp_gl` with Monado, which I totally forgot about too. Just something to make it clear that `timewarp_gl` should not inherit it.

rsingh76:
Debug window is blank for native?

display_vk/plugin.cpp
```diff
@@ -0,0 +1,199 @@
...
+        swapchain_image_views  = vkb_swapchain.get_image_views().value();
+        swapchain_image_format = vkb_swapchain.image_format;
+        swapchain_extent       = vkb_swapchain.extent;
+    }
+
+    /**
+     * @brief This function polls GLFW events. See display_sink::poll_window_events().
+     */
+    void poll_window_events() override {
+        glfwPollEvents();
```
Jebbly, line 51: Not sure if others have had the same issue as me, but I'm getting an issue here with:
```
X Error of failed request:  BadMatch (invalid parameter attributes)
  Major opcode of failed request:  149 ()
  Minor opcode of failed request:  4
  Serial number of failed request:  230
  Current serial number in output stream:  238
```
I'm not _quite_ sure what the issue is, but I'm guessing it has to do with the fact that [glfwPolllEvents() can only be called from the main thread](https://www.glfw.org/docs/3.3/group__window.html#ga37bd57223967b4211d60ca1a0bf3c832]. Running `gdb`, it seems like the thread that calls `setup_glfw()` is different from the thread that ends up calling `poll_window_events()`. I think the easiest workaround here is to just remove this function call (i.e. this function is no-op) because it doesn't seem like we're doing anything with the glfw events anyways.

shg8: On some DEs such as Gnome, the window would freeze and gray out if we don't poll events. I'll look into the error.

display_vk/plugin.cpp
```diff
@@ -0,0 +1,228 @@
...
+    void setup_glfw() {
+        if (!glfwInit()) {
+            ILLIXR::abort("Failed to initalize glfw");
+        }
+
+        glfwWindowHint(GLFW_CLIENT_API, GLFW_NO_API);
+        glfwWindowHint(GLFW_RESIZABLE, GL_FALSE);
+
+        window =
+            glfwCreateWindow(display_params::width_pixels, display_params::height_pixels, "Vulkan window", nullptr, nullptr);
```
Jebbly, line 72: Can we rename this to something like "ILLIXR Extended Vulkan Window" or something similar? And then incorporate preprocessor ifdefs to disable the `xlib_gl_extended_window` in `runtime_impl` when Vulkan is enabled. Right now just seems to be an empty window for me. Otherwise LGTM.

shg8: I just renamed it to the eyebuffer window. I enabled debugview in again, so we would still need the extended window for that to work properly.

Jebbly: I think debugview has its own GLFW/GL context and is able to run fine (at least, it doesn't seem to consume anything provided by the extended window, and it runs fine for me when I comment it out manually in runtime_impl. What does the behavior look like for you?

shg8: Yes, you're right. Fixed.

Jebbly (approved):
Overall looks ready to merge, might have to rebase first though

common/vk_util/render_pass.hpp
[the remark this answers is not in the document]
shg8: Right now, it's in the vk_util folder, which should be sufficient to impart that they're not for GL. However, I do agree that we could structure this better in the future.

PR #395, opened 2023-07-26
Vulkan-OpenGL Interop with Monado

Jebbly:
Updated PR to #373.

PR #404, opened 2023-08-21
Rebase OpenGL Support

Jebbly:
This PR rebases the changes from #395 on top of the new Vulkan changes. The main purpose is to update our OpenGL infrastructure to support Monado's out-of-process compositor and to maintain feature parity as well as possible with our Vulkan backend.

The main changes are:
- Creating a ``native-gl.yaml`` config file to enable 
- Modifying macros to define ``ILLIXR_VULKAN`` and ``ILLIXR_MONADO`` for the appropriate backends. This is necessary because timewarp's behavior may change depending on if ``native-gl.yaml`` or ``monado-gl.yaml`` is being used (an example is provided in the following bullet point)
- ``timewarp_gl`` now runs as a plugin with Monado, and a threadloop with native.


shg8 (commented):
LGTM. I assume the Vulkan versions are also tested?

astro-friedel (dismissed):

Jebbly:
> LGTM. I assume the Vulkan versions are also tested?

Yeah, they were working for me the last time I checked. It would be good to verify if it can run on multiple machines though.

gldemo/plugin.cpp
```diff
@@ -302,9 +304,18 @@ class gldemo : public threadloop {
         glEnable(GL_DEBUG_OUTPUT);
         glDebugMessageCallback(MessageCallback, 0);
 
-        // Create two shared eye textures, one for each eye
+        // Create two shared eye textures.
+        // Note; each "eye texture" actually contains two eyes.
+        // The two eye textures here are actually for double-buffering
+        // the Switchboard connection.
+        // ^ The above comment seems to be related non-Monado style eyebuffers,
+        // ^ which seem to no longer be used.
```
qinjunj, line 312: Please fix the comment to reflect the current implementation.

timewarp_gl/plugin.cpp
```diff
@@ -185,10 +311,75 @@ class timewarp_gl : public threadloop {
...
+        }
+        }
+    }
+
     void BuildTimewarp(HMD::hmd_info_t& hmdInfo) {
         // Calculate the number of vertices+indices in the distortion mesh.
         num_distortion_vertices = (hmdInfo.eyeTilesHigh + 1) * (hmdInfo.eyeTilesWide + 1);
         num_distortion_indices  = hmdInfo.eyeTilesHigh * hmdInfo.eyeTilesWide * 6;
+        // What is the 6 here? What's the difference between vertices and indices? Each rectangular has two triangles, each
+        // triangle has three vertices (i.e., three indices to represent).
```
qinjunj, line 382: Seems like leftover comments for understanding the code. Remove if not no longer needed.

timewarp_gl/plugin.cpp
```diff
@@ -235,7 +427,7 @@ class timewarp_gl : public threadloop {
                     // Set the physical distortion mesh coordinates. These are rectangular/gridlike, not distorted.
                     // The distortion is handled by the UVs, not the actual mesh coordinates!
                     distortion_positions[eye * num_distortion_vertices + index].x =
-                        (-1.0f + eye + ((float) x / hmdInfo.eyeTilesWide));
+                        (-1.0f + 2.0f * ((float) x / hmdInfo.eyeTilesWide));
```
qinjunj, line 430: What's the reason for this change?

timewarp_gl/plugin.cpp
```diff
@@ -428,29 +602,68 @@ class timewarp_gl : public threadloop {
         assert(distortion_indices_data != nullptr && "Timewarp allocation should not fail");
         glBufferData(GL_ELEMENT_ARRAY_BUFFER, num_distortion_indices * sizeof(GLuint), distortion_indices_data, GL_STATIC_DRAW);
 
-        if (enable_offload) {
```
qinjunj, line 431: Do we want to keep the function of `offload_data` to save rendered textures? That plugin is not currently used and is not supported by `timewarp_vk`, but in the long run it may be of use (e.g. for offloading rendering and debugging).

timewarp_gl/plugin.cpp
Jebbly: The difference here depends on how we construct the timewarp mesh. From what I remember, the previous timewarp assumed that it was operating on a single buffer, so it would need two distinct meshes. In normalized device coordinates, the x-axis of the left eye would range from [-1.0, 0] while the x-axis of the right eye would range from [0, 1.0]. 

On the other hand, when we run timewarp on each eye, we want to construct the mesh to occupy the entire width, i.e. an NDC range of [-1.0, 1.0].

common/data_format.hpp
```diff
@@ -138,36 +139,107 @@ typedef struct {
...
+                   fast_pose_type render_pose_, time_point sample_time_, time_point render_time_)
+        : swapchain_indices{std::move(swapchain_indices_)}
         , swap_indices{std::move(swap_indices_)}
         , render_pose(render_pose_)
         , sample_time(sample_time_)
         , render_time(render_time_) { }
 };
 
 struct hologram_input : public switchboard::event {
-    uint seq;
```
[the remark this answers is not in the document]
Jebbly: Sorry, this seems like a leftover change I had when rebasing from the other branch. Fixed this and also removed the other leftover structs.

timewarp_gl/plugin.cpp
Jebbly: I've re-enabled it for now, but had to make some changes to `readTextureImage()` in order for the entire texture to be copied appropriately.

shg8 (approved):

qinjunj (approved):

rsingh76 (approved):

qinjunj (approved):

Issue #89, opened 2020-09-03
Integrate ORB-SLAM3


mhuzai:
Rationale: full VI-SLAM system with loop closure. ORB-SLAM3 has maps that stay persistent across runs, and we will need an efficient mechanism in ILLIXR to store/retrieve them.

hungdche:
Will close with #407

PR #340, opened 2022-06-12
Create ORB SLAM plugin

hungdche:

PR #407, opened 2023-09-07
ORB-SLAM plugin for new CMake build

hungdche:
This closes issue #89 
Duplicates with #340, but targets `convertBuildToCmake` instead of `master`


hungdche:
Need testing on different machines before merging!

hungdche:
PR now ready to be reviewed. Please review the actual plugin [here](https://github.com/ILLIXR/ORB_SLAM3/pull/7).

cmake/GetORB_SLAM.cmake
```diff
@@ -0,0 +1,30 @@
...
+        GIT_TAG 0de3408da0d26ec44afa1f4be0def151fe54af92         # sha5 hash for specific commit to pull (if there is no specific tag to use)
+        PREFIX ${ORB_SLAM_SOURCE_DIR}                            # the build directory
+        DEPENDS ${DBoW2_OS3_DEP_STR} ${Pangolin_DEP_STR} ${g2o_DEP_STR} ${Sophus_DEP_STR} ${OpenCV_DEP_STR}   # dependencies of this module
+        # force serialized build, otherwise the machine might get slogged down
+        BUILD_COMMAND cmake --build . -j1
+        #arguments to pass to CMake
+        CMAKE_ARGS -DCMAKE_INSTALL_PREFIX=${CMAKE_INSTALL_PREFIX} -DCMAKE_CXX_FLAGS=-L${CMAKE_INSTALL_PREFIX}/lib -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE} -DILLIXR_ROOT=${PROJECT_SOURCE_DIR}/include -DILLIXR_BUILD_SUFFIX=${ILLIXR_BUILD_SUFFIX} ${ORB_SLAM_CMAKE_ARGS}
+        )
+
+execute_process(
```
hungdche, line 27: @astro-friedel this command sometimes fail to extract the tar file and I couldn't seem to know why. Any idea will be much appreciated!

astro-friedel: The 'command' has no dependencies, so it will run when cmake decides, regardless of the status of the ORB_Slam build. A dependency should be added to the 'command', or a more complex build command given to 'external_project_add'.

astro-friedel: Or, even better, move that into the ORB_Slam cmake itself

astro-friedel (dismissed):

astro-friedel (approved):

Jebbly (approved):
Rebased on master and added some changes to support the ZED. Also added some changes on the ORB-SLAM side to fix some weird errors from (possibly) Eigen optimization.

profiles/all.yaml
```diff
@@ -0,0 +1,7 @@
+# This file was auto generated and is intended for debugging an entire build, take caution if editing manually.
```
qinjunj, line 1: The cmake process cannot detect `offload_vio.device_tx` and the other three plugins correctly. It only reports that `offload_vio` is not enabled.

astro-friedel: This branch was working with an outdated plugins.yaml file. I just committed an update to fix it.

Jebbly (dismissed):

qinjunj:
In Debug mode, got the error message of `Could not load plugin dependency map file (plugin_deps.yaml), cannot verify plugin dependencies.` The error is from `src/plugin.cpp:222`, because of the first path in the `dep_list` is not full. @astro-friedel Could you please take a look?

qinjunj:
There is an inconsistency in the naming of ORB_SLAM in the repo. The generated .so file is `libplugin.orb_slam3.opt.so` while currently `orb_slam` is specified in the yaml file and is expected by the configuration process. @astro-friedel Could you please fix it? We can go with orb_slam3.

Jebbly:
> There is an inconsistency in the naming of ORB_SLAM in the repo. The generated .so file is `libplugin.orb_slam3.opt.so` while currently `orb_slam` is specified in the yaml file and is expected by the configuration process. @astro-friedel Could you please fix it? We can go with orb_slam3.

I'm in favor of removing the ``--vis`` / ``visualizer`` thing entirely, unless there is a specific purpose for it. To me, it makes more sense to just consider OpenVINS / ORB-SLAM like any other plugin (it's also not clear what "visualizer" refers to in this context).

astro-friedel:
@qinjunj I have renamed stuff to orb_slam3

astro-friedel:
@Jebbly The --visualizer issue is being addressed by PR #439

astro-friedel (dismissed):

src/plugin.cpp
```diff
@@ -169,7 +169,8 @@ int ILLIXR::run(const cxxopts::ParseResult& options) {
        // read in yaml config file
         YAML::Node  config;
         std::string exec_path = get_exec_path();
-        std::string home_dir  = get_home_dir();
+        setenv("ILLIXR_BINARY_PATH", exec_path.c_str(), 1);
```
qinjunj, line 172: Better use the `cmake_install_prefix/bin` as the binary path. It has trouble finding the vocab file when I ran the executable from my build directory.

astro-friedel: The CMAKE_INSTALL_PREFIX/bin is not available to C++, unless we add it as a compile definition, which would hard code it. If that is ok, then we can do it that way.

qinjunj: I am okay with this. But what do you think would be a better solution?

astro-friedel: I am not sure.

astro-friedel (approved):

qinjunj (approved):

astro-friedel (approved):

PR #409, opened 2023-09-15
Offload rebase

qinjunj:
This PR rebases on top of the new master and closes #405. All comments at #405 are resolved.


jianxiapyh (dismissed):
The code looks good to me for now. However, I can only run 'offload-device' and not 'offload-monado', since your current code assumes that CMake 3.24 is being used (otherwise, 'timewarp_vk' will not compile). However, Steven mentioned that Doug might have fixed this in a later branch, so I am inclined to approve it for now. 
**with 3.24, both offload-device and offload-monado worked**
**with 3.20, only offload-device worked**

jianxiapyh (approved):

Boyuan-Tian (approved):

Issue #118, opened 2020-09-17
Unify File I/O


mhuzai:
All the SLAM dataset readers -- offline_imu_cam, ground_truth_slam, pose_lookup -- roll their own reader, and the readers all look the same. We should have a common `dataset_reader` class that everyone can use.

charmoniumQ:
Also see #51 .

PR #416, opened 2024-03-12
Unify file io - issue #118

astro-friedel:
This addresses issue #118. The file loading has been compacted into a single header. The header code will check for file existence and open it. Anything that want to use this functionality will need to supply a function which actually reads the data into the proper structures.


include/illixr/data_loading.hpp
```diff
@@ -0,0 +1,34 @@
...
+#include <map>
+#include <spdlog/spdlog.h>
+#include <string>
+
+using namespace ILLIXR;
+
+template<typename T>
+static std::map<ullong, T> load_data(const std::string& spath, const std::string& plugin_name,
+                                     std::map<ullong, T> (*func)(std::ifstream&, const std::string&)) {
+    const char* illixr_data_c_str = std::getenv("ILLIXR_DATA");
```
Jebbly, line 19: For generality, we might consider adding another string argument instead of hard-coding `ILLIXR_DATA`. Maybe same for `data.csv` below.

Jebbly:
Overall the changes look fine and can be merged soon, but I'm not sure if this fully addresses #118. My impression of that issue is that we want a central _plugin_ that can replay things, rather than having a separate plugin for each data file.

jianxiapyh:
Overall, this looks solid and is a step towards addressing #118. However, it may not fully resolve the issue yet.

@Jebbly , as you know, the XR Fellowship member was working on a universal dataset reader in #393 some times ago but didn't complete it before stepping away. Now that it seems he is back, we can check on the status of that solution and see if it can complement this approach.

jianxiapyh:
Hey @astro-friedel , did anything change since our comments?

jianxiapyh (approved):

include/illixr/data_loading.hpp
```diff
@@ -0,0 +1,34 @@
...
+#include <fstream>
+#include <iostream>
+#include <map>
+#include <spdlog/spdlog.h>
+#include <string>
+
+using namespace ILLIXR;
+
+template<typename T>
+static std::map<ullong, T> load_data(const std::string& spath, const std::string& plugin_name,
```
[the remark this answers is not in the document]
astro-friedel: As far as I know T can be anything since you can put anything as the value for a map pair. If it is incompatable, the compiler should throw an error.

include/illixr/data_loading.hpp
astro-friedel: Possibly, but I think that issue would be better addressed in the generalized data set plugin which is under development.

Jebbly (approved):

Issue #408, opened 2023-09-14
Dependency handling for plugin loader


shg8:
ILLIXR loads the plugins by the order they are passed in right now. Dependency conflicts easily arise for plugins requesting other services and better handling is required. For now, the plugin list has been manually reordered. However, we should consider using a dependency graph in the future.

PR #417, opened 2024-03-12
Tracking dependencies - issue #408

astro-friedel:
This code will automatically parse the code base looking for plugins which require other plugins to be loaded first. It then orders the plugins.yaml file entries so that plugins are loaded in the proper order.


README.md
```diff
@@ -90,15 +90,21 @@ We request that you cite our following [paper][8] when you use ILLIXR for a publ
...
-@inproceedings{HuzaifaDesai2021,
-  author={Huzaifa, Muhammad and Desai, Rishi and Grayson, Samuel and Jiang, Xutao and Jing, Ying and Lee, Jae and Lu, Fang and Pang, Yihan and Ravichandran, Joseph and Sinclair, Finn and Tian, Boyuan and Yuan, Hengzhi and Zhang, Jeffrey and Adve, Sarita V.},
-  booktitle={2021 IEEE International Symposium on Workload Characterization (IISWC)}, 
-  title={ILLIXR: Enabling End-to-End Extended Reality Research}, 
-  year={2021},
-  volume={},
-  number={},
-  pages={24-38},
-  doi={10.1109/IISWC53511.2021.00014}
+@ARTICLE {9741292,
```
Jebbly, line 93: Could we also split this PR so that documentation updates are reviewed / merged separately?

astro-friedel: I removed this change

docs/docs/illixr_plugins.md
```diff
@@ -196,8 +196,24 @@ ILLIXR supports additional plugins to replace some of the default plugins.
...
+
+Some plugins require other plugins to be loaded in order to work. The table below gives a listing of the plugin interdependencies.
+
+| Plugin          | Requires        | Provided by plugin                      |
+|:----------------|:----------------|:----------------------------------------|
+| debugview       | pose_prediction | faux_pose, pose_lookup, pose_prediction |
+| gldemo          | pose_prediction | faux_pose, pose_lookup, pose_prediction |
+| native_renderer | app             | vkdemo                                  |
+|                 | display_sink    | display_vk                              |
+|                 | pose_prediction | faux_pose, pose_lookup, pose_prediction |
```
shg8, line 209: Just so that I don't forget when we merge in the offload rendering branch, could we add `lighthouse` to pose_prediction

src/plugin.cpp
```diff
@@ -2,124 +2,295 @@
...
+
+    bool operator==(const std::string& rhs) const {
+        return name == rhs;
+    }
+};
+} // namespace ILLIXR
+
+namespace YAML {
+template<>
+struct convert<ILLIXR::Dependency> {
```
shg8, line 31: Just on naming conventions, I think we're doing lower case from struct names?

shg8 (approved):

src/plugin.cpp
astro-friedel: ok, will fix on merge.

Jebbly (approved):
LGTM. We can try something like pybind sometime down the line as well.

qinjunj (commented):
LGTM

docs/docs/modules.json
```diff
@@ -55,6 +55,7 @@
         "glm",
         "OpenXR",
         "libusb",
+        "libuvc",
```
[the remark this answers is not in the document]
astro-friedel: usb video devices.  I added it because it was required to build under Ubuntu 24, and there is no harm in including it. Though we don't require it directly, some of the packages we use require it, so it is added for completeness.

PR #418, opened 2024-03-19
Fix changing profile names

astro-friedel:
The current code does not properly change the used plugins when a different profile is used in cmake. This change fixes this issue byt not caching the assembled full path of the profile file.

### DIFF
--- a/cmake/HelperFunctions.cmake
+++ b/cmake/HelperFunctions.cmake
@@ -6,7 +6,7 @@ function(generate_yaml)
 
     file(TIMESTAMP "${PROJECT_SOURCE_DIR}/plugins/plugins.yaml" PROFILE_FILE_TIMESTAMP "%s" UTC)
 
-    if(DEFINED CACHE{LAST_YAML_BUILD} AND DEFINED CACHE{PLUGIN_LIST} AND DEFINED CACHE{VISUALIZER_LIST_LIST})
+    if(DEFINED CACHE{LAST_YAML_BUILD} AND DEFINED CACHE{PLUGIN_LIST} AND DEFINED CACHE{VISUALIZER_LIST})
         #message("Profile files are up to date")
     else()
         set(LAST_YAML_BUILD "0" CACHE INTERNAL "")
@@ -161,6 +161,7 @@ function(read_yaml FILENAME)
         FIND_FILE(FILEPATH ${FILENAME}
                 PATHS profiles ${CMAKE_SOURCE_DIR} $ENV{HOME} $ENV{HOME}/.illixr
                 PATH_SUFFIXES profiles
+                NO_CACHE
                 )
         if(${FILEPATH} STREQUAL "FILEPATH-NOTFOUND")
             message(FATAL_ERROR "Could not find ${FILENAME} in any of the following paths:\n    profiles\n    ${CMAKE_SOURCE_DIR}\n    ${CMAKE_SOURCE_DIR}/profiles\n    $ENV{HOME}\n    $ENV{HOME}/profiles\n    $ENV{HOME}/.illixr\n    $ENV{HOME}/.illixr/profiles")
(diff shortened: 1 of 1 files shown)


Jebbly:
Should the first commit be part of this PR?

astro-friedel:
Fixed

cmake/HelperFunctions.cmake
```diff
@@ -6,7 +6,7 @@ function(generate_yaml)
 
     file(TIMESTAMP "${PROJECT_SOURCE_DIR}/plugins/plugins.yaml" PROFILE_FILE_TIMESTAMP "%s" UTC)
 
-    if(DEFINED CACHE{LAST_YAML_BUILD} AND DEFINED CACHE{PLUGIN_LIST} AND DEFINED CACHE{VISUALIZER_LIST_LIST})
+    if(DEFINED CACHE{LAST_YAML_BUILD} AND DEFINED CACHE{PLUGIN_LIST} AND DEFINED CACHE{VISUALIZER_LIST})
```
Jebbly, line 9: Is it necessary to check the cache for all 3 variables? Or it is sufficient to just check for the ``LAST_YAML_BUILD``? If it's the latter case, I think it would be more digestible to have something like:
```
if(NOT DEFINED CACHE{LAST_YAML_BUILD})
    set(LAST_YAML_BUILD "0" CACHE INTERNAL "")
endif()
```

Jebbly (commented):
Tested with ``native_vk.yaml`` and ``native_gl.yaml``, LGTM. Left one small question / comment to address.

Out of curiosity, is it possible to also avoid rebuilding plugins that have been built without change? E.g. it seems like ``debugview`` is built again when I swap profiles, even though it has already been built before.

cmake/HelperFunctions.cmake
astro-friedel: I do not know about controlling the rebuilding of plugins. It comes down to how CMake decides something is stale. If the number of targets (e.g. plugins) changes, or their order changes it may just decide that everything is stale. I am not sure there is much we can do.

Regarding the check, yes we need to check all three because each is needed. There are ways that these entries could be removed/invalidated from the cache, so we need to be sure they are there.

Jebbly: Got it. In that case, I think we should keep the ``message`` and have it say something like, "Using cached profiles" or whatever that should represent.

Jebbly (dismissed):
LGTM after the comment I left is addressed.

cmake/HelperFunctions.cmake
astro-friedel: Just pushed a fix which adds a message

Jebbly (approved):

Jebbly (approved):
LGTM, tested earlier with a few different plugins and seems to work as expected.

sriramdvt:
Looks good to me.

Output of cmake when a different profile is used before this PR:
```
root@5a0462f043ca:/home/ILLIXR/build# cmake .. -DYAML_FILE=native_gl.yaml
Found OS: ubuntu
Using cached profile yaml files
-- Reading profile file /home/ILLIXR/profiles/native_lookup.yaml
...
```

Output of cmake when a different profile is used after this PR:
```
root@5a0462f043ca:/home/ILLIXR/build# cmake .. -DYAML_FILE=native_gl.yaml
Found OS: ubuntu
Using cached profile yaml files
-- Reading profile file /home/ILLIXR/profiles/native_gl.yaml
...
```

sriramdvt (approved):
Looks good to me.

Output of cmake when a different profile is used before this PR:
```
root@5a0462f043ca:/home/ILLIXR/build# cmake .. -DYAML_FILE=native_gl.yaml
Found OS: ubuntu
Using cached profile yaml files
-- Reading profile file /home/ILLIXR/profiles/native_lookup.yaml
...
```

Output of cmake when a different profile is used after this PR:
```
root@5a0462f043ca:/home/ILLIXR/build# cmake .. -DYAML_FILE=native_gl.yaml
Found OS: ubuntu
Using cached profile yaml files
-- Reading profile file /home/ILLIXR/profiles/native_gl.yaml
...
```

Issue #415, opened 2024-03-11
Generate Guidelines and Proceedures


astro-friedel:
This is an agglomeration of 3 related issues (#60, #357, and #359 )

The following guidelines, proceedures, and documents should be created

- [x] Document how to file a bug
- [x] Document how to get help (Discord, GitHub issue, etc.)
- [x] How to suggest a feature
- [x] A code of conduct
- [x] How to cite "borrowed" code
- [x] What licenses can we use
- [x] Programming style guide
- [x] Code review guidelines

Assocaited reading:
- [GitHub Open Source Guides](https://opensource.guide)
- [TODO group guides](https://todogroup.org/guides/)
- [Governance Models](http://oss-watch.ac.uk/resources/governancemodels)
- [Guidelines for PR reviewer](https://google.github.io/eng-practices/review/reviewer/)
- [Guidelines for PR submitter](https://google.github.io/eng-practices/review/developer/)

PR #419, opened 2024-03-29
Adding documentation for policies

astro-friedel:
This is to fulfill issue #415, adding general policies and instructions for contributors.


docs/docs/contributing/style_guide.md
```diff
@@ -0,0 +1,106 @@
+## Programming Style Guide
```
Jebbly, line 1: Moving forward, we should start enforcing stylistic naming conventions as well, and not just formatting. This is an issue across plugins because different plugins use different types of variable names (e.g. camel case, snake case, etc). Especially for member variables because some plugins use an ``m_`` prefix, while others do not, which might make it confusing.

astro-friedel: I agree. I think as part of this PR I will go through and standardise as much as I can.

astro-friedel: I have gone through the code base and applied the standards best I could

Jebbly: Sorry for the late response. Follow up on this, can we also choose a standard regarding `.h` and `.hpp`? It's fine to leave third-party alone, but for example, we have `plugins/offload_vio/device_tx/video_encoder.h` but `plugins/timewarp_vk/utils/hmd.hpp`. It seems like most files use HPP, so that's what I'm in favor of.

For the future, I would also like to encourage multiple files for our plugin implementations - at the very least, a header file where the plugin class is declared along with member variables and functions. I find it easier to digest what a plugin's doing if I can see all the functions declared in a header file. I think it's also easier to parse member variables from the header file than look for where they are between `public` and `private`, etc.

astro-friedel: I agree, I will add it to the documentation and make the code changes.

astro-friedel: changes done

shg8:
We might want to add some notes about `CMAKE_INSTALL_PREFIX=<LOCATION>`. We should direct the user to install to a local directory if they don't want to install to `/usr/local/lib`. Additionally, if they pick a custom install location, they'd need to supply `LD_LIBRARY_PATH`.

shg8 (approved):
LGTM

docs/docs/writing_your_plugin.md
```diff
@@ -192,7 +192,7 @@ To add your own functionality via the plugin interface:
       [`sb->schedule(...)`][14].
 			
     - If you spin your own threads, they **must** wait for
-          `pb->lookup_impl<Stoplight>()->wait_for_ready()` the first time they
+          `pb->lookup_impl<stoplight>()->wait_for_ready()` the first time they
```
sriramdvt, line 195: Is this supposed to be lowercase? From the implementation of [`lookup_impl()`](https://github.com/ILLIXR/ILLIXR/blob/1963c106d4850a2a5749584f265592fb8c9bd899/include/illixr/phonebook.hpp#L122), the call would indeed look like `pb->lookup_impl<Stoplight>()->wait_for_ready()`.

The offloading VIO plugin that uses `lookup_impl` has [`pb->lookup_impl<Stoplight>()`](https://github.com/ILLIXR/ILLIXR/blob/1963c106d4850a2a5749584f265592fb8c9bd899/plugins/offload_vio/device_tx/plugin.cpp#L36C24-L36C52).

include/illixr/cpu_timer.hpp
```diff
@@ -159,35 +160,29 @@ class should_profile_class {
 public:
     should_profile_class() {
         const char* ILLIXR_STDOUT_METRICS = getenv("ILLIXR_STDOUT_METRICS");
-        actually_should_profile           = ILLIXR_STDOUT_METRICS && (strcmp(ILLIXR_STDOUT_METRICS, "y") == 0);
+        actually_should_profile_          = ILLIXR_STDOUT_METRICS && (strcmp(ILLIXR_STDOUT_METRICS, "y") == 0);
```
sriramdvt, line 163: Nit: should we consider renaming this private variable to `should_profile_bool_`? Not important, this comment can be ignored.

include/illixr/plugin.hpp
```diff
@@ -16,9 +17,9 @@ namespace ILLIXR {
 using plugin_id_t = std::size_t;
 
 /*
- * This gets included, but it is functionally 'private'. Hence the double-underscores.
+ * This gets included, but it is functionally 'private'. Hence, the double-underscores.
```
sriramdvt, line 20: Isn't this comment out-dated with the new name in the next line?

src/illixr.hpp
```diff
@@ -30,14 +30,14 @@
...
-#define STRINGIZE(x) _STR(x)
-#define GET_LONG(NAME, ENV)                                                    \
-    if (options.count(#NAME)) {                                                \
-        setenv(#ENV, std::to_string(options[#NAME].as<long>()).c_str(), true); \
-    } else if (config[#NAME]) {                                                \
-        setenv(#ENV, std::to_string(config[#NAME].as<long>()).c_str(), true);  \
-    }
+#define STR(y)       #y
+#define STRINGIZE(x) STR(x)
+//#define GET_LONG(NAME, ENV)                                                    \
```
sriramdvt, line 35: Nit: Build step throws a multi-line comment warning. Suggest using a `/* */` comment for the multi-line define `#define GET_LONG...`.

sriramdvt (dismissed):
Documentation for contributing looks good to me. A couple of questions and nits.

The PR also has a lot of code-style changes. I am assuming that they don't break anything since I am able to build this PR.

src/illixr.hpp
astro-friedel: I will fix this in the final merge

include/illixr/plugin.hpp
astro-friedel: yes, I will fix it in the merge

include/illixr/cpu_timer.hpp
astro-friedel: should_profile_bool_ seems redundant as it names the type in the variable. we should probably leave it as is for now

astro-friedel:
I tested as much as I could, and I believe the changes do not break anything.

astro-friedel:
I just merged in master to bring it up to date

shg8 (approved):
LGTM

sriramdvt (approved):
lgtm

PR #423, opened 2024-04-13
Change the TCP socket implementation

williamsentosa95:
I rewrote the TCP socket implementation to a minimalist one. I made sure to include all the core functionality for the offload-vio. Could you review this and port this to the master branch? Thanks!

PR #424, opened 2024-06-17
Refactor networking code and add the new minimalist tcp socket

williamsentosa95:
This is the new updated tcp code that I added from the master branch code (rather than the "offload-timing-tcp-socket" branch as in #423). I haven't run and tested this on my laptop since my laptop cannot run the VIO offload with compression from the master branch due to the driver issues. 

However, I have tested and run the #423 "offload-new-tcp-socket" that uses VIO without compression. I carefully adapt the changes from #423, but I will need help to run this and test for any bug or error.


astro-friedel (approved):

qinjunj (approved):
Tested with EuRoC and Zed, compression enabled, and ZenGarden.

astro-friedel (approved):

Issue #425, opened 2024-06-25
Spdlog has issues with set_pattern with multiple threads


astro-friedel:
Spdlog has several functions which are known to not be thread safe. One of these is `set_pattern`, which we use in several places. The most likely solutions are either to mutex these calls or to create a logging singleton to control all logging interactions. 

The mutex solution is problematic as many of these calls are in header only code (e.g. switchboard) which makes using a shared mutex from the disparate plugin libraries problematic.

Using a singleton instance of a logging class should avoid the above issue, as well as making adding logging code in plugins more straight forward (won't need to think about mutexs)

PR #426, opened 2024-09-08
Add spdlogger cmake option (#425)

e3m3:
Mitigates SPD logger thread concurrency issue (#425).

In addition to the cited concurrency issue, the project does not build on a fresh Fedora 39/40 container image. Rather than replace the functionality with a new implementation of a logger, add an optional CMake build flag which toggles the inclusion of the spdlog library dependency.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -82,6 +82,7 @@ set(DATA_FILE "http://robotics.ethz.ch/~asl-datasets/ijrr_euroc_mav_dataset/vico
 
 option(BUILD_DOCS "Build documentation" OFF)
 option(USE_MONADO "Build for monado use" OFF)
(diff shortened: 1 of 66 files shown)


astro-friedel:
After review, we have decided that the spdlog funcitonality will not be an optional component of ILLIXR. It is available as a OS provided package for both ubuntu (libspdlog-dev) and fedora (spdlog-devel), and is thus easily installable.

astro-friedel:
The concurrency issue will be addressed though.

e3m3:
> After review, we have decided that the spdlog funcitonality will not be an optional component of ILLIXR. It is available as a OS provided package for both ubuntu (libspdlog-dev) and fedora (spdlog-devel), and is thus easily installable.

See issue [discussion](https://github.com/ILLIXR/ILLIXR/issues/427#issuecomment-2439877818).

PR #432, opened 2024-12-23
Filament plugin

yavuz650:
Here's a plugin that uses the [Filament](https://github.com/google/filament) engine to render graphics in ILLIXR. While the plugin works on my end, there are a few things that need to be addressed before the plugin is fully integrated with the ILLIXR build system. I built and ran everything in a Docker container.

-  The CMake file in the plugin directory assumes the Filament library resides under `/opt/ILLIXR/filament`. If I understand the ILLIXR build system correctly, the Filament repository should be cloned and built as an external dependency and should be kept in `build/_deps`. This requires a CMake script to clone and build the Filament repository(`GetFilament.cmake`), similar to existing 3rd party libraries. I'm not familiar with the infrastructure so I'm hoping you guys can take care of that, if you'd like. 
- By default the plugin loads and renders the [Intel Sponza ](https://www.intel.com/content/www/us/en/developer/topic-technology/graphics-research/samples.html)scene, which is freely available. However, any .glTF asset should work, as Filament supports the .glTF format. I have a couple more scenes(Bistro and San Miguel) that I used with Filament that I can provide if you'd like. Or perhaps you could simply convert the ILLIXR demo scene to .glTF and use that. 
- Filament requires the Clang compiler for build. I have been building the ILLIXR project using Clang with no problems. I remember trying to build Filament with Clang, and the rest of the ILLIXR modules with gcc, but that would lead to problems(deadlock, program hanging.). That was two years ago, I'm not sure if that is still the case. 
- The upstream Filament repository does not build/link with the ILLIXR system out-of-the-box. I have my own clone of Filament [here](https://github.com/yavuz650/filament), which you can clone/host under the ILLIXR repo collection.

You'll probably want to merge this to a new branch and not `master`. We can keep working from there.

### DIFF
--- a/plugins/filament_gldemo/CMakeLists.txt
+++ b/plugins/filament_gldemo/CMakeLists.txt
@@ -1,0 +1,69 @@
+# module to build and install the filament_gldemo ILLIXR plugin
+
+set(PLUGIN_NAME plugin.filament_gldemo${ILLIXR_BUILD_SUFFIX})
+
+# source files, listed individually so that any changes will trigger a rebuild
+add_library(${PLUGIN_NAME} SHARED plugin.cpp
+        IBL.cpp
+        IBL.h
+        ../../include/illixr/shaders/demo_shader.hpp
+        ../../include/illixr/data_format.hpp
+        ../../include/illixr/error_util.hpp
+        ../../include/illixr/extended_window.hpp
+        ../../include/illixr/gl_util/obj.hpp
+        ../../include/illixr/global_module_defs.hpp
+        ../../include/illixr/math_util.hpp
+        ../../include/illixr/pose_prediction.hpp
+        ../../include/illixr/shader_util.hpp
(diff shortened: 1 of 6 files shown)


astro-friedel:
Do you know why the filament engine requires clang for the compiler?

yavuz650:
I'm not sure really, but there are checks in the CMake scripts that explicitly look for Clang, so I imagine filament would not be happy if we forced build with gcc, although I haven't really tried it.

astro-friedel:
I would prefer if we could get it to build with any compiler, as we do not specify a compiler for ILLIXR. I did some reading on this issue and the google team seem to be kind of jerks about it, essentially telling people who inquire about gcc compatibility, or even try to fix it to go away. I will try and take a look at this issue in the future, but I have a lot on my plate right now as we have a release coming out very soon.

PR #437, opened 2025-01-21
Offload rendering and VIO

qinjunj:
This PR introduces the new networking interface, multiple plugins related to offloading rendering, support for light house tracking, and updates to offloading VIO plugins that conform to the new network interface.


CMakeLists.txt
```diff
@@ -122,7 +141,8 @@ elseif(lower_type MATCHES "relwithdebinfo")
     set(ILLIXR_BUILD_SUFFIX ".optdbg")
 endif()
 
-set(CMAKE_PREFIX_PATH "${CMAKE_INSTALL_PREFIX}/lib/cmake ${CMAKE_INSTALL_PREFIX}/lib64/cmake")
+set(Vulkan_GLSLANG_VALIDATOR_EXECUTABLE "")
+set(CMAKE_PREFIX_PATH "${CMAKE_INSTALL_PREFIX}/lib/cmake ${CMAKE_INSTALL_PREFIX}/lib64/cmake /usr/lib/x86_64-linux-gnu/cmake /usr/lib/x86_64-linux-gnu/")
```
astro-friedel, line 145: It is best to not specify hard coded paths here. There is no guarantee that ever OS contains these specific paths. Also these particular ones should be included by CMake itself.

cmake/GetGTSAM.cmake
```diff
@@ -1,9 +1,10 @@
 # CMake module to look for GTSAM
 # if it is not found then it is downloaded and marked for compilation and install
 
-find_package(GTSAM 4.3.0 EXACT QUIET)
+find_package(GTSAM 4.3.0 EXACT)
```
astro-friedel, line 4: Why remove the QUIET here. If GTSAM is not found CMake will now issue a warning, which is not necessary since we build our own in the next few lines.

illixr_oc.yaml
astro-friedel, line 1: This should be more clearly named, e.g. to represent what it does

CMakeLists.txt
qinjunj: Yes you are right. There could be some leftover code that is specific to my system when I tried to build the code. Could you please correct them to what they should to be?

cmake/GetVulkan.cmake
```diff
@@ -0,0 +1,79 @@
...
+        elseif (Vulkan${ITEM})
+            unset(Vulkan${ITEM} CACHE)
+            unset(Vulkan${ITEM} PARENT_SCOPE)
+        endif()
+    endforeach()
+endfunction()
+
+if(NOT Vulkan_FOUND)
+    message(STATUS "Checking for module 'vulkan' > 1.3.255")
+    find_package(Vulkan REQUIRED)
```
astro-friedel, line 19: This should be
`find_package(Vulkan 1.3.255 QUIET)`
Having REQUIRED will cause it to fail immediately and never get to the `pkg_check_modules` version

cmake/GetVulkan.cmake
```diff
@@ -0,0 +1,79 @@
...
+        # message(STATUS "  Found Vulkan, version ${Vulkan_VERSION} ${Vulkan_LIBRARIES} ${Vulkan_INCLUDE_DIRS}")
+        # Since we now have both Vulkan and glslang found, they will use the system wide Vulkan
+        # set(Vulkan_DEP_STR Vulkan_ext)
+        # set(Vulkan_EXTERNAL Yes)
+        set(Vulkan_INCLUDE_DIRS ${CMAKE_INSTALL_PREFIX}/include)
+        # set(Vulkan_INCLUDE_DIRS ${CMAKE_BINARY_DIR}/_deps/vulkan/src/Vulkan_ext/Vulkan-Headers/include)
+        set(Vulkan_LIBRARIES ${CMAKE_INSTALL_PREFIX}/lib/libvulkan.so)
+        # set(Vulkan_LIBRARIES /usr/lib/x86_64-linux-gnu/libvulkan.so)
+        set(Vulkan_GLSLANG_VALIDATOR_EXECUTABLE ${CMAKE_INSTALL_PREFIX}/bin/glslangValidator)
+        # set(Vulkan_GLSLANG_VALIDATOR_EXECUTABLE /usr/bin/glslangValidator)
```
astro-friedel, line 54: These should not be here. If Vulkan is already found these should not be set, especially to the illixr install directory.

cmake/GetVulkan.cmake
```diff
@@ -0,0 +1,79 @@
...
+    endif()
+endif()
+
+set(Vulkan_GLSLANG_VALIDATOR_EXECUTABLE ${Vulkan_GLSLANG_VALIDATOR_EXECUTABLE} CACHE INTERNAL "" FORCE)
```
astro-friedel, line 79: why are we emptying this variable? I think this is the cause of   timewarp_vk not finding glslangValidator.

cmake/HelperFunctions.cmake
```diff
@@ -215,6 +215,16 @@ function(read_yaml FILENAME)
             list(GET ITEM_LIST 1 TEMP_BUILD_FLAGS)
             string(REPLACE "," " " LOC_BUILD_FLAGS "${TEMP_BUILD_FLAGS}")
             set(LOCAL_BUILD_FLAGS ${LOC_BUILD_FLAGS} PARENT_SCOPE)
+        elseif(KEY STREQUAL "vulkan")
+            list(GET ITEM_LIST 1 VULKAN_STR)
+            string(TOUPPER ${VULKAN_STR} VULKAN_UPPER)
+            if(VULKAN_UPPER STREQUAL "TRUE")
+                set(ILLIXR_VULKAN ON PARENT_SCOPE)
+                add_compile_definitions(ILLIXR_VULKAN)
+            endif()
```
astro-friedel, line 224: Why is this here? I see no profiles that specify 'vulkan' as an option.

include/illixr/network/network_backend.hpp
```diff
@@ -0,0 +1,43 @@
+#ifndef ILLIXR_NETWORK_BACKEND_HPP
+#define ILLIXR_NETWORK_BACKEND_HPP
```
astro-friedel, line 2: This should be
#pragma once

include/illixr/network/topic_config.hpp
```diff
@@ -0,0 +1,21 @@
+#ifndef ILLIXR_TOPIC_CONFIG_HPP
+#define ILLIXR_TOPIC_CONFIG_HPP
```
astro-friedel, line 2: This should be
#pragma once

cmake/HelperFunctions.cmake
Jebbly: This should be addressed now (with the latest commits)

plugins/lighthouse/README.md
```diff
@@ -0,0 +1,3 @@
+# Lighthouse
+## Summary
+The ``lighthouse`` plugin supports lighthouse tracking using the [libsurvive library](https://github.com/collabora/libsurvive). To use this plugin, libsurvive should first be built, installed, and calibrated according to [their documentation](https://github.com/collabora/libsurvive/blob/master/README.md).
```
astro-friedel, line 3: Don't think we need to say to build libsurvive as we do it automatically.

start.sh
astro-friedel, line 1: Not sure this file should be here. It seems very customized.

astro-friedel (commented):
When running the timewarp_vk.yaml I run into a few issues.
1. When it starts up, it resized my display to 2880x1600 pixels, when it ends it does not revert this back. The previous version of this (from master) does not resize my display. This issue needs to be addressed.

CMakeLists.txt
```diff
@@ -43,6 +43,9 @@ if(UNIX)
     else()
         set(HAVE_CENTOS No)
     endif()
+
+    find_package(ECM REQUIRED NO_MODULE)
```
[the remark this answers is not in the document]
Jebbly: ECM is [extra-cmake-modules](https://launchpad.net/ubuntu/+source/extra-cmake-modules), although I'm not exactly sure how they're being used either.

astro-friedel:
I have tested the following setups

- [ ] moado_openwarp_vk
- [ ] monado_timewarp_gl
- [ ] monado_timewarp_vk
- [ ] monado_vk_lookup
- [ ] native_gl_lookup
- [ ] native_gl_slam
- [ ] native_openwarp_vk
- [ ] native_timewarp_gl
- [ ] native_timewarp_vk
- [ ] native_vk_lighthouse
- [ ] native_vk_lookup
- [ ] native_vk_slam

qinjunj:
Tested the offloading setup, including offloading rendering, offloading head tracking, and offloading both together.

astro-friedel (approved):

rsingh76 (approved):
Based on Qinjun's and Doug's testing and to meet extreme deadlines.

PR #438, opened 2025-01-23
changed the way docker images are made

astro-friedel:
This changes the way we generate docker images. Now we only supply images and Dockerfiles for Ubuntu based builds. The root image is an NVIDIA/cuda image so that the container can utilize any NVIDIA gpus. We now will supply several images per build:

- full build
- build with no GPL licensed code
- build with no LGPL or GPL licensed code
- build without support for ZED
- build with no GPL licensed code or support for ZED
- build with no LGPL or GPL licensed code or support for ZED


shg8:
I think we need to provide instructions on GPU passthrough, or some of the plugins wouldn't function at all. [nvidia-container-toolkit](https://docs.docker.com/engine/containers/resource_constraints/#install-nvidia-container-toolkit) is required for NVIDIA GPUs, and I'm not sure what the requirements on for other vendors. Additionally, we should provide sample Docker Compose files that have the passthrough configured so that the process is easier.

astro-friedel:
There is a file, docs/docs/docker.md that explains what the user needs to do in order to use the container.(s)

astro-friedel:
Also do we actively support AMD cards?

shg8:
> There is a file, docs/docs/docker.md that explains what the user needs to do in order to use the container.(s)

Could we also add instructions for X11 forwarding from the old repo? A docker compose file might be easier for this.

PR #440, opened 2025-02-17
Refactor debug preprocessor

astro-friedel:
This PR changes the way ILLIXR shuts down due to an error and how general logging is handled. This PR removes all calls to `abort()` from ILLIXR code, replacing them with exceptions. This makes debugging easier as abort can be very abrupt. 

Additionally, calls to loggers enclosed in NDEBUG `ifdef` statements have been simplified to remove the `ifdef`. This allows for any logging level to be set, regardless of how a library was built (debug vs release).


include/illixr/record_logger.hpp
```diff
@@ -160,26 +158,24 @@ class record {
...
-#endif
     }
 
     record() = default;
 
     ~record() {
-#ifndef NDEBUG
         if (record_header_ && !data_use_indicator_.is_used()) {
             spdlog::get("illixr")->error("[record_logger] Record was deleted without being logged.");
-            abort();
```
Boyuan-Tian, line 175: Is there a need to throw another `std::runtime_error` here to ensure consistency?

astro-friedel: I don't think so because destructors are labeled as `no-throw` and shouldn't throw as a general rule. This particular error is not severe enough to break that mold.

Boyuan-Tian (approved):
Looks good to me.

astro-friedel:
converted back to draft as there are other related issues that should be addressed as well

PR #442, opened 2025-02-24
fix for broken openvins

astro-friedel:
This reverts openvins to a previous base


qinjunj (commented):
`c++: error: ENABLE_MONADO: linker input file not found: No such file or directory`
Same error as I got in the other PR #422

astro-friedel:
@qinjunj I think your checkout may not be clean. I just grepped the entire code base and ENABLE_MONADO does not appear anywhere, in fact now that I think about it that variable name only appears in the offload_rendering_VIO branch.

qinjunj (dismissed):
Tested with `native_gl` and `native_vk`. Worked fine.

Jebbly (approved):
LGTM, thanks!

qinjunj (approved):

PR #444, opened 2025-02-27
Fix thread join in display_vk plugin stop method

shg8:
Ensure proper thread cleanup by joining the main thread if it is joinable when stopping the display_vk plugin.


shg8:
I've been testing the CI pipeline and need the master branch to cleanly exit. Right now, we have a segfault on exit. Since this is a minor chang and I need this to unblock my work, I'm going ahead with the merge.

PR #445, opened 2025-02-27
Add automated CI compilation and testing pipeline

shg8:
This CI pipeline runs on a self-hosted runner in our lab. It compile and runs ILLIXR with the `native_vk` profile. An exit code of 0 means a successful run, and any other exit code would cause the job to fail.


astro-friedel (approved):
This looks good for now. Once we have the proper docker images we can just pull them rather than building them in place.

PR #456, opened 2025-05-21
Jetson Offload Rendering Client

Jebbly:
Introduces an offloading rendering client plugin for the Jetson, and some minor QOL fixes to the server-side encoder.


src/runtime_impl.cpp
```diff
@@ -60,9 +60,6 @@ class runtime_impl : public runtime {
             RAC_ERRNO_MSG("runtime_impl before creating the dynamic library");
             return dynamic_lib::create(so_path);
         });
-        for (auto& i : libraries_) {
```
qinjunj, line 63: `enable_monado` is defaulted to `false` now, and it will not work for the server side where the display is provided by monado.

Issue #465, opened 2025-10-07
Re-work the utils cmake file to only build what is actually needed


astro-friedel:
The current version of utils/CMakeLists.txt builds all objects, even if they are not needed. This should be fixed to only build the parts that are actually needed. This is primarily to avoid requiring dependencies that may not actually be needed.

PR #467, opened 2025-10-07
Fix utils component build

astro-friedel:
This addresses issue #465 so that only the necessary parts ofg the code are built


astro-friedel:
Based on incorrect branch

PR #474, opened 2026-04-29
Enable Windows build

### LINKED ISSUE #52: Windows support
- [ ] Windows support for offline IMU Camera
- [ ] Document Panacea plugin translation for Windows
- [ ] Windows support for OpenVINS
- [ ] Investigate what would be necessary for MacOS/iOS?

### LINKED ISSUE #463: Convert ExternalProject to Fetch content
Several recent coding efforts have forced the use of CMake's `FetchContent` (specifically, Windows), as `ExternalProject` was not finding already installed packages. This should be a rather straightforward but delicate change.

astro-friedel:
We have wanted to expand the operating systems ILLIXR works on. This PR will begin the process. These updates are to the general build system and some plugins. The changes include:

- Definitions not pre-defined (e.g., M_PI)
- Clock updates to use OS specific mechanisms
- Installation of dependencies
  - We have to build dependencies from source, as Windows installers (for those that have them) often do not include the headers, or do not give us the version selections we need.
  - We use vcpkg for this process
  - vcpkg works well with CMake, so it fits in with the existing system
  - vcpkg allows us to set dependencies on a per-plugin basis
  - The dependency builds happen at configuration time, and should happen only once
  - This does increase the initial configuration time to several hours at a minimum
- Rework vulkan initialization, as Windows does not allow our current style of struct initialization
- Update Gl code to work on Windows
- Update CMakeLists.txt files to work with Windows (proper library linking
- Update src files to use OS specific library loading and environment variable retrieval
- Removal of all symlinks, which led to some restructuring of the code
- Added OS detection to all plugin CMakeLists.txt files, defaulting to not building on Windows. Individual plugins will be enabled as they are updated.
- Updates to the following plugins to work under Windows
  - native_renderer
  - offline_cam
  - offline_imu
  - passthrough_integrator
  - rk4_integrator
  - timewarp_vk
  - fauxpose
  - pose_lookup
  - pose_prediction
  - vkdemo
  - tcp_network_backend


sriramdvt:
This PR does not finish the Windows integration, and the goal of this review is to try and make sure that the Linux side of ILLIXR does not break.

I tested the `monado_vk_lookup` profile and built ILLIXR on a Linux system. Here are the fixes that I needed to make for  commit https://github.com/ILLIXR/ILLIXR/pull/474/changes/94d9474db701e5168f47f0261ea6c555cb29b53f that I didn't have to do for the current main branch:

- There was a conflict between my Conda libraries and system libraries that CMake had an issue with. I had to add `-DCMAKE_PREFIX_PATH=$CONDA_PREFIX` to the build command. This might be a non-issue if you just disable Conda, but it might be nice to have this as a note somewhere, or perhaps have some FAQs.
- The linker kept throwing an error for the vulkan utils: `...recompile with -fPIC...`. Enabling position independent code (PIC) for vulkan by adding `set_target_properties(illixr_vulkan_utils PROPERTIES POSITION_INDEPENDENT_CODE ON)` right after the `add_library(illixr_vulkan_utils ...)`  line in `utils/CMakeLists.txt` fixed this. The latest commit https://github.com/ILLIXR/ILLIXR/pull/474/changes/c794524ac6cb5086f3f11a913d2f3fe9a10ac6c5 adds this.

sriramdvt:
Thanks for the updates. I was also able to successfully build all the modified plugins on a Linux system by choosing them from the [ILLIXR Getting Started](https://illixr.github.io/ILLIXR/getting_started/) page.

sriramdvt (approved):
This PR does not seem to break the Linux side of ILLIXR. Tested this PR by building the `monado_vk_lookup` profile, and the updated plugins individually.

PR #479, opened 2026-05-04
New UDP Based Network Plugin

astro-friedel:
This PR should not be reviewed until after [474](https://github.com/ILLIXR/ILLIXR/pull/474) is approved and merged.

This new plugin adds to ILLIXR's networking capabilities. The impetus behind this plugin is:

- On some operating systems/devices, small TCP packets can accumulate in the buffers before being released to be sent or made available for processing
- Some data are needed at high cadence (e.g., head poses), but can be clumped by TCP, rather than a steady flow
- The use of these high-cadence data can accept some packet loss, because the steady flow is more important

The user can choose which backend (TCP or UDP) to use for each networked topic. Both TCP and UDP can be used simultaneously, each with its own set of topics. The backend is selected as part of the network configuration, just as the serialization method is.


plugins/udp_network_backend/plugin.cpp
```diff
@@ -0,0 +1,227 @@
...
+    packet.append(reinterpret_cast<const char*>(&topic_name_length), 4);
+    packet.append("illixr_control");
+    packet.append(message);
+
+    if (!peer_socket_->write_data(packet))
+        spdlog::get("illixr")->warn("[udp_network_backend] send_control failed");
+}
+
+extern "C" MY_EXPORT_API plugin* this_plugin_factory(phonebook* pb) {
+    auto plugin_ptr = std::make_shared<udp_network_backend>("udp_network_backend", pb);
```
jianxiapyh, line 243: Blocking: the plugin factory creates incompatible ownership for the UDP backend.

This factory constructs the backend with std::make_shared and registers that shared ownership in the phonebook, but then returns plugin_ptr.get(). runtime_impl::load_so() wraps the returned raw pointer in independent ownership at src/runtime_impl.cpp:95.

On shutdown, the runtime calls delete on that raw pointer. Because the object was created inside a make_shared allocation, it is not a separately allocated pointer that can be deleted this way. The phonebook also still owns the same object.

I reproduced this using the normal built executable with only udp_network_backend enabled for one second. Shutdown failed with:
`double free or corruption (out)
exit code 139`

plugins/udp_network_backend/plugin.cpp
```diff
@@ -0,0 +1,227 @@
...
+    ready_       = true;
+    read_loop(socket);
+}
+
+void udp_network_backend::read_loop(network::UDPSocket* socket) {
+    std::string buffer;
+    while (running_) {
+        // Each recvfrom() returns exactly one datagram — no partial-read reassembly needed
+        sockaddr_in src_addr{};
+        std::string packet = socket->read_data(&src_addr);
```
jianxiapyh, line 112: Blocking: the UDP server never records the client as its peer, so server-to-client topics cannot work.

read_data(&src_addr) retrieves the sender’s address here, but src_addr is never passed to set_peer() or retained for write_data_to(). Unlike the client, start_server() only binds the socket, so its peer_set_ remains false.

This causes two failures:
topic_create() skips the control message because has_peer() is false.
topic_send() calls write_data(), which returns false when no peer has been set.

I reproduced this through the public switchboard::get_network_writer API:
`client_to_server=PASS
[udp_network_backend] write_data failed for topic=server_to_client
server_to_client=FAIL`

plugins/udp_network_backend/plugin.cpp
```diff
@@ -0,0 +1,227 @@
...
+        // Packet format: total_length(4) | topic_name_length(4) | topic_name | message
+        if (packet.size() < 8) {
+            spdlog::get("illixr")->warn("[udp_network_backend] Undersized datagram ({} bytes), dropping", packet.size());
+            continue;
+        }
+
+        uint32_t total_length      = *reinterpret_cast<const uint32_t*>(packet.data());
+        uint32_t topic_name_length = *reinterpret_cast<const uint32_t*>(packet.data() + 4);
+
+        if (packet.size() < total_length || total_length < 8 + topic_name_length) {
```
jianxiapyh, line 133: topic_name_length comes directly from the network. A sufficiently large value can overflow 8 + topic_name_length, bypassing this validation.

plugins/udp_network_backend/plugin.cpp
```diff
@@ -0,0 +1,227 @@
...
+        spdlog::get("illixr")->warn("[udp_network_backend] write_data failed for topic={}", topic_name);
+}
+
+// Helper function to queue a received message into the corresponding topic
+void udp_network_backend::topic_receive(const std::string& topic_name, std::vector<char>& message) {
+    if (topic_name == "illixr_control") {
+        std::string message_str(message.begin(), message.end());
+        if (message_str.find("create_topic") == 0) {
+            size_t d_pos = message_str.find(delimiter_);
+            assert(d_pos != std::string::npos);
```
jianxiapyh, line 196: A validly framed illixr_control datagram whose payload begins with create_topic but does not contain | causes this assertion to abort the entire process.

jianxiapyh (changes requested):

plugins/udp_network_backend/plugin.cpp
astro-friedel: What OS and compiler are you using? This code is identical to what is in tcp_network_backend, and has been like that since it was created. I do not get these errors on any of the OS's I have tried it on: Ubuntu 24, Android, or Windows 11.

plugins/udp_network_backend/plugin.cpp
astro-friedel: I discovered this while testing before I went on vacation. I will have a fix in shortly.

plugins/udp_network_backend/plugin.cpp
astro-friedel: Given that topic_name_length can have a value of over 4.29 billion, I think this is a very very unlikely corner case. If you make the topic name 4+ billion characters long, you are aiming to break things anyway.

plugins/udp_network_backend/plugin.cpp
astro-friedel: Currently, that is the point. The assert is only checked in debug mode, so detecting an invalid control message (all control messages must have the | by definition) and aborting is the expected behavior. This code is identical to tcp_network_backend.

plugins/udp_network_backend/plugin.cpp
astro-friedel: I just pushed the fix for this.

plugins/udp_network_backend/plugin.cpp
jianxiapyh: I tested on Ubuntu 24.04.4 LTS x86-64 using Clang 18.1.3, CMake 3.28.3

astro-friedel: Ok, I see the issue and agree it needs to be addressed, but I think we should leave the code as is for now because:

- This is the same way the tcp_network_backend has worked since it was created, and I do not want to wrap that plugin into this PR.
- We are currently working to implement multi-user networking, and this would fit nicely into that work.

jianxiapyh (approved):

PR #480, opened 2026-05-04
Refactor poses and serialization

astro-friedel:
This PR should not be reviewed until after [479](https://github.com/ILLIXR/ILLIXR/pull/479/changes) is approved and merged.

This PR addresses the growing complexity of the pose data structures used by ILLIXR. With the addition of hand tracking and hand interactions, there are numerous pose-based structs. This work breaks up the pose header into several, more specific headers. Similarly, the serialization code was also broken into more specific files. This change allows a plugin to use specific types of poses without the need to bring in the entire pose framework. The poses are now more hierarchical, with a common base struct.

Since ILLIXR is now starting to work more closely with OpenXR programs, the pose structs now have compile-time defines to make the structs OpenXR-compliant or the native ILLIXR format.

The head_pose_type has also been updated to include velocity information, following the OpenXR standard. This allows for pose prediction based on a single pose to be done.

PR #481, opened 2026-05-04
New Latency Tracking Plugin

astro-friedel:
This PR should not be reviewed until after [480](https://github.com/ILLIXR/ILLIXR/pull/480) is approved and merged.

This PR introduces a set of plugins to track the network latency between a client and server. It works by sending a small packet from the client to the server; the server returns the packet after adding some timing information. The client then calculates the round-trip latency based on the time it took the packet to return.

This plugin can measure the one-way latency (in both directions), but only if both client and server are running the same operating system and have well-synchronized clocks.


sriramdvt (commented):
I am able to build the `network_latency.rx` and `network_latency.tx` plugins. I'm still going through the code. Is there a way to test this plugin?

astro-friedel:
Yes, you will need to have 2 illixr instances running.
Setup 1: tcp_network_backend,network_latency.tx
Setup 2: tcp_network_backend,network_latency.rx

plugins/network_latency/tx/pong_rx/plugin.cpp
```diff
@@ -0,0 +1,111 @@
...
+
+    spdlog::get("illixr")->warn("Rx time: {}ns", receive_time);
+
+    spdlog::get("illixr")->warn("  Client time: {}ns", pong->client_timestamp_ns);
+    spdlog::get("illixr")->warn("  RTT: {:.3f}ms", rtt_ms);
+    spdlog::get("illixr")->warn("  Latency: {:.3f}ms", c2s_ms);
+    spdlog::get("illixr")->warn("  Server time: {}ns", pong->server_timestamp_ns);
+    spdlog::get("illixr")->warn("  Clock offset: {:.3f}ms", clock_offset_ms);
+
+    // spdlog::get("illixr")->warn("Rx time: {}ns  Client time: {}ns  RTT: {.3f}ms  Latency: {.3f}ms  Server time: {}ns  Clock
```
sriramdvt, line 60: Nit: remove these unnecessary logs.

sriramdvt: Lines 60-67.

plugins/network_latency/tx/pong_rx/plugin.cpp
```diff
@@ -0,0 +1,111 @@
...
+    // Reject samples where RTT is implausibly large (> 500ms on a local
+    // network indicates a clock jump or network anomaly)
+    if (rtt_ms > 500.0 || rtt_ms < 0.0) {
+        spdlog::get("illixr")->warn("[network_latency_pong_rx] Rejecting implausible RTT={:.3f}ms", rtt_ms);
+        return;
+    }
+
+    // Reject offset samples that differ from the current smoothed value
+    // by more than 10ms, which indicates a clock adjustment mid-session
+    if (clock_offset_initialized_ && std::abs(clock_offset_ms - smoothed_clock_offset_ms_) > 10.0) {
```
sriramdvt, line 78: I don't think we should do this, especially since we have the smoothed moving average anyway. Perhaps increase this threshold to 50ms to decrease the likelihood of this rejection happening?

plugins/network_latency/tx/pong_rx/plugin.cpp
```diff
@@ -0,0 +1,111 @@
...
+        spdlog::get("illixr")->warn("[network_latency_pong_rx] Rejecting outlier offset={:.3f}ms "
+                                    "(smoothed={:.3f}ms)",
+                                    clock_offset_ms, smoothed_clock_offset_ms_.load());
+        return;
+    }
+
+    // Maintain a rolling average of clock offset over recent samples to
+    // smooth out network jitter. Weight recent samples more heavily.
+    constexpr double alpha = 0.1; // EMA smoothing factor
```
sriramdvt, line 87: Have the full version of EMA (exponential moving average) at least once somewhere in the network_latency plugins. Perhaps here is a good place. Or in `latency_data.hpp`.

sriramdvt (approved):
Am able to build these plugins and tested them too. Have some minor comments, but otherwise approve this PR.

Notes when trying to test the PR on a single machine ->
Build the plugins `network_latency.rx,network_latency.tx,tcp_network_backend`. Set these environment variables for the `tcp_network_backend` plugin (assuming that ports 9001 and 9002 are free) and to disable any display:
```bash
export ILLIXR_TCP_CLIENT_IP=127.0.0.1
export ILLIXR_TCP_SERVER_IP=127.0.0.1
export ILLIXR_TCP_CLIENT_PORT=9001
export ILLIXR_TCP_SERVER_PORT=9002
export ILLIXR_DISPLAY_MODE=none
```

On two terminals, set the environment variable `ILLIXR_IS_CLIENT` to `0/1`:
```bash
export ILLIXR_IS_CLIENT=0 # on terminal 1
export ILLIXR_IS_CLIENT=1 # on terminal 2
```

Run the `rx/tx` plugins on the two terminals:
```bash
main.<>.exe -p tcp_network_backend,network_latency.rx ... # terminal 1
main.<>.exe -p tcp_network_backend,network_latency.tx ... # terminal 2
```

plugins/network_latency/tx/pong_rx/plugin.cpp
astro-friedel: I will leave these for now as they are good for debugging, especially when dealing with two different operating systems

plugins/network_latency/tx/pong_rx/plugin.cpp
astro-friedel: This can be discussed, but for now 10ms is ok.

plugins/network_latency/tx/pong_rx/plugin.cpp
astro-friedel: This could be implemented in the future, after discussion.

sriramdvt: Oh I meant having the abbreviation written in the comments, but okay no worries.

PR #483, opened 2026-05-22
Offload rendering server nvenc

astro-friedel:
This PR should not be reviewed until after https://github.com/ILLIXR/ILLIXR/pull/481 is approved and merged.

NVIDIA encoder version of this plugin

PR #484, opened 2026-05-22
Removing Symlinks

astro-friedel:
This PR should not be reviewed until after https://github.com/ILLIXR/ILLIXR/pull/483 is approved and merged.

This PR refactors ILLIXR to remove all symlinks. This is because Windows does not support them and will break its builds. Specifically:

Headers have been moved around so that any simlinks are no longer necessary
Symlinks in documentation now rely on CMake making copies of the files at configure time
README.md file symliks have been removed from plugin directories


include/illixr/data_format/poses/palm_pose.hpp
```diff
@@ -40,7 +40,7 @@ struct palm_pose : xrt_space_relation {
  * XR_EXT_palm_pose or the PALM joint of XR_EXT_hand_tracking.
  */
 struct palm_poses_pair : public switchboard::event {
-    std::map<hand, palm_pose> hands;       //!< Per-hand palm pose keyed by @c hand
+    std::map<side, palm_pose> hands;       //!< Per-hand palm pose keyed by @c hand
```
qinjunj, line 43: Just a note: this change seems to be out of the scope of this PR. 

What does `side` mean?

include/illixr/data_format/serialization/combined_pose.hpp
```diff
@@ -11,13 +11,13 @@ namespace boost::serialization {
 #ifdef USING_OPENXR
 
 template<class Archive>
-void serialize(Archive& ar, ILLIXR::data_format::pose::combined_pose& data, const unsigned int version) {
-    (void) version;
+void serialize(Archive& ar, ILLIXR::data_format::pose::combined_pose& data, const unsigned int) {
```
qinjunj, line 14: Is the second argument still needed?

include/illixr/imgui/backends/imgui_impl_allegro5.h
```diff
@@ -0,0 +1,40 @@
+// dear imgui: Renderer + Platform Backend for Allegro 5
```
qinjunj, line 1: What is `Allegro` used for?

include/illixr/global_module_defs.hpp
```diff
@@ -16,10 +16,10 @@ namespace ILLIXR {
 /// Display parameters
 struct display_params {
     // Display width in pixels
-    static constexpr unsigned width_pixels = 2880;
+    static constexpr unsigned width_pixels = 1680;
```
qinjunj, line 19: Why are the changes?

qinjunj (commented):
Can you explain the purpose and usage of the entire `include/illixr/imgui/`? And why would it be part of this PR? 

Same question applies to many other changes in this PR. Please make sure that the PRs are rebased correctly so we are not applying changes repeatedly.

include/illixr/data_format/serialization/combined_pose.hpp
astro-friedel: It is a required argument for all serialize functions. The API is set by Boost, not ILLIXR.

include/illixr/data_format/poses/palm_pose.hpp
astro-friedel: It is technically not part of this PR. The PR's in this series each derive from master, so each one contains the changes from all the previous ones, which makes reviewing them out of sequence difficult.

include/illixr/global_module_defs.hpp
astro-friedel: They are changed to match the Quest 3.

include/illixr/imgui/backends/imgui_impl_allegro5.h
astro-friedel: As far as I know it is not. I just moved the imgui code we had to another location to remove the symlink, I did not add or remove any of the files.

astro-friedel:
We use imgui in the debugview plugin to display the data. It is part of this PR because it originally lived in utils, with a symlink in include (include/imgui -> utils/imgui). Since symlinks needed to be removed, this necessitated moving this code.

astro-friedel:

PR #485, opened 2026-05-22
Android base system

astro-friedel:
This PR should not be reviewed until after https://github.com/ILLIXR/ILLIXR/pull/484 is approved and merged.

This work merges the Android-based repository into the main ILLIXR repo. Specific changes include:

- Android-specific headers
- Updates to headers with ifdef logic to enable Android features
- Android build files (Gradle, Java, etc.)
- Updates to src for Android builds


src/main.cpp
```diff
@@ -39,7 +46,45 @@ static void sigint_handler([[maybe_unused]] int sig) {
...
 int main(int argc, const char* argv[]) {
+#    endif
+
+#    ifdef __ANDROID__
+    if (cmd == APP_CMD_INIT_WINDOW) {
+        const std::vector<std::string> plugins = {"tcp_network_backend", "udp_network_backend", "network_latency.tx",
+                                                  "offload_rendering_client", "openxr_interface"};
+
+        // EuRoC
+        setenv("ILLIXR_DATA", "/sdcard/Android/data/com.example.native_activity/mav0", true);
```
qinjunj, line 73: all hardcoded?

src/main.cpp
```diff
@@ -59,15 +104,46 @@ int main(int argc, const char* argv[]) {
...
-#ifndef NDEBUG
-    /// When debugging, register the SIGILL and SIGABRT handlers for capturing more info
-    std::signal(SIGILL, sigill_handler);
-    std::signal(SIGABRT, sigabrt_handler);
-#endif /// NDEBUG
+        /// Shutting down method 1: Ctrl+C
+        std::signal(SIGINT, sigint_handler);
+#    ifdef __ANDROID__
+        std::thread runtime_thread(ILLIXR::run, plugins, app);
+        runtime_thread.join();
```
qinjunj: `handle_cmd()` starts `ILLIXR::run(...)` on a new thread, but then immediately calls `runtime_thread.join()`. Since this callback runs from the NativeActivity event loop via `source->process(state)`, the join blocks the Android looper until the runtime exits. With `ILLIXR_RUN_DURATION` set to 1000000, lifecycle events like pause, term-window, destroy, and other app commands may not be processed for a very long time.

Is this by design?

src/main.cpp
```diff
@@ -59,15 +104,46 @@ int main(int argc, const char* argv[]) {
...
+        int                         ident;
+        int                         events;
+        struct android_poll_source* source;
 
-    return ILLIXR::run(result);
+        // If not animating, we will block forever waiting for events.
+        // If animating, we loop until all events are read, then continue
+        // to draw the next frame of animation.
+        do {
+            ident = ALooper_pollOnce(0, nullptr, &events, (void**) &source);
```
qinjunj: Consider `ALooper_pollOnce(-1, nullptr, &events, (void**) &source);` to avoid spinning

app/build.gradle
```diff
@@ -0,0 +1,81 @@
...
+
+def root_path = rootProject.getLayout().getProjectDirectory().getAsFile().getPath()
+def root_library_path = "${root_path}/android_libraries/third_party/"
+
+def home = System.getenv('HOME') ?: System.getenv('USERPROFILE')
+
+android {
+    signingConfigs {
+        debug {
+            storeFile file("${home}/illixr.keystore")
```
qinjunj, line 13: Are we making the release key and password public?

plugins/network_latency/rx/plugin.cpp
```diff
@@ -0,0 +1,71 @@
...
+threadloop::skip_option network_latency_rx::_p_should_skip() {
+    // Always run to check for incoming pings with minimal latency
+    // The threadloop will yield appropriately
+    return skip_option::run;
+}
+
+void network_latency_rx::_p_one_iteration() {
+    // Process all available pings from the buffered reader
+    while (true) {
+        auto ping = ping_reader_.dequeue();
```
qinjunj, line 35: Why using `buffered_reader` for the `ping_reader_`? Do we need to process all the old requests?

src/main.cpp
astro-friedel: Yes. On Android systems there is little choice for where files can go.

app/build.gradle
astro-friedel: We do not need to. Each person who uses the code will make their own key/password. There are instructions in the docs on how to do this.

src/main.cpp
qinjunj: Will others be able to directly use the code with hardcoded paths?

plugins/network_latency/rx/plugin.cpp
astro-friedel: We should, as it can also let us know about dropped packets and other diagnostics.

qinjunj: Okay. Just to note that dequeue() on buffered_reader is a blocking call. So if there is no ping available, this loop won't exit cleanly.

src/main.cpp
astro-friedel: this has been fixed

src/main.cpp
astro-friedel: this has been fixed

src/main.cpp
astro-friedel: There should be instructions for this. It looks like those got lost during all of my work. I will work on updating the documentation tomorrow.

plugins/network_latency/rx/plugin.cpp
astro-friedel: That is fine.

astro-friedel:

PR #486, opened 2026-05-22
Android services updates

astro-friedel:
This PR should not be reviewed until after https://github.com/ILLIXR/ILLIXR/pull/485 is approved and merged.

This work merges updates for the services plugins for the Android OS. Specific changes include:

- New common_lock and extended_window plugins (Android only)
- Changes to the remaining services plugins


include/illixr/switchboard.hpp
```diff
@@ -781,31 +792,39 @@ class MY_EXPORT_API switchboard : public phonebook::service {
...
     }
 
     [[maybe_unused]] unsigned long get_env_ulong(const std::string& var, const unsigned long _default = 0) {
-        std::string val = get_env(var, "");
-        if (val.empty())
-            return _default;
-        return std::stoul(val);
+        std::string val = get_env(var, std::to_string(_default));
+        try {
+            int res = std::stoul(val);
```
qinjunj: `int` will truncate the value

src/main.cpp
```diff
@@ -39,7 +46,45 @@ static void sigint_handler([[maybe_unused]] int sig) {
...
+        setenv("ILLIXR_ENABLE_PRE_SLEEP", "False", true);
+        setenv("ILLIXR_TCP_CLIENT_IP", "192.168.8.140", true);
+        setenv("ILLIXR_TCP_SERVER_IP", "192.168.8.158", true);
+        setenv("ILLIXR_TCP_CLIENT_PORT", "9000", true);
+        setenv("ILLIXR_UDP_CLIENT_PORT", "9002", true);
+        setenv("ILLIXR_TCP_SERVER_PORT", "9001", true);
+        setenv("ILLIXR_UDP_SERVER_PORT", "9003", true);
+        setenv("ILLIXR_IS_CLIENT", "1", true);
+        setenv("ILLIXR_USE_DEPTH_IMAGES", "0", true);
+        setenv("ILLIXR_USE_MOTION_VECTORS", "0", true);
```
qinjunj: What are motion vectors used for?
How is this flag different from `ILLIXR_USE_MOTION_VECTOR_IMAGES`?

include/illixr/switchboard.hpp
astro-friedel: fixed

src/main.cpp
astro-friedel: They are a Quest 3 specific data type that is used by its onboard spacewarp. The env var was incorrectly named, I have fixed it.

astro-friedel:

PR #487, opened 2026-05-22
Plugins not supported by android

astro-friedel:
This PR should not be reviewed until after https://github.com/ILLIXR/ILLIXR/pull/486 is approved and merged.

This work updates CMakeLists.txt files for plugins that do not currently work on Android. This consists of warning messages and the automatic disabling of these plugins when building for Android.


astro-friedel:

PR #488, opened 2026-05-22
Android data plugin

astro-friedel:
This PR should not be reviewed until after https://github.com/ILLIXR/ILLIXR/pull/487 is approved and merged.

This work introduces the new android_data plugin. This is an update to the original android_imu_cam plugin by Bhowmik.

This plugin captures camera images and imu data and submits them to the switchboard for other plugins to utilize.


astro-friedel:

PR #490, opened 2026-05-22
Openxr interface plugin

astro-friedel:
This PR should not be reviewed until after https://github.com/ILLIXR/ILLIXR/pull/489 is approved and merged.

This Android-specific plugin acts as a relay between the native OpenXR Android runtime and ILLIXR. It is intended for offloading. The code consists of three main parts:

1. oxr_relay - A plugin that gets the current head and hand poses and hand interaction data from the device and sends them over the network to the server. This plugin loops at ~120 Hz (this can be changed) to retrieve the most recent data and uses UDP for network communication.
2. stereo_renderer - Gets the decoded frames from the offload_rendering_client plugin and renders them to the Android hardware buffers, so they can be placed in the OpenXR swapchain. The work is done entirely on the GPU.
3. openxr_interface - The main plugin that controls all data flow. It sets up the OpenXR session and launches the oxr_relay plugin. It then loops, taking images from the stereo_renderer and submitting them to the swapchain.

PR #491, opened 2026-05-22
Android offline data plugins

astro-friedel:
This PR should not be reviewed until after https://github.com/ILLIXR/ILLIXR/pull/490 is approved and merged.

This work updates the offline_can and offline_imu plugins to work on Android devices.


astro-friedel:

PR #492, opened 2026-05-22
Android network plugins

astro-friedel:
This PR should not be reviewed until after https://github.com/ILLIXR/ILLIXR/pull/491 is approved and merged.

This work updates the tcp_network_backend and udp_network_backend plugins to work on Adnroid devices.


astro-friedel:

PR #496, opened 2026-08-20
Semantic python plugin

astro-friedel:
This PR provides a plugin that creates a bridge between the ILLIXR runtime and a Python process. This allows the Python code to interact with the switchboard to read and write topics.


cmake/GetSemanticXR.cmake
rsingh76, line 1: I am not sure if this file is ever called/included in the semantic_python/Cmakelists

plugins/semantic_python/CMakeLists.txt
```diff
@@ -0,0 +1,85 @@
...
+endif()
+set(CMAKE_CUDA_ARCHITECTURES native)
+find_package(pybind11 CONFIG REQUIRED)
+
+# ---- NVDEC / CUDA ----
+# NVDEC_SDK_DIR must point to the NVIDIA Video Codec SDK root, e.g.:
+#   -DNVDEC_SDK_DIR="C:/dev/NvidiaVideoCodecSDK"
+# Available from https://developer.nvidia.com/nvidia-video-codec-sdk
+# Supports SDK 12.x and 13.x.
+if(NOT DEFINED NVDEC_SDK_DIR)
```
rsingh76, line 17: can we not have a fallback for machines that don't have NVDEC? Could fallback be a software decode?

docs/docs/plugin_README/README_semantic_python.md
```diff
@@ -0,0 +1,201 @@
...
+loads the [`switchboard`][G10] interface functions into the interpreter, and launches the Python script. On teardown, it
+shuts down the interpreter. 
+
+!!! note
+
+    Due to the way the interpreter is shut down, internal Python cleanup (e.g. `finally:` blocks) are not guaranteed to be called.
+
+## SemanticXR
+
+This plugin was developed in conjunction with the [SemnticXR][L01] python repository that provides a real-time semantic
```
rsingh76: Typo in SemnticXR

docs/docs/plugin_README/README_semantic_python.md
```diff
@@ -0,0 +1,201 @@
...
+SLAM system with object detection, segmentation, and CLIP-based understanding for interactive 3D scene mapping and 
+querying. This code uses the `sematic_python` plugin to integrate it into ILLIXR. When enabling this plugin, the
```
rsingh76: Typo -- sematic_python

plugins/semantic_python/CMakeLists.txt
```diff
@@ -0,0 +1,85 @@
...
+        "Example: -DNVDEC_SDK_DIR=\"C:/dev/NvidiaVideoCodecSDK\"")
+endif()
+
+enable_language(CUDA)
+find_package(CUDAToolkit REQUIRED)
+
+# Detect AV1 decode support from the SDK headers
+include(CheckCXXSymbolExists)
+set(CMAKE_REQUIRED_INCLUDES "${NVDEC_SDK_DIR}/Interface")
+check_cxx_symbol_exists(cudaVideoCodec_AV1 "cuviddec.h" NVDEC_HAS_AV1)
```
rsingh76, line 29: Is NVDEC_HAS_AV1 used somewhere? can't seem to find the enum.

docs/docs/modules.json
```diff
@@ -498,6 +498,13 @@
         "windows"
       ]
     },
+    {
+      "name": "semantic_python",
+      "cmake_flag": "USE_SEMANTIC_PYTHON",
+      "dependencies": [
+        "pybind11"
+      ]
```
rsingh76, line 509: No mention of OS here unlike other plugins, is that ok?

docs/docs/modules.json
```diff
@@ -498,6 +498,13 @@
         "windows"
       ]
     },
+    {
+      "name": "semantic_python",
+      "cmake_flag": "USE_SEMANTIC_PYTHON",
```
rsingh76: should this have -- "=ON"

cmake/GetSemanticXR.cmake
```diff
@@ -0,0 +1,7 @@
+set(SEMANTIC_XR_DIR "${CMAKE_SOURCE_DIR}")
+fetch_git(NAME SemanticXR
+          REPO https://github.com/ILLIXR/SemanticXR.git
+          TAG c959b9fce15d337b16cd7680b74c520446f7a3a0
+          OVERRIDE_BUILD ""
+          SRC_DIR ${SEMANTIC_XR_DIR}/python
+)
```
rsingh76, line 7: I am unsure, but should this file also contain "FetchContent_MakeAvailable(SemanticXR)"

rsingh76:
In my case, when I try to build I needed to change yaml-cpp::yaml-cpp -> yaml-cpp in 
https://github.com/ILLIXR/ILLIXR/blob/e1d9b50c793e904c7b0c95049f58b36639beb6f5/src/CMakeLists.txt#L68

plugins/semantic_python/switchboard_bindings.hpp
```diff
@@ -0,0 +1,235 @@
...
+        auto val = reader->get_ro_nullable();
+        if (!val)
+            return pybind11::none();
+
+        pybind11::dict data;
+        // ---- Decoded RGB ----
+        // Decoded on arrival by on_semantic_data(). Look up by frame_number.
+        // Falls back to latest if the exact frame was evicted from the cache.
+        const decode::decoded_frame_cache::Entry* entry = cache ? cache->find(val->frame_number) : nullptr;
+        if (entry == nullptr && cache != nullptr)
```
rsingh76: Could you sanity-check this fallback? The comment mentions eviction; I think a miss can also happen while decode is still catching up, in which case the entry returned would belong to a different frame than the pose and depth next to it.

plugins/semantic_python/plugin.cpp
```diff
@@ -0,0 +1,223 @@
...
+    }
+
+    if (!decoder_)
+        return;
+
+    try {
+        const uint8_t* rgb = decoder_->decode(frame->image.data(), frame->image.size());
+
+        if (rgb != nullptr) {
+            decoded_frames_.store(frame->frame_number, decoder_->width(), decoder_->height(), rgb);
```
rsingh76: The decoder here is constructed with bLowLatency=false. In the SDK source this target compiles in (${NVDEC_SDK_DIR}/Samples/NvCodec/NvDecoder/NvDecoder.cpp — grep ulMaxDisplayDelay, line 1068 in 13.1.15) that sets the parser's display delay to 1, so Decode() doesn't hand back the picture just submitted; it hands back an earlier one. That would make the RGB stored here belong to a previous packet, while frame->frame_number names the packet just fed. Could you double-check the two line up? Decode() takes an nTimestamp that GetFrame(&ts) returns, which would make the pairing explicit.

This will also affect the switchboard_bindings.hpp where we declare get(). Please sanity check this. 

Similar bugs had existed right before the IMMERSE symposium, although on the quest side, I think. But the core issue was the same, mismatch between the frame and pose/metadata

cmake/GetSemanticXR.cmake
astro-friedel: fixed

plugins/semantic_python/CMakeLists.txt
astro-friedel: We could, but it will take me some time to write given the huge amount of work I have to finish this week.

docs/docs/plugin_README/README_semantic_python.md
astro-friedel: fixed

plugins/semantic_python/CMakeLists.txt
astro-friedel: It is farther down near line 80. It is not an enum but CMake variable.

docs/docs/modules.json
astro-friedel: Fixed. I tagged it as Linux-only for now, since I haven't had a chance to test it on Windows.

docs/docs/modules.json
astro-friedel: fixed

cmake/GetSemanticXR.cmake
astro-friedel: It is not needed in this instance. FetchContentMakeAvailable is intended for compiled code. It runs CMake on the given target, but since the SemanticXR repo is Python-only and does not need a CMakeLists.txt, there is no need to run it.

docs/docs/plugin_README/README_semantic_python.md
astro-friedel: fixed

plugins/semantic_python/CMakeLists.txt
astro-friedel: I looked into this and we have a few options:
1. FFMPEG - will be a decent amount of work, may not work on Windows easily, does work with HEVC and AV1 formats
2. libde265 - HEVC only, easily installed as a system package on Ubuntu

plugins/semantic_python/switchboard_bindings.hpp
astro-friedel: I updated the code, and it should now ensure that the correct pose is attached to each image

plugins/semantic_python/plugin.cpp
astro-friedel: My fixes above will address this as well.

plugins/semantic_python/CMakeLists.txt
rsingh76: We should do this in the future. I assume neither can be done within an hour? libde265?

cmake/GetSemanticXR.cmake
rsingh76: So I don't think that the  SemanticXR repo is actually fetched at this point. Adding the above line fetched it for me I think. Please double-check on your end if the repo is actually fetched.

In addition - I willl change the TAG, for now I am on 25ae9c7af6461e774e4fd0ae30215cbc059c173b, but thhat will likely change.

plugins/semantic_python/decoder_cache.hpp
```diff
@@ -72,6 +75,62 @@ class decoded_frame_cache {
...
+        Entry&                      slot = entries_[next_slot_];
+        slot.data                        = frame;
+        slot.frame_number                = frame_number;
+        slot.valid                       = true;
+        next_slot_                       = (next_slot_ + 1) % CAPACITY;
+    }
+
+    // Find a frame's metadata by frame_number.
+    // Returns nullptr if not found (evicted or not yet arrived).
+    const Entry* find(int32_t frame_number) const {
```
rsingh76: latest() (line 65) and both find()s return a pointer into entries_ and the lock_guard releases as they return, so get() copies entry->rgb (switchboard_bindings.hpp:105) and meta->data (:142) with no lock held while store() at line 47/117 may be rewriting that slot. Could you double-check whether returning a copy taken under the lock would be safer — and is decoded_frame_cache::find() still used anywhere?

docs/docs/plugin_README/README_semantic_python.md
```diff
@@ -0,0 +1,201 @@
...
+    float depth_pose[16];
+    float max_depth;
+};
+```
+
+becomes (assuming the `struct` instance id named `mydata`)
+
+```
+{
+    "image":              = numpy 1D array of mydata.image
```
rsingh76: Double-check these descriptions; I think some of these are outdated.

plugins/semantic_python/CMakeLists.txt
astro-friedel: libde265 could probably be done within an hour, but I have a lot to do right now, so part of a future update would be better, where we could make a more robust solution.

docs/docs/plugin_README/README_semantic_python.md
astro-friedel: They are. I have updated them.

plugins/semantic_python/decoder_cache.hpp
astro-friedel: This has been fixed

utils/CMakeLists.txt
```diff
@@ -75,99 +79,100 @@ if(USE_DEBUGVIEW)
...
+            target_link_libraries(video_decoder_ada PUBLIC
+                                  ${OpenCV_LIBS}
+                                  ${LINK_LIBS}
+            )
+            target_compile_definitions(video_decoder_ada PUBLIC ADA)
+        endif()
+
+        if(USE_OFFLOAD_VIO.DEVICE_TX)
+            add_library(video_encoder_vio OBJECT
+                        video_encoder.hpp
```
rsingh76: I don't see a video_encoder.hpp, there is cpp here. hpp are in the include/illixr - so you probably need to fix path for line 142 and 143

CMakeLists.txt
rsingh76: Had to do this after the rebase, when I was building the client libraries -- 
if(BUILD_ANDROID)
    message(STATUS "Android build enabled")
elseif(WIN32 OR MSVC)

CMakeLists.txt
```diff
@@ -169,7 +183,33 @@ if(YAML_FILE)
...
+    set(USE_OFFLINE_CAM OFF)
+    set(USE_OFFLINE_IMU OFF)
+    set(USE_OPENVINS OFF)
+    set(USE_ANDROID_DATA OFF)
+    set(USE_GLDEMO OFF)
+    set(USE_GTSAM_INTEGRATOR OFF)
+    set(USE_RK4_INTEGRATOR OFF)
+    set(USE_TIMEWARP_GL OFF)
+    set(USE_TCP_NETWORK_BACKEND ON)
+    set(USE_UDP_NETWORK_BACKEND ON)
```
rsingh76: Had to make these changes to compile the ILLIXR libraries for Quest:
    set(USE_UDP_NETWORK_BACKEND OFF)
    set(USE_OPENXR_INTERFACE OFF)
    set(USE_OFFLOAD_RENDERING_CLIENT OFF)
    set(USE_NETWORK_LATENCY.TX OFF)
    set(USE_QUEST3.UNITY ON)

Do you hardcode things for  offload render? Perhaps some kind of gating is needed?

utils/serialization/latency.cpp
rsingh76, line 1: This file is refernced in the plugin - network_latency/tx/CMakeLists - please remove that reference

CMakeLists.txt
```diff
@@ -57,18 +57,19 @@ if(WIN32 OR MSVC)
     endif()
     set(VCPKG_OVERLAY_PORTS "${CMAKE_CURRENT_SOURCE_DIR}/cmake/ports")
     add_compile_definitions("-DWIN_OS")
```
rsingh76: Doesnt cmake automatically add "-D". I think this used to be just WIN_OS?

CMakeLists.txt
astro-friedel: Ah, yes I mis-merged this. It is now fixed.

CMakeLists.txt
astro-friedel: Currently, we need to change the variables here manually, as there is no way to read the profile files for Android yet. I have been working on a fix for this, but it will not make it into this release. There is documentation on how to select the different plugins for Andoird in our documentation.

CMakeLists.txt
astro-friedel: fixed

utils/serialization/latency.cpp
astro-friedel: It should remain there as it needs to be compiled into that library, otherwise the serialization will not work.

rsingh76 (approved):

PR #494, opened 2026-05-22
Android offload rendering client

astro-friedel:
This PR should not be reviewed until after https://github.com/ILLIXR/ILLIXR/pull/493 is approved and merged.

This work updates the offload_rendering_client plugin to work on Android systems. It supports HEVC and AV1 encoding schemes, but not H264 (this is due to hardware limitations on our test platform)

PR #497, opened 2026-08-20
Android network latency

astro-friedel:
Adding network latency measurements to Android devices


astro-friedel:

